PR #15, opened 2015-03-18
setup.py & rpm spec file

jpopelka:


TomasTomecek:
Great work Jirko!

Will inline comment.

.gitignore
```diff
@@ -26,7 +26,7 @@ var/
 #  Usually these files are written by a python script from a template
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
 *.manifest
-*.spec
+#*.spec
```
TomasTomecek: I think this line could be safely removed.

osbs.spec
```diff
@@ -0,0 +1,100 @@
...
+
+BuildRequires:  python2-devel
+BuildRequires:  python-setuptools
+
+%if 0%{?with_python3}
+BuildRequires:  python3-devel
+BuildRequires:  python3-setuptools
+%endif
+
+Requires:       python-pycurl
```
TomasTomecek: This is a tough one. The way osbs is written (you've noticed it probably), you can either choose pycurl or request -- you don't need both. The preferred library being pycurl (since it supports kerberos and transfer chunked encoding natively).
Therefore I would be in favor of requiring pycurl only (and probably comment `import requests` in code). What do you think?

PR #18, opened 2015-03-25
pycurl.URL doesn't accept unicode, so use str().

twaugh:


TomasTomecek:
Good point!

Have you tried this with python 3?

twaugh:
No, I haven't, I've been testing the client on RHEL-7.1.

TomasTomecek:
I think that the string should be properly `encode`d. Let's merge this and figure out possible issues in future.

PR #20, opened 2015-03-26
Some fixes for reporting HTTP errors

twaugh:


TomasTomecek:
osbs definitely needs more python 3 love; thanks for these fixes

PR #24, opened 2015-04-01
Build types

mmilata:
Add `simple` build JSON. Make switching between build JSONs easier.

### DIFF
--- a/inputs/simple.json
+++ b/inputs/simple.json
@@ -1,0 +1,30 @@
+{
+      "metadata":{
+        "name": "{{NAME}}"
+      },
+      "kind": "Build",
+      "apiVersion": "v1beta1",
+      "parameters": {
+        "source" : {
+          "type" : "Git",
+          "git" : {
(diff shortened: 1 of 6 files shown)


TomasTomecek:
This approach looks great. There are just couple minor things I inline-commented.

What I didn't mention is `BuildRequest` constructor (because it's not really an issue). It takes arguments like `user` and `component`. That means that every build type should consume those. Well, if we find a build type which won't use those, we can still update it on the fly.

mmilata:
Yeah, I wasn't really sure regarding which parameters will be common so I figured I'll split it like this and then rearrange it when we add more types.

I'll address your issues and update the PR today.

TomasTomecek:
Great! Thanks Martin.

mmilata:
Updated. I've added another commit to improve the behaviour when a required parameter is missing though I'm not very happy about it - I think that it duplicates some of the code from `conf.py` but I'm not sure what a better solution should look like.

PR #37, opened 2015-04-20
Test build JSON for implicit labels.

twaugh:
Here's an extension of one of the existing test cases to check that the `Architecture`, `Vendor`, `Build_Host`, and `Authoritative_Registry` labels will be set via the prebuild plugin.

It doesn't test loading those values from config, only the translation from build request keywords into build JSON.


TomasTomecek:
Well done!

PR #38, opened 2015-04-20
Removed unused var from DockJsonManipulator.

twaugh:


TomasTomecek:
+1

twaugh:
The failed CI check seems to be due to me merging it too quickly or something.

PR #40, opened 2015-04-22
http: Don't verify host if verify_ssl is disabled

pbabinca:
Sometimes OpenShift self-signed certificate isn't created correctly and
CommonName is 127.0.0.1 instead of IP address/FQDN under which is
OpenShift available. This change allows ignore this inconsistency which
would otherwise cause following fatal error:

>    SSL: certificate subject name (127.0.0.1) does not match target host
>    name 'openshift.example.com'


TomasTomecek:
:+1:

PR #45, opened 2015-04-28
More coherent exception handling

### LINKED ISSUE #43: introduce better exception strategy
Currently, error management within osbs client is pretty weak. It would be great if all exceptions would have same type (and not that some of them are from pycurl, some from python etc.)
E.g.
```
error: (28, \'Operation timed out after 0 milliseconds with 0 out of 0 bytes received\')
error: (7, \'Failed to connect to <ip> port 8443: Connection refused\')
error: (56, \'GnuTLS recv error (-110): The TLS connection was non-properly terminated.\')
```
These should be under some `OsbsConnectionError`, not pycurl.

twaugh:
New exceptions module with base OsbsException class and two
exceptions derived from it:
 * OsbsResponseException, for OpenShift 'fail' responses
 * OsbsNetworkException, for (possibly transient) network
   failures

The OsbsResponseException is thrown by the api and core modules
when they see a fail HTTP response.

The PycurlAdapter class catches exceptions thrown during
request handling and maps them to either OsbsNetworkException
or OsbsException as appropriate.

This fixes #43.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -2,7 +2,8 @@ from __future__ import print_function, unicode_literals, absolute_import
 import json
 from osbs.build import BuildManager, BuildResponse
 from osbs.constants import DEFAULT_NAMESPACE
-from osbs.core import Openshift, OpenshiftException
+from osbs.core import Openshift
+from osbs.exceptions import OsbsResponseException
 
 
 class OSBS(object):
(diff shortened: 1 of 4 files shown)


TomasTomecek:
@twaugh Tim, could you please take care of the issues found by @landscape-bot

twaugh:
@landscape-bot ++

PR #47, opened 2015-04-28
Avoid access to protected class member.

twaugh:
Hopefully this increases the project's health according to Landscape.

### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -86,7 +86,7 @@ class Response(object):
     def headers(self):
         if self._headers is None:
             ### FIXME: the API for this is different in Python3
-            m = httplib.HTTPMessage(self._raw_headers, False)
+            m = httplib.HTTPMessage(self.raw_headers, False)
             m.readheaders()
             self._headers = m.dict
         return self._headers
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -76,7 +76,7 @@ class Response(object):
 
     @property
     def raw_headers(self):
-        raise RuntimeError("Only settable")
+        return self._raw_headers
 
     @raw_headers.setter
     def raw_headers(self, raw_headers):
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -75,6 +75,14 @@ class Response(object):
         self._headers = None
 
     @property
+    def raw_headers(self):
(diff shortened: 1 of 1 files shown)


osbs/http.py
```diff
@@ -75,6 +75,14 @@ def __init__(self, status_code=0, content=b'', curl=None, raw_headers=None):
         self._headers = None
 
     @property
+    def raw_headers(self):
+        raise RuntimeError("Only settable")
```
TomasTomecek: Why not make them readable?

TomasTomecek:
plus in `headers` I would read via `raw_headers` and not `_raw_headers`. Otherwise it looks great! Maybe we could exclude those complexity checkers and rump up the health to 100 :)

twaugh:
Well, in that case, why not just change `_raw_headers` to `raw_headers`?

TomasTomecek:
That's also possible. Doing it with `property` is the official pythonic way. I don't have no strong opinions for one way or the other. After all, these are inner guts of osbs client.

TomasTomecek:
could you please change the access of `raw_headers` here? https://github.com/DBuildService/osbs/pull/47/files#diff-a49481931f2791e303e626d1083bc601R89 we can merge then from my pov

twaugh:
Not sure where you mean -- the accesses from `Response.__init__()` and `Response.headers()` maybe?

TomasTomecek:
Ah, I though I'm including line number too. Silly me!
https://github.com/DBuildService/osbs/pull/47/files#diff-a49481931f2791e303e626d1083bc601L81
```diff
- m = httplib.HTTPMessage(self._raw_headers, False)
+ m = httplib.HTTPMessage(self.raw_headers, False)
```

twaugh:
OK. How about the access from `__init__()`?

```diff
-        self._raw_headers = raw_headers
+        self.raw_headers = raw_headers
```

TomasTomecek:
Good catch! But I think that some static analyzer will complain that `self._raw_headers` is not defined in constructor.

twaugh:
OK, I committed just the change you mentioned.

TomasTomecek:
Thanks for doing this Tim!

Issue #51, opened 2015-05-01
Nondescript IOError Exception


maxamillion:
```
# osbs build -g https://github.com/maxamillion/osbs-dockerfile-test.git -c fedora-httpd-test -u admiller
2015-05-01 16:51:51,314 - osbs - ERROR - Exception caught: IOError(2, 'No such file or directory')
```

Contents of /etc/osbs.conf are as follows:

```
[general]
build_json_dir = /usr/share/osbs/

[default]
openshift_uri = https://localhost:8443/
# with openshift 0.4.3+ this will no longer be needed
kubelet_uri = http://localhost:10250/
# if you want to get packages from koji (koji plugin in dock)
# you need to setup koji hub and root
# this sample is for fedora
koji_root = http://koji.fedoraproject.org/
koji_hub = http://koji.fedoraproject.org/kojihub
# in case of using artifacts plugin, you should provide a command
# how to fetch artifacts
sources_command = fedpkg sources
# from where should be images pulled and where should be pushed?
registry_uri = localhost
verify_ssl = false
build_type = simple
```

And I am able to at least check the status of builds, just unable to submit any

```
# osbs list-builds
BUILD NAME                                       STATUS           IMAGE NAME                                                      
```

twaugh:
Try adding `--verbose` at the beginning to see what's causing it?

Hmm, config errors really ought to be trapped better.

maxamillion:
This looks a little more promising.

```
# osbs --verbose build -g https://github.com/maxamillion/osbs-dockerfile-test.git -c fedora-httpd-test -u admiller 
2015-05-04 14:49:30,771 - osbs - DEBUG - Logging level set to debug
Traceback (most recent call last):
  File "/bin/osbs", line 9, in <module>
    load_entry_point('osbs==0.4', 'console_scripts', 'osbs')()
  File "build/bdist.linux-x86_64/egg/osbs/cli/main.py", line 306, in main
  File "build/bdist.linux-x86_64/egg/osbs/cli/main.py", line 136, in cmd_build
  File "build/bdist.linux-x86_64/egg/osbs/api.py", line 66, in create_build
  File "build/bdist.linux-x86_64/egg/osbs/build.py", line 322, in get_build
  File "build/bdist.linux-x86_64/egg/osbs/build.py", line 144, in render
  File "build/bdist.linux-x86_64/egg/osbs/build.py", line 174, in template
[...]

PR #55, opened 2015-05-05
More robustness for exception handling

twaugh:
As reported in #51, there are several cases where exceptions are raised without descriptive messages, and that do not match OsbsException.

This commit makes several improvements:

* The `osbs.api.OSBS` API functions are now decorated by `@osbsapi`, which
ensures exceptions raised will match OsbsException.
* Exceptions while handling chunked HTTP responses now correctly raise
OsbsNetworkException or OsbsException as appropriate.
* Configuration errors now also raise OsbsException, and there are some
test cases for this.
* The command-line interface describes network errors more fully.


osbs/api.py
```diff
@@ -10,11 +10,25 @@
 from osbs.build import BuildManager, BuildResponse
 from osbs.constants import DEFAULT_NAMESPACE
 from osbs.core import Openshift
-from osbs.exceptions import OsbsResponseException
+from osbs.exceptions import OsbsException, OsbsResponseException
 
 
+def osbsapi(func):
+    def catch_exceptions(*args, **kwargs):
```
TomasTomecek: could you please decorate this with [`wraps`](https://docs.python.org/2/library/functools.html#functools.wraps)? Creating decorators without `wraps` [changes `__doc__` and `__name__`](http://www.jeffknupp.com/blog/2013/11/29/improve-your-python-decorators-explained/).

osbs/api.py
```diff
@@ -10,11 +10,25 @@
...
+        try:
+            return func(*args, **kwargs)
+        except OsbsException:
+            # Re-raise OsbsExceptions
+            raise
+        except Exception as e:
+            # Convert anything else to OsbsException
+            raise OsbsException(e)
```
TomasTomecek: I would be in favor of expanding message here, because we lose information about original exc type:
```
In [2]: from osbs.exceptions import OsbsException
In [3]: a = {}
In [4]: try:
   ...:     a["a"]
   ...: except Exception as ex:
   ...:     raise OsbsException(ex)
   ...: 
---------------------------------------------------------------------------
OsbsException                             Traceback (most recent call last)
<ipython-input-4-9e9da677c841> in <module>()
      2     a["a"]
      3 except Exception as ex:
----> 4     raise OsbsException(ex)
        global OsbsException = <class 'osbs.exceptions.OsbsException'>
        global ex = KeyError('a',)
      5 

OsbsException: 'a'

In [5]: try:
    a["a"]
except Exception as ex:
    raise OsbsException(repr(ex))
   ...: 
---------------------------------------------------------------------------
OsbsException                             Traceback (most recent call last)
<ipython-input-5-fe87eae34c24> in <module>()
      2     a["a"]
      3 except Exception as ex:
----> 4     raise OsbsException(repr(ex))
        global OsbsException = <class 'osbs.exceptions.OsbsException'>
        global repr = undefined
        global ex = KeyError('a',)
      5 

OsbsException: KeyError('a',)
```
What do you think, Tim?

TomasTomecek:
I really like this, awesome job!

osbs/api.py
twaugh: Sure, we can do that. In python3, of course, `__context__` makes it clear what the original exception was (and the traceback shows it).

twaugh:
Thanks. OK, `@wraps` added, and now `repr()` is used when chaining up to OsbsException.

PR #56, opened 2015-05-05
add landscape.io config

### LINKED ISSUE #53: add config for landscape.io to ignore complexity checkers
...so we can be **100%**!
EDIT: we should probably check the code with pep8 too so the code looks neat; anyone wants to play with landscape config file?

TomasTomecek:
Fixes #53 

This enables [pep8](https://www.python.org/dev/peps/pep-0008/). I would really like if our code could be pep8 compliant. Especially when you can enable pep8 in you text editors/IDEs. What do you think?

### DIFF
--- a/.landscape.yml
+++ b/.landscape.yml
@@ -1,0 +1,18 @@
+doc-warnings: no
+test-warnings: no
+strictness: veryhigh
+max-line-length: 120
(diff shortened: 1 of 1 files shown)


mmilata:
:+1:

PR #59, opened 2015-05-06
Use email.parser.Parser() to parse HTTP headers

twaugh:
This works on both Python 2 and 3.

Low priority.

### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -19,6 +19,7 @@ from __future__ import print_function, absolute_import, unicode_literals
 import re
 import json
 import logging
+import email.parser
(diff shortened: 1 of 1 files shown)


TomasTomecek:
Well, I'm not really in favor of parsing http headers with parser for email, but let's give it a shot.

twaugh:
The thing is, the `HTTPMessage` interface was never defined. I'm not sure what better interface there is.

TomasTomecek:
Looks like that your approach actually makes a lot of sense. In py3, [`HttpMessage` is implemented](https://docs.python.org/3/library/http.client.html#httpmessage-objects) via `email.Message`.

PR #61, opened 2015-05-07
Decode HTTP headers from ISO-8859-1 before using split() on them

twaugh:
This was causing failures with Python 3.

### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -321,7 +321,7 @@ class PycurlAdapter(object):
             # self.response_headers contains headers from all responses - even
             # without FOLLOWLOCATION there might be multiple sets of headers
             # due to 401 Unauthorized. We only care about the last response.
-            allheaders = self.response_headers.getvalue().decode('iso-8859-1')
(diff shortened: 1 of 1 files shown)


TomasTomecek:
How do you know the name of the encoding?

twaugh:
[RFC 7203 §3.2.4](https://tools.ietf.org/html/rfc7230#section-3.2.4) says:
> Historically, HTTP has allowed field content with text in the
> ISO-8859-1 charset [ISO-8859-1], supporting other charsets only
> through use of [RFC2047] encoding.  In practice, most HTTP header
> field values use only a subset of the US-ASCII charset [USASCII].
> Newly defined header fields SHOULD limit their field values to
> US-ASCII octets.  A recipient SHOULD treat other octets in field
> content (obs-text) as opaque data.

So maybe it should be `decode('ascii')` (or just `decode()`) instead?

TomasTomecek:
I'm not sure here. Let's merge it and we'll see. Maybe we could create some encoding-related tests. Ideally against real webserver. Will open up an issue.

PR #63, opened 2015-05-12
Better specfile

bkabrda:
This is pretty much the same as https://github.com/DBuildService/dock/commit/6c4d0421d7983aa677bcb0cd025a8781af8fd330 to keep things in good shape and consistent.


TomasTomecek:
+1

PR #69, opened 2015-05-13
http: fix parsing headers on 2.7

TomasTomecek:
@pbabinca this should resolve your issue

Unfortunately, it still doesn't work on py 3 (on my machine).


TomasTomecek:
py 3 should be resolved at some point; let's at least fix py 2 for now

Issue #74, opened 2015-05-18
Fix headers parsing on Python 3


twaugh:
The `test_parse_headers()` test fails on Python 3.

PR #75, opened 2015-05-18
Skip failing Python 3 test

twaugh:
We know that this test fails with Python 3, and issue #74 is open to track that. How about skipping it on Python 3 until it's important enough to fix?


TomasTomecek:
LGTM

PR #77, opened 2015-05-18
OpenShift 0.5 compatibility

mmilata:


osbs/core.py
```diff
@@ -175,16 +171,15 @@ def logs(self, build_id, follow=False, namespace=DEFAULT_NAMESPACE):
...
-            # 0.4.1
-            # FIXME: remove this once 0.4.3 is deployed everywhere
-            buildlogs_url = self._build_k8s_url(
-                "containerLogs/default/build-%s/custom-build?follow=%d" % (
-                    build_id, 1 if follow else 0))
+            # 0.4.3
+            # FIXME: remove this once 0.5.? is deployed everywhere
+            buildlogs_url = self._build_url("proxy/buildLogs/%s/" % build_id,
+                    follow=(1 if follow else 0), namespace=namespace)
             response.close_multi()
```
mmilata: I'm not sure what this does and if it should stay here even if we are GETting from the same host (which was previously not the case).

mmilata:
(Will fix pep8 issues, do not merge yet.)

osbs/core.py
```diff
@@ -175,16 +171,17 @@ def logs(self, build_id, follow=False, namespace=DEFAULT_NAMESPACE):
...
-            # FIXME: remove this once 0.4.3 is deployed everywhere
-            buildlogs_url = self._build_k8s_url(
-                "containerLogs/default/build-%s/custom-build?follow=%d" % (
-                    build_id, 1 if follow else 0))
+            # 0.4.3
+            # FIXME: remove this once 0.5.? is deployed everywhere
+            buildlogs_url = self._build_url("proxy/buildLogs/%s/" % build_id,
+                                            follow=(1 if follow else 0),
+                                            namespace=namespace)
             response.close_multi()
```
mmilata: I'm not sure what this does and if it should stay here even if we are GETting from the same host (which was previously not the case).

TomasTomecek: The problem was, that curl handle stayed open on the multi request and subsequent request raised an exception. I guess we could implement it as context manager.

osbs/core.py
```diff
@@ -59,15 +58,12 @@ def __init__(self, openshift_api_url, openshift_oauth_url, kubelet_base, verbose
...
-        if namespace:
-            url += "?namespace=%s" % namespace
-            return urlparse.urljoin(self.os_api_url, url)
-        else:
-            return urlparse.urljoin(self.os_api_url, url)
-
-    def _build_k8s_url(self, url):
-        return urlparse.urljoin(self.kubelet_base, url)
+    def _build_url(self, url, **query):
+        query = [k + '=' + str(v) for k, v in query.items()]
```
TomasTomecek: Am checking if urlparse doesn't have method for this but looks like it doesn't =/

osbs/core.py
mmilata: So that means we should keep the `response.close_multi()` line for now, right?

osbs/core.py
mmilata: Actually there's `urllib.urlencode` and `urllib.parse.urlencode` (py3). I'll update the PR.

osbs/core.py
TomasTomecek: yes, definitely; it could be great to add comment there so anyone doesn't remove it by accident

TomasTomecek:
Merging and am going to give it a try. Thanks Martin!

twaugh:
Would be great to get add `tests/mock_jsons/0.5.x` once it's working.

mmilata:
@twaugh good idea! I'd like to add them once 0.5.2 is out and tested against osbs (0.5.1 has a bug that prevents it from working with osbs using the v1beta1 api).

TomasTomecek:
TBH, that shell script is really bad and I feel bad. I think we should rewrite it to python (and ideally use `osbs.http`)

PR #87, opened 2015-05-20
inputs: add_labels_in_dockerfile must be run before add_dockerfile

jpopelka:
add_dockerfile gets N-V-R either from nvr argument or from
LABELs in Dockerfile. If the LABELs are added to Dockerfile with
add_labels_in_dockerfile then it has to be run before add_dockerfile.

### DIFF
--- a/inputs/prod-without-koji_inner.json
+++ b/inputs/prod-without-koji_inner.json
@@ -4,13 +4,13 @@
       "name": "change_from_in_dockerfile"
     },
     {
-      "name": "add_dockerfile"
-    },
-    {
       "args": {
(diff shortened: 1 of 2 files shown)


TomasTomecek:
To be honest, I've made the order by purpose. The thing is that we are setting some vendor specific labels with the `add_labels` plugin which I don't think should be used by anyone except from us (`Build_Host`, `Vendor`). `Name`, `Version`, `Release` should be specified by maintainer **always**. osbs won't do that for you.

jpopelka:
> Name, Version, Release should be specified by maintainer always.

OK, closing then.

PR #93, opened 2015-05-22
Make it possible to package post-releases properly. Fixes #91

### LINKED ISSUE #91: Update spec file to create pre-releases
I'd like to test pre-released osbs versions but I need to have it as a rpm package. Please modify spec file to create dev pre-releases (with correct upgrade path between versions).

bkabrda:


TomasTomecek:
what should we put to `setup.py`?

bkabrda:
`setup.py` doesn't require any change AFAICS... or does it?

TomasTomecek:
aha, so we'll just bump spec and update commit in spec? sounds good to me

bkabrda:
Yeah, exactly.

TomasTomecek:
Thanks Slavek! I'm going to try it right away :)

PR #99, opened 2015-05-25
New method for BuildResponse() object: is_pending()

pbabinca:
I use it together with method is_running() to find out if OpenShift will
do something with the build or I can proceed with getting the
build and/or buildlogs.


TomasTomecek:
:+1:

PR #101, opened 2015-05-25
Ensure build_id property of BuildResponse is always unicode

pbabinca:
Without this osbs.core.wait():238 never ever found the build:

    if obj_name == build_id:

Line above compared isinstance(obj_name, unicode) with
isinstance(build_id, str). Tested on python 2.6.

For the rest of the osbs code where strings are used as identifiers you may want to explicitly type this as unicode().

### DIFF
--- a/osbs/build/build_response.py
+++ b/osbs/build/build_response.py
@@ -38,7 +38,7 @@ class BuildResponse(object):
     @property
     def status(self):
         if self._status is None:
-            self._status = self.json['status'].lower()
+            self._status = unicode(self.json['status'].lower())
         return self._status
 
     @property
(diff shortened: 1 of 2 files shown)


pbabinca:
Please don't merge the patch now. I suspect this might not be enough even for that wait().

TomasTomecek:
Could someone look into this please? @jpopelka @twaugh @mmilata @bkabrda 

This is basically number one bug which should be solved asap. We should also cover this by tests so it won't happen ever again.

pbabinca:
python 3 support needs to be added. I leave that on you. :/

TomasTomecek:
I feel like that putting everywhere

```python
from __future__ import unicode_literals
```

should be enough

PR #103, opened 2015-05-26
fix curl multi and watch

### LINKED ISSUE #102: osbs.wait_for_build_to_finish() doesn't wait for build to finish
`osbs.wait_for_build_to_finish()` doesn't work as advertised. `wait()` sometimes (always?) prints:
```
osbs.core:237 object has changed: 'u'MODIFIED'', status: 'Running', name: 'container-docker-20150525-142550'
osbs.core:239 matching build found
```
but this isn't the state `wait()` expects:
```
obj_status: u'running' not in states: [u'failed', u'complete', u'error', u'cancelled']
```

If the connection is closed now the error with build not found is printed.

I guess it should restart wait() in this case. Maybe some retry count or timeout needs to be implemented to not wait forever.

Note: I use #101 with unicode literals on python 2.6.

TomasTomecek:
Resolves #102
Closes #101

+ fix repositories in cli output (am lazy to open new PR)
+ display commit_id in cli output (more laziness)

### DIFF
--- a/osbs/__init__.py
+++ b/osbs/__init__.py
@@ -5,6 +5,8 @@ All rights reserved.
 This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
+from __future__ import print_function, absolute_import, unicode_literals
+
 import logging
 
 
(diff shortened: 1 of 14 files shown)


mmilata:
Looks good & fixes the issue for me.

TomasTomecek:
Also verified on rhel 6; unfortunately still hitting https://github.com/openshift/origin/issues/2348

PR #109, opened 2015-05-26
Support for custom builds with source secrets

twaugh:
This provides support for builds that require a secret, such as a key or certificate, as part of their workflow.

It assumes the OpenShift 'Secret'-type resource has already been created, and simply sets the `sourceSecret` resource name as directed.


twaugh:
Rebased against master.

TomasTomecek:
do you think we should also add argparse argument for the secret name?

twaugh:
I thought I had, soryr. Added now.

twaugh:
Re-based. I guess we should hold off merging this until https://github.com/DBuildService/dock/issues/124 is resolved.

twaugh:
OK, I think the problem I was seeing is at least understood now (@csrwng is investigating).

I don't think it needs to hold up merging this PR.

PR #121, opened 2015-06-03
Change source registry for simple builds

### LINKED ISSUE #117: simple build type doesn't work with insecure registries
all we need to do is to stuff this plugin to build json:

https://github.com/DBuildService/dock/blob/master/dock/plugins/pre_change_source_registry.py

mmilata:
Fixes #117 - can you please check if it really works? The ticket is a bit vague so I'm not sure if it fixes the right issue.

### DIFF
--- a/inputs/simple_inner.json
+++ b/inputs/simple_inner.json
@@ -5,6 +5,13 @@
     },
     {
       "name": "dockerfile_content"
+    },
+    {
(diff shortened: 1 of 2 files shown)


mmilata:
After offline discussion it appears that the issue is indeed fixed. Let me just merge this.

PR #135, opened 2015-06-16
Use v1beta3 OpenShift API

mmilata:
This should make osbs work agaist OpenShift 0.5.2 (currently deployed) and 0.5.4 (supports v1 API, last to support v1beta3).

The tests are in separate commit for easier review, I can squash the commits if needed.

There's a small problem with `prod-with-secret` build - if no output docker image is specified then the build hangs in New state. So we currently specify dummy image name.


mmilata:
Rebased on top of master.

TomasTomecek:
I'll look into this on Monday. The number of changed files scared me a lot.

mmilata:
Rebasing

TomasTomecek:
listing builds, getting logs works fine, creating builds not:
```
u'output': {u'to': {u'kind': u'DockerImage', u'name': u'osbs/FIXME'}}}
```
which resulted into:
```
docker.errors.APIError: 500 Server Error: Internal Server Error ("Invalid repository name (FIXME), only [a-z0-9-_.] are allowed")
```

mmilata:
Oh ... what about now? I changed the dummy image name to `osbs/fixme` - openshift needs to have some docker repository name otherwise the build hangs.

mmilata:
Reported as openshift/origin#3391.

twaugh:
I've reviewed these changes and they look good to me. I didn't find any problems testing it locally.

PR #140, opened 2015-06-17
Set resource limits for builds if specified (#85)

### LINKED ISSUE #85: limit number of parallel builds
openshift should help us with this somehow

suggested by @goldmann

@csrwng is there a way to limit number of parallel builds in openshift?

twaugh:
To set resource quota, use `osc create -f quota.yaml` on a file like this:

```yaml
apiVersion: v1
kind: ResourceQuota
metadata:
  name: quota
spec:
  hard:
    # Number of parallel builds
    pods: 10
    # Compute units (KCU) for all builds
    cpu: "2"
    # Memory for all builds
    memory: 2Gi
    # Storage for all builds
    storage: 20Gi
```

Now `osc describe quota myquota` will show the used and total amount of each resource type.

You can set default per-build limits either by setting them in the OSBS build configuration file (`cpu_limit`, `memory_limit`, `storage_limit`), or by setting defaults that OpenShift will apply to builds that do not specify their limits:

```yaml
apiVersion: v1
kind: LimitRange
metadata:
  name: limits
spec:
  limits:
  - default:
      # 100m == 0.1 KCU
      cpu: 100m
      memory: 512Mi
      storage: 2Gi
    type: Container
```

You can set per-build limits using the `--cpu-limit`, `--memory-limit`, and `--storage-limit` OSBS command line options.

You can set quota for `pods`, `cpu`, `memory`, and `storage`, but you do not have to set quota for all of them. If you set quota for `cpu`, `memory`, or `storage`, you must also set limits for that resource: builds not specifying limits are unbounded and will not be admitted.

You do not have to set limits for `pods`. Each build takes a fixed quantity (1 pod).

Fixes #85.

### DIFF
--- a/docs/resource.md
+++ b/docs/resource.md
@@ -1,0 +1,45 @@
+# Resource limiting
+
+To set resource quota, use `osc create -f quota.yaml` on a file like this:
+
+```yaml
+apiVersion: v1
+kind: ResourceQuota
+metadata:
+  name: myquota
+spec:
+  hard:
+    # Number of parallel builds
+    pods: 10
+    # Compute units (KCU) for all builds
+    cpu: "2"
+    # Memory for all builds
+    memory: 2Gi
+    # Storage for all builds
+    storage: 20Gi
+```
+
+Now `osc describe quota myquota` will show the used and total amount of each resource type.
+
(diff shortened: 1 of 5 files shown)


TomasTomecek:
How about # of concurrent pods? Also, could you please create a new document in docs with the content of your first post? That's really valuable info.

twaugh:
I've adjusted the description above to have the same content as the new `docs/resource.md`.

TomasTomecek:
Looks great!

PR #145, opened 2015-06-29
API fixes for atomic-reactor

mmilata:
Needed for DBuildService/atomic-reactor#197.


twaugh:
This looks good to me.

PR #146, opened 2015-06-30
new command: cancel build

### LINKED ISSUE #144: New call to osbs to cancel-build
I'd like to have similar functionality as osc-provides with  `osc cancel-build` command via osbs-client API.

TomasTomecek:
Fixes #144

OpenShift does the exact same thing.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -84,6 +84,12 @@ class OSBS(object):
         return build_response
 
     @osbsapi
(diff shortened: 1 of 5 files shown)


mmilata:
Looks good and flips the build into Cancelled state which is indeed what `oc cancel-build` does. However the build does not really seem to stop (immediately) and after restarting openshift it is run again. Openshift team is probably aware of that: openshift/origin#3353.

PR #148, opened 2015-07-01
secret build type: use unique image name

TomasTomecek:
This is used to construct unique tag for image.

@mmilata can you review and merge please?


mmilata:
Looks good but my openshift instance seems to ignore the :tag ... let me investigate a bit.

mmilata:
The build json has the tag but openshift 1.0.0 seems to ignore it (so the image is tagged as latest). Merging this anyway since it's still an improvement over the previous behaviour.

TomasTomecek:
Oh boy, we should definitely track the regression. Thanks for doing the merges Martin.

mmilata:
Opened #149 for that.

PR #156, opened 2015-07-08
Openshift.start_build(): give correct argument to create_build()

twaugh:
`Openshift.create_build()` needs the JSON, not the internal representation of it.


TomasTomecek:
LGTM

PR #168, opened 2015-07-24
Reject URIs as values of registry_uri

mmilata:
The name is confusing:/

### DIFF
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -136,7 +136,10 @@ class CommonSpec(BuildTypeSpec):
         self.git_ref.value = git_ref
         self.user.value = user
         self.component.value = component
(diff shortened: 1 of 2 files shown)


twaugh:
Maybe we could just drop the scheme part altogether? e.g. https://github.com/projectatomic/atomic-reactor/commit/53cc20e0570a030dd15031bd43758c7aca8e0283

mmilata:
Good idea - I'll update the PR.

mmilata:
Updated.

twaugh:
@mmilata Mind if I add a test case?

twaugh:
LGTM

mmilata:
Thanks! Feel free to not merge until I add a test case next time;)

Issue #129, opened 2015-06-11
merge all prod build types together


TomasTomecek:
They all do very similar thing: let's adapt the workflow according to provided configuration:

* [x] if yum repo is specified, don't use koji plugin
* [x] pulp
 * [x] if admin and password are not specified, ensure secrets is on
 * [ ] if squashing is on, don't load squashed image to engine
* [x] if secret is specified, use k8s secrets

twaugh:
[ ] if no NFS server path is specified, don't use `cp_built_image_to_nfs` plugin

mmilata:
I wonder if we should get rid of *simple* build type as well? It was supposed to do less things than prod builds, however most of the things are now optional.

twaugh:
Agreed.

PR #169, opened 2015-07-27
Merge prod build types (#129)

twaugh:
This does most of the things mentioned in #129. It doesn't set squashing config for the pulp plugin yet as I'm not sure I understand how it needs to behave.


twaugh:
Test coverage appears to decrease: this seems to be because there is less code. :-)

TomasTomecek:
This is exactly what I had in mind!

PR #174, opened 2015-07-28
Name exit plugins as exit plugins (fixes #171)

### LINKED ISSUE #171: Some postbuild plugins are now exit plugins
Some plugins were previously postbuild plugins but are now exit plugins. Although they will continue to work when named in `postbuild_plugins`, they will actually run in the exit phase and we should name them in `exit_plugins` instead.

- [ ] `remove_built_image`
- [X] `store_logs_to_file` (nothing to do)
- [ ] `store_metadata_in_osv3`

Probably worth waiting until #169 is merged before fixing this.

twaugh:
With this change, any local `osbs.conf` referencing JSON files that still list these plugins as post-build plugins should still continue to work.

### DIFF
--- a/inputs/prod_inner.json
+++ b/inputs/prod_inner.json
@@ -80,7 +80,9 @@
         "image_id": "BUILT_IMAGE_ID"
       },
       "name": "all_rpm_packages"
-    },
+    }
+  ],
+  "exit_plugins": [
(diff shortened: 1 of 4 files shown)


TomasTomecek:
LGTM

PR #180, opened 2015-07-29
Use DockJsonManipulator member in BuildRequest subclasses

mmilata:
Cosmetic change - not that creating new instances is wrong, I was just confused why each subclass creates new instance when we already have it as a member.


twaugh:
LGTM

twaugh:
@bkabrda happy for me to merge this? I know you're probably working in the same area.

bkabrda:
Feel free to merge, I actually didn't need to touch any of this (see #181 for WIP).

PR #181, opened 2015-07-29
Create build by creating BuildConfig and instantiating it

bkabrda:
Don't merge yet.

In this PR, I'd like to achieve two things:
- Creating BuildConfig and instantiating it instead of simply creating a build.
- Creating ImageStream for the layer that this image will build from.

The first commit kind of achieves the first, but there's more to be done (setting some labels for the BuildConfig etc)


twaugh:
This looks really great!

bkabrda:
So, most of the work is done here; as a next step, I'll create a document that describes the whole build process in detail and submit it here, so that we can check what precisely we have and possibly implement the missing pieces in atomic reactor as well. Note that the imagestream trigger depends on implementation of https://github.com/openshift/origin/pull/3748 and won't do anything right now (although everything still works).

bkabrda:
I just added the document that describes the build process (to be more precise, it describes the state that I'm targeting with this PR).
Things that still need to be done:
- It is now assumed that the name of the built image in registry is branch/component. I think this is ok, but it'd be nice if someone could verify that this makes sense.
- Do we want to add the rebuild triggers to all images? E.g. I think it may not make sense to add triggers for images that base of specific tag (e.g. `rhel7.1/python:sometag), since the images under the tags will not be updated (I think @twaugh mentioned that during one of our conversations. OTOH it doesn't hurt to add this trigger and it would save us some specialcasing (also, this scheme may only be true for distro building, but other people interested in using OSBS for different usecases may want to have these triggers in place).
- AFAICS we already have working `ImportImagePlugin`, but it only *updates* the `ImageStream`, but we also want it to *create* the `ImageStream`, if it doesn't exist yet. @twaugh is this ok? Should I open a bug for the plugin?
- I have to find out the best way to label the `BuildConfigs` so that the system can distinguish between an automatically scheduled rebuild and a user-run build. I'll update both documentation and code at once in another commit. When that is done, I think it'll finally be possible to merge this PR.

Comments are welcome @TomasTomecek @mmilata @twaugh @jpopelka

twaugh:
Doc looks really good.

I think users need to be able to determine whether rebuilds are enabled on a per-build basis.

Yes, please open a new issue for `import_image`.

osbs/api.py
```diff
@@ -127,15 +129,58 @@ def create_build_from_buildrequest(self, build_request, namespace=DEFAULT_NAMESP
...
+    def _create_build_config_and_build(self, build_json, namespace):
+        # TODO: test this method more thoroughly
+        build_config_name = build_json['metadata']['name']
+
+        # check if a build already exists for this config; if so then raise
+        # TODO: could we do this more effectively than traversing all?
+        running_builds = self.os.list_builds(namespace=namespace).json()
+        for rb in running_builds['items']:
+            try:
+                phase = rb['status']['phase'].lower()
```
TomasTomecek: I would prefer using one piece of code for this kind of stuff: https://github.com/projectatomic/osbs-client/blob/c60ede765537d322135431ecda241d88d045dab5/osbs/build/build_response.py#L51

osbs/api.py
```diff
@@ -127,15 +129,58 @@ def create_build_from_buildrequest(self, build_request, namespace=DEFAULT_NAMESP
...
+            deep_update(existing_bc, build_json)
+            logger.debug('build config for %s already exists, updating...', build_config_name)
+            self.os.update_build_config(build_config_name, json.dumps(existing_bc), namespace)
+        else:
+            # if it doesn't exist, then create it
+            logger.debug('build config for %s doesn\'t exist, creating...', build_config_name)
+            self.os.create_build_config(json.dumps(build_json), namespace=namespace)
+        return self.os.start_build(build_config_name, namespace=namespace)
+
+    def _get_base_image(self, git_uri, git_ref):
```
TomasTomecek: I think this should be part of utils.

osbs/utils.py
```diff
@@ -20,3 +25,25 @@ def graceful_chain_get(d, *args):
...
+
+def checkout_git_repo(uri, commit):
+    tmpdir = tempfile.mkdtemp()
+    subprocess.check_call(['git', 'clone', uri, '-b', commit, tmpdir], stdout=subprocess.PIPE,
+                          stderr=subprocess.PIPE)
+    return tmpdir
+
+
+def get_base_image(repo_dir):
+    df_path = os.path.join(repo_dir, 'Dockerfile')
```
TomasTomecek: I wonder if we should ever support different names for dockerfiles.

twaugh: Currently `dockerfile_parse` requires that the file is named `Dockerfile`. If it isn't, it assumes it is a directory and looks for a `Dockerfile` inside it.

inputs/simple_inner.json
```diff
@@ -1,6 +1,13 @@
 {
   "prebuild_plugins": [
     {
+      "args": {
+        "key": "is_autorebuild",
```
twaugh: This now needs to be `label_key`.

inputs/simple_inner.json
```diff
@@ -1,6 +1,13 @@
...
+        "value": "true"
```
twaugh: This now needs to be `label_value`.

inputs/simple_inner.json
```diff
@@ -1,6 +1,13 @@
...
+      },
```
twaugh: and we'll also need to provide `url`, `verify_ssl`, and `use_auth`, like with the other OSv3-using plugins.

bkabrda:
I added 2 new commits that fix comments by @TomasTomecek and @twaugh. This should be ready to merge now.

bkabrda:
Oh, wait, I forgot to update the specfile with new dependencies. One more commit coming.

twaugh:
LGTM but I guess we'd better wait until https://github.com/projectatomic/atomic-reactor/pull/253 is merged.

bkabrda:
Yup, agreed.

bkabrda:
So, one more commit for https://github.com/projectatomic/osbs-client/issues/190. Sometimes I think that this PR will just live forever :)

twaugh:
LGTM

Issue #164, opened 2015-07-23
osbs uses system:anonymous no matter what username passed in


maxamillion:
I am attempting to pass an username to osbs but it appears to not be using it when calling to the OpenShift REST API. This is using a build off of latest git master as of this afternoon I made with `tito build --rpm --test --builder mock --arg mock=epel-7-x86_64`

```
# osbs --verbose --username system:builder get-user
2015-07-23 16:05:02,344 - osbs - DEBUG - Logging level set to debug
* About to connect() to localhost port 8443 (#0)
*   Trying ::1...
* Connected to localhost (::1) port 8443 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
* skipping SSL peer certificate verification
* NSS: client certificate not found (nickname not specified)
* SSL connection using TLS_RSA_WITH_AES_128_CBC_SHA
* Server certificate:
*       subject: CN=192.168.122.36
*       start date: Jul 14 21:08:46 2015 GMT
*       expire date: Jul 13 21:08:47 2016 GMT
*       common name: 192.168.122.36
*       issuer: CN=openshift-signer@1436908122
> GET /osapi/v1beta3/users/~/ HTTP/1.1
User-Agent: PycURL/7.29.0
Host: localhost:8443
Accept: */*

< HTTP/1.1 403 Forbidden
< Cache-Control: no-store
< Content-Type: application/json
< Date: Thu, 23 Jul 2015 21:05:02 GMT
< Content-Length: 227
< 
* Connection #0 to host localhost left intact
2015-07-23 16:05:02,573 - osbs.core - ERROR - [403] {
  "kind": "Status",
  "apiVersion": "v1beta3",
  "metadata": {},
  "status": "Failure",
  "message": "User \"system:anonymous\" cannot get users at the cluster scope",
  "reason": "Forbidden",
  "details": {},
  "code": 403
}
Traceback (most recent call last):
  File "/usr/bin/osbs", line 9, in <module>
    load_entry_point('osbs==0.14', 'console_scripts', 'osbs')()
  File "/usr/lib/python2.7/site-packages/osbs/cli/main.py", line 353, in main
    args.func(args, osbs)
  File "/usr/lib/python2.7/site-packages/osbs/cli/main.py", line 197, in cmd_get_user
    user_json = osbs.get_user()
  File "/usr/lib/python2.7/site-packages/osbs/api.py", line 28, in catch_exceptions
    return func(*args, **kwargs)
[...]

PR #182, opened 2015-07-29
[do not merge] Try harder when getting http error response body

mmilata:
This fixes showing http error response body in the particular case of #164. I hope I didn't break something else.

@TomasTomecek - The code in http.py is rather tangled. What do you think about getting rid of the curl-requests emulation layer? I think we're not able to use requests anymore anyway.

### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -181,7 +181,7 @@ class Response(object):
                 url = getattr(self.curl, "url", None)
             else:
                 url = None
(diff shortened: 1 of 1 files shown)


TomasTomecek:
http.py needs rewrite, no question; I'm fine with getting rid of requests compat thingy

mmilata:
(obsoleted by #191)

PR #185, opened 2015-07-30
read config lazily

TomasTomecek:
So we don't have
```
2015-07-30 11:24:20,991 - osbs - ERROR - Configuration error: Specified section 'default' not found in '/etc/osbs.conf'
```
all over the place

### DIFF
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -7,6 +7,9 @@ of the BSD license. See the LICENSE file for details.
 """
 from __future__ import print_function, absolute_import, unicode_literals
 
+import logging
+import os
+
 try:
     # py2
     import ConfigParser as configparser
@@ -18,6 +21,9 @@ except ImportError:
 
 from osbs.constants import DEFAULT_CONFIGURATION_FILE, DEFAULT_CONFIGURATION_SECTION, GENERAL_CONFIGURATION_SECTION
 from osbs.exceptions import OsbsException
+
+
+logger = logging.getLogger(__name__)
 
 
 class Configuration(object):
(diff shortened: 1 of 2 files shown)


osbs/conf.py
```diff
@@ -20,6 +23,9 @@
 from osbs.exceptions import OsbsException
 
 
+logger = logging.getLogger('osbs')
```
mmilata: Why not `getLogger(__name__)` like in other modules?

osbs/conf.py
```diff
@@ -43,14 +49,15 @@ def __init__(self, conf_file=DEFAULT_CONFIGURATION_FILE, conf_section=DEFAULT_CO
...
-        try:
-            self.scp.read(conf_file)
-        except (IOError, TypeError):
-            pass
-        else:
-            if not self.scp.has_section(conf_section):
-                raise OsbsException("Specified section '%s' not found in '%s'" %
-                                    (conf_section, conf_file))
+        if conf_file and os.path.isfile(conf_file) and os.access(conf_file, os.R_OK):
+            try:
```
mmilata: I'd suggest removing the try-catch block as we're not expecting `self.scp.read()` to fail now, or do we?

mmilata:
Wrote some nitpicks inline, LGTM otherwise.

osbs/conf.py
```diff
@@ -43,14 +49,15 @@ def __init__(self, conf_file=DEFAULT_CONFIGURATION_FILE, conf_section=DEFAULT_CO
...
+                self.scp.read(conf_file)
+            except (IOError, TypeError):
+                pass
+            else:
+                if not self.scp.has_section(conf_section):
+                    logger.warning("Specified section '%s' not found in '%s'" %
```
twaugh: Putting the landscape.io message here as it's being flaky:
> Specify string format arguments as logging function parameters

TomasTomecek:
27 days? I'm a mess...

rebased, ready for review

mmilata:
Well, it's not like you didn't do anything else;) LGTM->merged

PR #189, opened 2015-08-03
New set_labels_on_build_config() API method (fixes #188)

### LINKED ISSUE #188: Needs set_labels_on_build_config() API
The Atomic Reactor plugin for detecting automated rebuilds needs to be able to set labels on the BuildConfig. See https://github.com/projectatomic/atomic-reactor/issues/249.

twaugh:


bkabrda:
A naming issue: intuitively, I'd say that these methods would reset the labels to precisely the dict that I pass, but instead they do `.update()`. Shouldn't they be called `update_labels...` instead of `set_labels...`?

twaugh:
Possibly. That would be an API change at this point, as `set_labels_on_build()` already existed before this PR. @TomasTomecek, what do you think?

TomasTomecek:
I sort of agree with Slavek. Having `set` and `update` methods would make total sense. In reactor we use `set_annotations`.

twaugh:
Right, we use `set_annotations_on_build()` in atomic-reactor's `store_metadata_in_osv3` plugin, but it's actually an `update()`. Do we want that to be a set, or an update?

I can rename the `set_` methods to `update_`, and introduce new `set_` methods, but should I do that in a separate PR to this?

TomasTomecek:
AFAIK there are no annotations on the build object, so even though it's update, in the end the result is the same as in case of set. Not sure if your chain rebuild work adds some annotations (or if openshift has some native ones).

twaugh:
That's true, and there aren't any native ones set on Builds at the moment.

Should I do this renaming work in a separate PR, or extend this one?

bkabrda:
I'd say there's not big need to create another PR, just add a commit here.

bkabrda:
LGTM

PR #191, opened 2015-08-04
Refactor http module

mmilata:
Blocking and streaming http responses now live in a separate classes.  The blocking response is implemented using the streaming response. Each streaming response has its own curl and curl_multi handle.

The streaming response has to be explicitly closed. The preferred way is to use it as a context manager which ensures that it is closed when exception is raised in the middle of reading the stream. Because it doesn't fit into our current API, the class also tries to free the curl resources when it finishes reading the http stream and also when it's gargabe collected.

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -18,7 +18,7 @@ from osbs import set_logging
 from osbs.api import OSBS
 from osbs.conf import Configuration
 from osbs.constants import DEFAULT_CONFIGURATION_FILE, DEFAULT_CONFIGURATION_SECTION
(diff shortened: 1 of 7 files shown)


mmilata:
Found a bug, please do not review yet.

mmilata:
Ready for review.

@pbabinca can you check if this fixes your curl problem? Please note that due to recent incompatible changes, osbs from git probably won't work with older atomic-reactor build images.

TomasTomecek:
Took a brief look at the code and I gotta say I really like it. Will try to test it out next week.

mmilata:
Rebased on top of master w/ fixes for what @twaugh and @landscape-bot found.

twaugh:
LGTM

PR #194, opened 2015-08-05
Imagestream naming

bkabrda:
This PR makes sure we get three things right:
- `imagestream` argument for `ImportImagePlugin`
- `docker_image_repo` argument for `ImportImagePlugin`
- name of trigger that built image depends on

Also, the second commit fixes minor issue, where openshift url for `check_and_set_rebuild` wasn't previously substituted in `BuildRequest`.

### DIFF
--- a/inputs/simple_inner.json
+++ b/inputs/simple_inner.json
@@ -1,15 +1,6 @@
 {
   "prebuild_plugins": [
     {
-      "args": {
-        "label_key": "is_autorebuild",
-        "label_value": "true",
-        "url": "{{OPENSHIFT_URI}}",
-        "verify_ssl": false
-      },
-      "name": "check_and_set_rebuild"
-    },
-    {
       "name": "pull_base_image",
       "args": {
         "parent_registry": "{{REGISTRY_URI}}",
@@ -29,15 +20,6 @@
         "image_id": "BUILT_IMAGE_ID"
       },
       "name": "all_rpm_packages"
-    },
-    {
-      "args": {
-        "imagestream": "{{IMAGESTREAM}}",
-        "docker_image_repo": "{{DOCKER_IMAGE_REPO}}",
-        "url": "{{OPENSHIFT_URI}}",
-        "verify_ssl": false
-      },
-      "name": "import_image"
     }
   ],
   "exit_plugins": [
--- a/inputs/simple_inner.json
+++ b/inputs/simple_inner.json
@@ -29,6 +29,15 @@
         "image_id": "BUILT_IMAGE_ID"
       },
       "name": "all_rpm_packages"
+    },
+    {
+      "args": {
+        "imagestream": "{{IMAGESTREAM}}",
+        "docker_image_repo": "{{DOCKER_IMAGE_REPO}}",
+        "url": "{{OPENSHIFT_URI}}",
+        "verify_ssl": false
+      },
+      "name": "import_image"
     }
   ],
   "exit_plugins": [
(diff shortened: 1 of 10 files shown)


bkabrda:
The thing I'm not sure about is whether we want to limit osbs-client to Dockerfiles that have `Name` label specified (which this PR effectively does). IMO this is ok, but if someone thinks otherwise, speak up.

twaugh:
I've made a separate pull request for the `check_and_set_rebuild` fix, so that this pull request can be examined independently.

bkabrda:
I'll remote the commit from this PR, then. (Just for reference, the commit is now in PR #195).

twaugh:
Just a note: the BuildConfigs and ImageStreams are both tracking the same units, so they ought to use the same names as each other.

bkabrda:
Done, should be ready for merge.

osbs/build/build_request.py
```diff
@@ -185,6 +185,14 @@ def render(self):
                 self.dj.dock_json_set_arg('postbuild_plugins', "store_metadata_in_osv3",
                                           "use_auth", self.spec.use_auth.value)
 
+        if self.dj.dock_json_has_plugin_conf('postbuild_plugins', 'import_image'):
+            self.dj.dock_json_set_arg('postbuild_plugins', 'import_image', 'imagestream',
+                                      self.spec.imagestream_name.value)
+            self.dj.dock_json_set_arg('postbuild_plugins', 'import_image', 'docker_image_repo',
+                                      self.spec.imagestream_url.value)
+            self.dj.dock_json_set_arg('postbuild_plugins', 'import_image', 'url',
+                                      self.spec.openshift_uri.value)
```
twaugh: I think we also want to set `use_auth` here, don't we? cf. line 182 further up.

inputs/simple_inner.json
```diff
@@ -29,6 +29,15 @@
...
       "name": "all_rpm_packages"
+    },
+    {
+      "args": {
+        "imagestream": "{{IMAGESTREAM}}",
+        "docker_image_repo": "{{DOCKER_IMAGE_REPO}}",
+        "url": "{{OPENSHIFT_URI}}",
+        "verify_ssl": false
+      },
+      "name": "import_image"
```
twaugh: Does the simple build type get all the rebuild trigger bits? Maybe that should only be in the prod type.

osbs/build/spec.py
```diff
@@ -144,8 +147,11 @@ def set_params(self, git_uri=None, git_ref=None, git_branch=None, registry_uri=N
             raise OsbsValidationException("yum_repourls must be a list")
         self.yum_repourls.value = yum_repourls or []
         self.use_auth.value = use_auth
-        self.name.value = '%s-%s' % (self.component.value, self.git_branch.value)
-        self.base_image.value = os.path.join(registry_uri, base_image)
+        self.name.value = name_label.replace('/', '-')
+        self.trigger_imagestream_name.value = get_imagestream_name_from_image(base_image)
+        # imagestream and buildconfig have precisely the same name
+        self.imagestream_name.value = self.name.value
```
twaugh: I'm happy about this now. The `Name` label, with '/' replaced by '-', looks like the correct name to use for the ImageStream.

mmilata: Shouldn't the value of `trigger_imagestream_name` also have the slashes replaced? It looks like `get_imagestream_name_from_image` doesn't do such substitution.

inputs/simple_inner.json
bkabrda: Hmm, not sure. I guess this really depends on how we define simple builds. I'm fine with only having triggers for prod builds, if others agree.

osbs/build/spec.py
bkabrda: @mmilata you're right, I'll fix that.

bkabrda:
I added a commit that addresses issues found in previous commits. Everything should be fine now.

bkabrda:
Good catch. I'll just do a force push to not introduce a new commit...

twaugh:
LGTM

PR #195, opened 2015-08-05
Actually set check_and_set_rebuild plugin url argument

twaugh:


twaugh:
@bkabrda I ran into this today while testing the `bump_release` plugin. The plugin gets further now, but still fails:

```
  File "/usr/lib/python2.7/site-packages/atomic_reactor-1.4.0-py2.7.egg/atomic_reactor/plugin.py", line 183, in run
    plugin_response = plugin_instance.run()
  File "/usr/lib/python2.7/site-packages/atomic_reactor-1.4.0-py2.7.egg/atomic_reactor/plugins/pre_check_and_set_rebuild.py", line 104, in run
    osbs.set_labels_on_build_config(buildconfig, labels, **kwargs)
  File "build/bdist.linux-x86_64/egg/osbs/api.py", line 31, in catch_exceptions
    return func(*args, **kwargs)
  File "build/bdist.linux-x86_64/egg/osbs/api.py", line 386, in set_labels_on_build_config
    namespace=namespace)
  File "build/bdist.linux-x86_64/egg/osbs/core.py", line 371, in set_labels_on_build_config
    namespace=namespace)
  File "build/bdist.linux-x86_64/egg/osbs/core.py", line 339, in adjust_attributes_on_object
    build_json = self._get(url).json()
  File "build/bdist.linux-x86_64/egg/osbs/core.py", line 88, in _get
    headers, kwargs = self._request_args(with_auth, **kwargs)
  File "build/bdist.linux-x86_64/egg/osbs/core.py", line 79, in _request_args
    raise OsbsAuthException("Please check your credentials. "
OsbsAuthException: Please check your credentials. Token was not retrieved successfully.
```

Did you get past this?

mmilata:
I think the stacktrace is caused by multiple underlying issues. Assuming you are using ansible-osbs, there are two supported authentication scenarios:

#### With authentication proxy

In this case `with_auth=True` has to be used. OSBS first tries to obtain OAuth token, which might require valid Kerberos ticket (and `use_kerberos=True`), or valid username/password for basic authentication (and `username=...`, `password=...` set), or having IP whitelisted in the proxy configuration (no additional parameters should be set).

Ansible-osbs currently only supports kerberos auth+ip whitelist.

Because atomic-reactor does not have access to a kerberos ticket, it's allowed to obtain the OAuth ticket based on the IP address. This `<If>` block is responsible for it in the proxy configuration: https://github.com/projectatomic/ansible-osbs/blob/master/roles/kerberos_proxy/templates/httpd-krbproxy.conf.j2#L43

#### Without authentication proxy

Without proxy, `with_auth=False` is needed. OSBS does not try to obtain OAuth ticket before sending requests to OSBS server. If `with_auth=True` is used, OSBS client tries to obtain OAuth ticket and fails.

We need to pass the information about whether OAuth token should be obtained to atomic-reactor. The `store_metadata_in_osv3` plugin has the `use_auth` variable which osbs-client sets based on the `metadata_plugin_use_auth` parameter, which in most circumstances should be set to same value as `with_auth`. It appears that `check_and_set_rebuild` also has this parameter but osbs-client doesn't set it.

#### Proposed solution

1. Set the `use_auth` parameter of the other plugins based on value of `metadata_plugin_use_auth` (and perhaps rename the configuration key?)
2. Add basic auth support to proxy deployed by ansible-osbs to make testing with proxy easier
3. Fix basic auth in osbs-client: projectatomic/osbs-client#196

Regarding 1., the current situation seems a bit messy. I'd be in favour of removing the `metadata_plugin_use_auth` and setting the plugin parameter based on `use_auth` (are there any cases where we don't want that?). Or perhaps always authenticating atomic-reactor using client SSL certificate (would need a bit more work).

twaugh:
Thanks for the analysis. I agree, I think we should remove `metadata_plugin_use_auth` and take the value from `use_auth` instead. I'll start on a new PR for that in #196.

twaugh:
This will need re-basing and adjusting once #197 is merged.

PR #197, opened 2015-08-06
Use value of use_auth for metadata_plugin_use_auth (fixes #196)

### LINKED ISSUE #196: Basic-auth doesn't work
When running osbs-client API functions from atomic-reactor plugins, I'm getting tracebacks like the one that @twaugh posted at https://github.com/projectatomic/osbs-client/pull/195#issuecomment-128051568
It seems that the problem is that osbs-client doesn't send proper `X-CSRF-Token`. For example:
```
# osbs is instance of osbs.api.OSBS
>>> r = osbs.os._get('https://192.168.122.121:8443/oauth/authorize?response_type=token&client_id=openshift-challenging-client', with_auth=False, allow_redirects=False)
>>> r.headers
{'date': 'Thu, 06 Aug 2015 08:18:04 GMT', 'content-length': '77', 'warning': '199 OpenShift "A non-empty X-CSRF-Token header is required to receive basic-auth challenges"', 'content-type': 'text/plain; charset=utf-8'}
```

The `_get` call reproduces what osbs-client does at https://github.com/projectatomic/osbs-client/blob/master/osbs/core.py#L110

twaugh:


bkabrda:
LGTM

PR #198, opened 2015-08-06
Finish renaming to osbs-client, fix old links.

bkabrda:


twaugh:
Not sure how the package will get renamed in Fedora. Will it be another package review?

bkabrda:
Yeah, it has to be a new review request. Don't worry, I'll do the work :) (once we actually release a new tarball called osbs-client).

mmilata:
LGTM, merged!

PR #201, opened 2015-08-07
Allow initalizing kerberos credentials from keytab

mmilata:
Do not merge yet, depends on #191.

Creates (soft) dependency on [python-krbcontext](https://github.com/tkdchen/python-krbcontext) which we'll need to package.

### DIFF
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -57,6 +57,12 @@ Some options are also mandatory.
 
 * `client_cert`, `client_key` (*optional*, `string`) - paths to PEM-encoded client certificate and key to be used for authentication
 
+* `kerberos_keytab` (*optional*, `string`) - location of the keytab that will be used to initialize kerberos credentials - usually in the form `FILE:<absolute_path>`, see [kerberos documentation](http://web.mit.edu/Kerberos/krb5-latest/doc/basic/keytab_def.html) for other possible values
+
+* `kerberos_principal` (*optional*, `string`) - kerberos principal for the keytab provided in `kerberos_keytab`
+
+* `kerberos_ccache` (*optional*, `string`) - location of credential cache to use when `kerberos_keytab` is set (please refer to [kerberos documentation](http://web.mit.edu/Kerberos/krb5-latest/doc/basic/ccache_def.html) for list of credential cache types)
+
 * `registry_uri` (*optional*, `string`) — docker registry URI to use for pulling and pushing images
 
 * `pulp_registry_name` (*optional*, `string`) — name of pulp registry within dockpulp config
--- a/osbs/kerberos_ccache.py
+++ b/osbs/kerberos_ccache.py
@@ -1,0 +1,91 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from __future__ import print_function, absolute_import, unicode_literals
+
+import re
+import os
+import logging
+import datetime
+import subprocess
+
+from osbs.exceptions import OsbsException
+
+logger = logging.getLogger(__name__)
+
+KLIST_TGT_RE = (r"\d\d/\d\d/\d{2,4}"
+                r" +"
+                r"\d\d:\d\d:\d\d"
+                r" +"
+                r"(?P<month>\d\d)"
+                r"/"
+                r"(?P<day>\d\d)"
+                r"/"
+                r"(?P<year>\d{2,4})"
+                r" +"
+                r"(?P<hour>\d\d)"
+                r":"
+                r"(?P<minute>\d\d)"
+                r":"
+                r"(?P<second>\d\d)"
+                r" +"
+                r"krbtgt/(?P<realm>[-.A-Z0-9]+)@(?P=realm)")
+
+def run(cmd, extraenv=None):
+    env = os.environ.copy()
+    if extraenv:
+        env.update(extraenv)
+
+    logger.debug("Subprocess: %s", ' '.join(cmd))
+    # universal_newlines=True causes stdout/stderr to be strings (not bytes) in py3
+    p = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=env,
+                         universal_newlines=True)
+    stdout, stderr = p.communicate()
+
+    return p.returncode, stdout, stderr
+
+def kerberos_ccache_init(principal, keytab_file, ccache_file=None):
+    """
+    Checks whether kerberos credential cache has ticket-granting ticket that is valid for at least
+    an hour.
+
+    Default ccache is used unless ccache_file is provided. In that case, KRB5CCNAME environment
+    variable is set to the value of ccache_file if we successfully obtain the ticket.
+    """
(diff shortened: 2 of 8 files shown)


bkabrda:
There's a slight problem with this - python-krbcontext depends on python-krbV, which is currently not ported to Python 3 and porting it would require significant effort. I know of no ongoing efforts to make python-krbV work on Python 3. I recognize that we need to implement this functionality, but I'm not thrilled by prospect of introducing such a dependency.
It seems that upstream already has an issue for Python 3 support, so maybe you could first make sure that they have a sane plan for this? https://github.com/tkdchen/python-krbcontext/issues/12

mmilata:
Understood. The code relies on python-krbV heavily so I'd assume there's no plan, though I can check with the author.

Let me see if we can use python-gssapi for this. If not then I think we'll have to call `kinit` from the script directly.

mmilata:
Rewritten to invoke `klist`/`kinit` binaries directly.

Should we include the `krb5-workstation` as a dependency or leave it optional since this feature will probably be seldom used?

TomasTomecek:
There is a lot of members of IPA team here at Flock. I asked them what is the best way to get kerberos support into a client and they advised we should use gssapi (which is python 3 friendly). I'll try to get some code.

mmilata:
It would be great if you could get some code or a pointer ... I looked into gssapi briefly but wasn't able to figure out how to initialize the kerberos credentials from keytab so that pycurl can use them (can't say I tried hard, though). Also, python-gssapi seems to be unavailable for rhel6 nor rhel7 currently.

frozencemetery:
mmilata writes:

> Also, python-gssapi seems to be unavailable for rhel6 nor rhel7 currently.

Hi, I'm a python-gssapi developer.  We will probably never see python-gssapi in rhel6 proper because the project is so much newer than rhel6, though since FreeIPA will soon depend on python-gssapi, it may end up in rhel7 eventually.

That said, we do test python-gssapi against the Kerberos versions in both rhel6 and rhel7, so everything should work fine if you build it yourself (and we will fix any bugs if it does not!).

mmilata:
@frozencemetery, thank you for your suggestion. After internal discussion we figured out that what we need - initializing kerberos credential cache - requires just setting the `KRB5_CLIENT_KTNAME` environment variable and gssapi (used by curl) does this automatically. However rhel6 lacks this feature, and I don't think it's feasible for us to maintain our own krb5-libs package.

It was suggested to run kinit from cron which is essentially what this patch does so I guess we can use it after all.

mmilata:
Rebased onto master & ready for review.

twaugh:
There are conflicts... want to re-base?

mmilata:
Rebased again:)

docs/configuration_file.md
```diff
@@ -57,6 +57,12 @@ Some options are also mandatory.
 
 * `client_cert`, `client_key` (*optional*, `string`) - paths to PEM-encoded client certificate and key to be used for authentication
 
+* `kerberos_keytab` (*optional*, `string`) - absolute path to a keytab that will be used to initialize kerberos credentials
+
+* `kerberos_principal` (*optional*, `string`) - kerberos principal for the keytab provided in `kerberos_keytab`
+
+* `kerberos_ccache` (*optional*, `string`) - absolute path to kerberos credential cache to use when `kerberos_keytab` is set (optional)
```
twaugh: Is it a path, or a "location"? Does it have to be implicitly `FILE:...` or can it be e.g. `KEYRING:...`?

mmilata: Well, it's the kerberos location so it can be `KEYRING:...`. From what I tried if the `PREFIX:` is not given it is treated as `FILE:...`. Good point, I'll improve the docs.

osbs/kerberos_ccache.py
```diff
@@ -0,0 +1,89 @@
...
+    # check if we have tgt that is valid more than one hour
+    rc, klist, _ = run(["klist"], extraenv=env)
+    if rc == 0:
+        for line in klist.splitlines():
+            m = re.match(KLIST_TGT_RE, line)
+            if m:
+                # rhel6 has only last two digits
+                year = m.group("year")
+                if len(year) == 2:
+                    year = "20" + year
```
twaugh: Does it list the full year in newer versions?

mmilata: Yep - see the examples in `tests/test_util.py`.

mmilata: Oops ... now that I'm looking at them I didn't get them exactly right.

twaugh:
Could add a couple of tests for subprocess failures?

mmilata:
Will do.

osbs/kerberos_ccache.py
twaugh: On Fedora 22 I'm only seeing 2-digit years.

mmilata: Ah, it's locale-dependent. This means we'll have to run it with `LC_ALL=C`.

mmilata:
Added a commit with fixes. Please let me know when I can squash it with the previous commit.

osbs/kerberos_ccache.py
```diff
@@ -0,0 +1,90 @@
...
+def kerberos_ccache_init(principal, keytab_file, ccache_file=None):
+    """
+    Checks whether kerberos credential cache has ticket-granting ticket that is valid for at least
+    an hour.
+
+    Default ccache is used unless ccache_file is provided. In that case, KRB5CCNAME environment
+    variable is set to the value of ccache_file if we successfully obtain the ticket.
+    """
+    tgt_valid = False
+    env = { "LC_ALL": "C" } # klist uses locales to format date on RHEL7+
```
TomasTomecek: have you also tried rhel6?

mmilata: Yes - on rhel6 klist seems to ignore locales and always uses two-digit year.

TomasTomecek:
LGTM

twaugh:
Looks good to me, you can go ahead and squash.

osbs/kerberos_ccache.py
twaugh: So, klist tries a variety of formats until it finds one that fits within the width it wants. :-/

mmilata: Uhh ... well, if the code fails to parse the date it will run kinit anyway so there should just be a little slowdown at worst.

mmilata:
Squashed.

mmilata:
Hopefully fixed what landscape-bot found.

osbs/kerberos_ccache.py
```diff
@@ -0,0 +1,91 @@
...
+                year = m.group("year")
+                if len(year) == 2:
+                    year = "20" + year
+
+                expires = datetime.datetime(
+                    int(year), int(m.group("month")), int(m.group("day")),
+                    int(m.group("hour")), int(m.group("minute")), int(m.group("second"))
+                )
+
+                if expires - datetime.datetime.now() > datetime.timedelta(hours=1):
```
twaugh: Just mentioning: over in https://github.com/projectatomic/osbs-client/pull/232 I'm looking at using `dateutil` to parse timestamps, but I don't think it would help here as we have to find it first anyway.

twaugh:
LGTM

PR #202, opened 2015-08-09
[do not merge] Enable rebuild triggers

twaugh:
This pull request is for enabling rebuild triggers, and should only be merged once all the plug-ins necessary for them to work are configured.
- [x] `check_and_set_rebuild` plug-in (https://github.com/projectatomic/osbs-client/pull/213)
- [x] `bump_release` plug-in (https://github.com/projectatomic/osbs-client/pull/214)
- [x] `import_image` plug-in (https://github.com/projectatomic/osbs-client/pull/194)
- [x] `sendmail` plugin for failed build notifications (https://github.com/projectatomic/atomic-reactor/pull/282, https://github.com/projectatomic/osbs-client/pull/248)
- [x] `koji_promote` plug-in (https://github.com/projectatomic/atomic-reactor/pull/319, https://github.com/projectatomic/osbs-client/pull/282)

### DIFF
--- a/inputs/prod.json
+++ b/inputs/prod.json
@@ -8,6 +8,17 @@
     }
   },
   "spec": {
+    "triggers": [
+      {
+        "type": "ImageChange",
+        "imageChange": {
+          "from": {
+            "kind": "ImageStreamTag",
+            "name": "{{BASE_IMAGE_STREAM}}"
+          }
+        }
+      }
+    ],
     "source": {
       "type": "Git",
       "git": {
--- a/tests/build/test_build_request.py
+++ b/tests/build/test_build_request.py
@@ -197,6 +197,7 @@ class TestBuildRequest(object):
         authoritative_registry = "registry.example.com"
         distribution_scope = "authoritative-source-only"
         assert isinstance(build_request, ProductionBuild)
+        push_url = 'ssh://git.example.com/git/{0}.git'
         kwargs = {
             'git_uri': TEST_GIT_URI,
             'git_ref': TEST_GIT_REF,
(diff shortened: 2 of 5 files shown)


twaugh:
Question: do we want 'simple'-type builds to have automated rebuilds, or only 'prod'-type builds?

bkabrda:
Originally, I thought we could have autorebuilds for all types of builds, but now I'm starting to see that it may be problematic, cause autorebuilds need a whole bunch of plugins to work etc... I guess noone is actually using simple builds in production, so we may as well drop the autorebuild support for them (and, if someone does want to use them in production, he may as well change the input jsons).

So I'm +1 for dropping autorebuild support in simple builds.

mmilata:
Agree with @bkabrda, I don't see a use case for simple builds with autorebuilds.

twaugh:
I've updated this so that the `bump_release` and `import_image` plugins are removed from prod bulids unless there are triggers present. I think with this change we might be able to merge this (minus the actual trigger JSON) sooner, without needing to wait for all the pieces.

bkabrda:
Note: this carries some significant changes that should be documented in `docs/build_process.md`. (notably, I think we should explain why we change git ref in BuildConfig to git branch).

twaugh:
Agreed.

I've started separating this into bits that can be merged now and bits that need to wait. https://github.com/projectatomic/osbs-client/pull/213 is all for bits that are ready now, and once the `bump_release` plugin in merged in atomic-reactor I'll submit a new PR for enabling/configuring that -- and documenting the branch/commit swap.

Then this PR will be left with just adding the trigger.

lcarva:
@twaugh, given that #640 uses the configuration file in repository to enable autorebuilds, and it's disabled by default, can we safely merge this (after conflicts are resolved)?

twaugh:
Yes, seems like that would be safe.

twaugh:
Closing in favour of #720.

PR #203, opened 2015-08-10
Fix/setup several releasing related things

bkabrda:
- Setup auto releases to PyPI
- Set author and email properly
- Include forgotten `inputs/image_stream.json` into tarball


twaugh:
LGTM

PR #204, opened 2015-08-10
Use compress plugin in prod, change load_squashed_image to load_exported_image

bkabrda:
As a side effect, this would mean that the image copied to NFS would also be gzipped (previously it wasn't compressed at all). I'm not sure if this is desirable. I can turn this off by running the `compress` plugin after the uncompressed squashed image gets copied to NFS. @pbabinca @twaugh @TomasTomecek any ideas if this is ok or if it should be switched back?

### DIFF
--- a/inputs/prod_inner.json
+++ b/inputs/prod_inner.json
@@ -66,6 +66,13 @@
   ],
   "postbuild_plugins": [
     {
+      "name": "compress",
+      "args": {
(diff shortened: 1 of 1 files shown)


twaugh:
I think we do want it compressed, don't we? IIRC that was part of the reason for changing how we compress.

twaugh:
LGTM

PR #208, opened 2015-08-11
Temporarily disable import_image plugin

twaugh:
Will move it to the enable-rebuild-triggers branch.


mmilata:
LGTM & I guess it won't break anything that isn't broken already.

PR #211, opened 2015-08-12
Make sure builds don't get instantiated twice. Fixes #205

### LINKED ISSUE #205: BuildConfigs with triggers don't need to be instantiated?
I'm pretty sure that when a BuildConfig is created with triggers, a Build is automatically instantiated. We should probably have a check for that so we don't end up with two builds when we want one, for the first build submission for a component.

CC @bkabrda

bkabrda:


bkabrda:
Just BTW, there are lots of cases that aren't tested for `_create_build_config_and_build` (and in fact the case being fixed by this PR is not tested either). I'm planning to refactor the tests a bit and would like to send a different PR for it - and after refactoring, I'd send more tests for this function. Hope this is ok.

osbs/api.py
```diff
@@ -166,7 +185,19 @@ def _create_build_config_and_build(self, build_json, namespace):
             # if it doesn't exist, then create it
             logger.debug('build config for %s doesn\'t exist, creating...', build_config_name)
             self.os.create_build_config(json.dumps(build_json), namespace=namespace)
-        return self.os.start_build(build_config_name, namespace=namespace)
+            # under some circumstances, creating a buildConfig might instantiate it, so
+            #  check for an already running build and return it if it exists
+            #  see https://github.com/projectatomic/osbs-client/issues/205
```
twaugh: I was more thinking we'd know before-hand whether we expect a build to be automatically instantiated, i.e. when (a) we're creating a BuildConfig, not updating one, and (b) when there is a trigger present.

Otherwise there might be a race condition: Origin might not have decided to instantiate the build yet, and I don't know whether doing so explicitly will stop it.

bkabrda: My intention was to make this general, to have the code prepared for possible future changes in Openshift, that would change that. IMO it doesn't hurt to have the change general in this way.

My understanding was that creating buildconfig and instantiating build from it was an atomic operation and thus we could afford to do this.

But you're right. If this is not an atomic operation, then we need to know the precise circumstances when the build will be instantiated and wait for it.

twaugh: I don't know whether it's atomic.

bkabrda: Be it atomic or not, I guess it'll be safer to assume that it's not in the long run... I'll create another commit that fixes this.

twaugh:
LGTM

PR #213, opened 2015-08-13
Enable rebuild-related plugins only if there are triggers

twaugh:
This pull request is for adding the `check_and_set_rebuild` and `import_image` plugins only if at least one trigger is specified. It also adds a test case so we can test both with-triggers and without-triggers.


bkabrda:
LGTM

PR #214, opened 2015-08-13
Configure bump_release plugin

twaugh:
This pull request is for adding the `bump_release` plugins only if at least one trigger is specified. It also adds a test case so we can test both with-triggers and without-triggers.


bkabrda:
LGTM

PR #218, opened 2015-08-14
Always return same value type from this method

twaugh:


bkabrda:
Another great catch! LGTM

PR #219, opened 2015-08-14
Fake API: framework changes to allow responses to change over time

twaugh:
The URL to response map has been moved into the `Connection` object, as well as the custom callbacks referenced by it and the lookup function, `get_definition_for()`.

This lookup function now also returns the key in the response map, allowing a custom callback to modify the response for that URL the next time it is requested.

This work prepares the way for https://github.com/projectatomic/osbs-client/pull/202.

### DIFF
--- a/tests/fake_api.py
+++ b/tests/fake_api.py
@@ -28,130 +28,138 @@ except ImportError:
     import urllib.parse as urlparse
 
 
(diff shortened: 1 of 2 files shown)


bkabrda:
LGTM

PR #220, opened 2015-08-16
Set correct BuildConfig names and imageChange ImageStreamTags

### LINKED ISSUE #217: BuildConfig names should be derived from repo name + branch
The use cases is that people want to build from different branches to a single image repository (into different tags of the repository). Moreover, the `Name` label may be the same across more branches, but we want to keep BuildConfigs separate.

twaugh:
Some `FROM` directives may want to specify a particular version of an image, i.e. the part after ':' in the tag name. To accommodate this, the BuildConfig names need to be more specific -- to allow more than one branch to have triggers at a time -- and the ImageStreamTag names used in triggers need to include the whole tag name (still excluding the registry part).

Fixes #217.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -185,7 +185,7 @@ class CommonBuild(BuildRequest):
         self.template['spec']['output']['to']['name'] = tag_with_registry
         if 'triggers' in self.template['spec']:
             self.template['spec']['triggers']\
-                [0]['imageChange']['from']['name'] = self.spec.trigger_imagestream_name.value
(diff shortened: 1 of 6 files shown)


bkabrda:
This looks very good, I'd just suggest to provide a more descriptive name for the `humanish_part` function. Perhaps something like `git_repo_humanish_part`?

twaugh:
Thanks. I've renamed it to `git_repo_humanish_part_from_uri()`.

bkabrda:
LGTM

PR #222, opened 2015-08-19
Refactor tests structure, include tests in dist tarball. Fixes #212

### LINKED ISSUE #212: Improvements to test suite
Currently, the tests are starting to be a huge unreadable blob. I suggest:
- splitting `unit_test.py` into more modules (each module should cover one file from `osbs`)
- using test classes to group together tests for functional classes
- and maybe some more stuff; increasing coverage would also be nice :)

I'll try to work on this when I have some time, but if someone wants to take this sooner, please speak up.

bkabrda:
This PR splits the tests to multiple files/classes and also includes tests in sdist (they will be in tarball, but won't install with `pip install osbs-client`).

This way, tests are better structured. It's now more easy to run tests for only separate part of osbs, e.g.:
- `py.test -k test_http`
- `py.test -k TestHttpSession`

Also, it's much easier to see what is tested and what is not even without having to see coverage report.


tests/build/test_build_request.py
```diff
@@ -0,0 +1,596 @@
+import copy
```
twaugh: No copyright notice.

tests/constants.py
```diff
@@ -7,6 +7,11 @@
 """
 from __future__ import absolute_import, unicode_literals
 
+import os
+
+HERE = os.path.dirname(__file__)
+INPUTS_PATH = os.path.join(HERE, '..', 'inputs')
```
twaugh: Great idea :-)

tests/test_http.py
```diff
@@ -0,0 +1,164 @@
...
+from flexmock import flexmock
+import pycurl
+import pytest
+
+import osbs.http as osbs_http
+from osbs.http import parse_headers, HttpSession, HttpStream
+
+from tests.fake_api import Connection, ResponseMapping
+
+logger = logging.getLogger("osbs.tests")
```
twaugh: Why not `logging.getLogger(__name__)` here?

twaugh:
I've added some comments above.

tests/test_http.py
bkabrda: I don't know, I copy-pasted that from the previous version of tests :) I will fix that.

bkabrda:
Thanks, I addressed the issues in the 2 subsequent commits.

twaugh:
LGTM

PR #224, opened 2015-08-20
simple: don't pull image from provided registry, let users decide

TomasTomecek:
Let's allow users to pull from any registry they will specify in dockerfile.


twaugh:
LGTM

PR #225, opened 2015-08-20
Some spec params moved to prod (fixes #221)

### LINKED ISSUE #221: 'Simple' build type requires build parameters only needed for BuildConfig/ImageStream
Some of the `CommonSpec` build parameters should be specific to `ProdSpec`:

* `git_branch`
* `trigger_imagestreamtag`
* `imagestream_name`
* `imagestream_url`

twaugh:


osbs/api.py
```diff
@@ -280,11 +280,8 @@ def create_simple_build(self, git_uri, git_ref, git_branch, user, component, yum
         build_request.set_params(
             git_uri=git_uri,
             git_ref=git_ref,
-            git_branch=git_branch,
             user=user,
             component=component,
-            base_image=df_parser.baseimage,
-            name_label=df_parser.labels['Name'],
             registry_uri=self.build_conf.get_registry_uri(),
```
bkabrda: So this looks like `create_simple_build` won't even need `df_parser` and it can be removed altogether from that method.

twaugh: True, and the same for the `git_branch` parameter.

twaugh:
Updated.

bkabrda:
LGTM

PR #226, opened 2015-08-20
simple: remove pull_base_image args now we don't set them

twaugh:


twaugh:
Hmm, please hold, not sure this works.

twaugh:
We either have `{{REGISTRY_URI}}` left for the value, which of course fails, or we remove the `parent_registry` and `parent_registry_insecure` arguments. If we remove them, the osv3 input plugin complains:

```
[...] - atomic_reactor.plugins.osv3 - ERROR - source registry is not configured
```

@TomasTomecek, do you think we should just remove that error message?

mmilata:
Hm, I think it changes the current behaviour wrt insecure registries - previously we sent `parent_registry_insecure: true`, now that we don't send it the default is false: https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/pre_pull_base_image.py#L20

We can either leave `parent_registry_insecure: true` in the build json, or, better, make it configurable.

twaugh:
I'm not sure it matters about 'parent_registry_insecure`, because `parent_registry` is no longer valid one way or the other.

mmilata:
Well, the `parent_registry_insecure` value is still used even if no `parent_registry` is supplied - the registry is either taken from the FROM field or the default one is searched. Probably not very intuitive but I'm not sure how else we should do it.

twaugh:
I didn't change this before merge, so currently we aren't supplying `parent_registry_insecure` for simple build profiles. Should I add it back in another PR?

PR #228, opened 2015-08-21
Switch to API version v1 (fixes #223)

### LINKED ISSUE #223: Should use v1 API endpoint
The v1beta3 endpoint will be going away very soon.

twaugh:
Included in this pull request is a change to the `Configuration` class so it defines the API version we code to, as well as as runtime check that the template build JSON uses the same API version.

It also updates the mock JSON files for v1, and drops the 0.5.2 files as v1 support requires at least 0.5.3.

The mock JSON files were generated using a new `--capture-dir` parameter to the command line tool, also included in this pull request.

However, I'm not entirely sure how to generate `get_user.json` and `authorize.txt` so I've carried them forward, changing the `apiVersion` and `selfLink` keys to adjust for v1. I read the upstream API differences between v1beta3 and v1 and didn't see anything that would affect `get_user.json`.

Once the `get_user.json` / `authorize.txt` issue is cleared up, we can probably remove `tests/mock_jsons.sh`.


twaugh:
I've now added a commit for the `get_user.json` files, from actual responses. I'm not sure how to add an identity map so the new JSON files have `"identities": null` instead of the list from the v1beta3 responses.

twaugh:
Re-based.

twaugh:
Squashed the `get_user.json` commit into the mock JSONs commit, and made a small fix to the capture commit.

mmilata:
Good job! Merged.

PR #233, opened 2015-09-01
enable basic auth for metadata plugin

TomasTomecek:
Fixes https://github.com/projectatomic/atomic-reactor/issues/294

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -254,6 +254,8 @@ class OSBS(object):
             yum_repourls=yum_repourls,
             source_secret=self.build_conf.get_source_secret(),
             use_auth=self.build_conf.get_use_auth(),
+            os_username=self.build_conf.get_username(),
+            os_password=self.build_conf.get_password(),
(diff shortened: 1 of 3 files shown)


twaugh:
In the GitHub issue, @mmilata said:

> I'd prefer if we didn't pass the username/password to atomic-reactor - it will get stored in the build logs and build metadata (iirc).

I have a suggestion for how plugins ought to authenticate: they can use the service account's token.
https://docs.openshift.org/latest/dev_guide/service_accounts.html#using-a-service-account-s-credentials-inside-a-container

mmilata:
@twaugh :+1: This way we can also get rid of the auth proxy snippet. This doesn't need the multiple secrets support that will be in 1.0.6, right?

twaugh:
No, it's independent of that.

twaugh:
I've filed #237 to track it.

TomasTomecek:
Sounds good, let's close this then.

PR #236, opened 2015-09-02
Use secrets array with Origin 1.0.6 (fixes #234)

### LINKED ISSUE #234: Configure `push_pulp` plugin to use service account secret
When Origin 1.0.6 is available we could configure the `push_pulp` plugin as described in https://github.com/projectatomic/atomic-reactor/issues/270.

twaugh:
New configuration parameter `openshift_required_version`.
When this is 1.0.6 or later, use the `secrets` array to give pulp its secret.
Otherwise, use the `sourceSecret`.
Rename `source_secret` build parameter to `pulp_secret` but retain compatibility.

### DIFF
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -27,6 +27,8 @@ Some options are also mandatory.
 
 * `verbose` (*optional*, `boolean`) — enable verbose logging
 
+* `openshift_required_version` (*optional*, `string`) — required version to run against (adjusts build template as appropriate)
+
 ### instance options
 
 * `openshift_uri` (**mandatory**, `string`) — root URL where openshift master API server is listening (e.g. `localhost:8443`)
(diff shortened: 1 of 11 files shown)


osbs/build/spec.py
```diff
@@ -159,7 +159,8 @@ class ProdSpec(CommonSpec):
     kojiroot = BuildParam("kojiroot", allow_none=True)
     kojihub = BuildParam("kojihub", allow_none=True)
     image_tag = BuildParam("image_tag")
-    source_secret = BuildParam("source_secret", allow_none=True)
+    pulp_secret = (BuildParam("pulp_secret", allow_none=True) or
+                   BuildParam("source_secret", allow_none=True))
```
mmilata: The `or` here does nothing because the left hand side always evaluates to true ... or am I missing something?

mmilata:
See my comment inline, otherwise LGTM.

osbs/build/spec.py
twaugh: Oh, that's right. Hm.

twaugh:
Should be fixed now.

mmilata:
@twaugh LGTM, want to squash it before merge?

docs/configuration_file.md
```diff
@@ -27,6 +27,8 @@ Some options are also mandatory.
 
 * `verbose` (*optional*, `boolean`) — enable verbose logging
 
+* `openshift_required_version` (*optional*, `string`) — required version to run against (adjusts build template as appropriate)
```
twaugh: Are we happy with this name?

twaugh:
Squashed. I added an inline note, just wanted to double-check the config name makes sense.

docs/configuration_file.md
mmilata: Not exactly happy but couldn't really think of anything better. `openshift_expected_min_version`? `openshift_compat_version`?

mmilata: It would be best not to need such option at all but there doesn't seem to be API endpoint for obtaining the server version.

mmilata:
Merged, thanks!

PR #239, opened 2015-09-03
New API method get_pod_for_build() (fixes #238)

### LINKED ISSUE #238: `get_pod_for_build()` API method
The build pod needs a way to report the docker image ID of the buildroot image in use. That information only seems to be available in the build pod's description, in status.containerStatuses[*].imageID.

Ideally I would be able to get a list of pods having a particular label (`openshift.io/build.name=` my build name) and be able to get the image ID that way.

twaugh:
Querying pods requires a different API endpoint as it is a Kubernetes API rather than specifically Origin, so I've added a new `Openshift` constructor parameter to allow for that.

There is a new `PodResponse` class, which `OSBS.get_pods_for_build()` returns a list of (cf. `list_builds`).

I've added a test case for it, which required some mock JSON. To fetch that, I also added a command line verb for it as well... which ironically has decreased overall test coverage.

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -217,6 +217,19 @@ def cmd_get_user(args, osbs):
         print("Name: \"%s\"\nFull Name: \"%s\"" % (name, full_name))
 
 
+def cmd_get_build_image_id(args, osbs):
+    pod = osbs.get_pod_for_build(args.BUILD_ID[0], namespace=args.namespace)
+    if args.output == 'json':
+        json_output = pod.get_container_image_ids()
+        print_json_nicely(json_output)
+    elif args.output == 'text':
+        format_str = "{tag:18} {image:64}"
+        print(format_str.format(tag='TAG', image='IMAGE ID'), file=sys.stderr)
+        image_ids = pod.get_container_image_ids()
+        for name, image_id in image_ids.items():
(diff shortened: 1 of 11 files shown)


osbs/cli/main.py
```diff
@@ -217,6 +217,23 @@ def cmd_get_user(args, osbs):
...
+    if len(pods) == 0:
+        return
+    assert len(pods) == 1
+    pod = pods[0]
+    if args.output == 'json':
+        json_output = pod.get_container_image_ids()
+        print_json_nicely(json_output)
+    elif args.output == 'text':
+        format_str = "{name:14} {image:64}"
+        print(format_str.format(name='NAME', image='IMAGE'))
```
twaugh: I meant to add `, file=sys.stderr` here for consistency with `list-builds`.

mmilata:
Can one build have multiple pods? If not I'd prefer if the method returned the single pod (or None).

twaugh:
I've changed it so the method returns a single pod. Rather than returning None is there is no single pod to return, I've made it raise an exception for consistency with `get_build()`.

mmilata:
Looks good! Wanna squash it before merging?

twaugh:
Re-based and squashed.

mmilata:
LGTM, merged.

Issue #241, opened 2015-09-07
Incorrect BuildConfig when updating existing one


twaugh:
When submitting a build config that uses (for example) the `squash` plugin, then submitting a new build config with the same parameters but a different build JSON template that does not use that same plugin, the plugin is still present in the BuildConfig.

PR #242, opened 2015-09-07
Fixes for updating BuildConfig

twaugh:
This pull request has fixes for two bugs which combined to cause #241. I was testing against Origin 1.0.5, which perhaps has stricter validation that earlier releases.

The first issue is that we were not checking whether a PUT request for an updated BuildConfig was successful.

The second issue is that `get_imagestreamtag_from_image()` could return an ImageStream-format name (e.g. 'fedora') instead of an ImageStreamTag (eg. 'fedora:latest').

The second issue became apparent once the first was fixed:

```
osbs.exceptions.OsbsResponseException: {
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "BuildConfig \"docker-hello-world-fedora\" is invalid: triggers[0].from.name: invalid value 'fedora', Details: ImageStreamTag object references must be in the form \u003cname\u003e:\u003ctag\u003e",
  "reason": "Invalid",
  "details": {
    "id": "docker-hello-world-rhel-fedora",
    "kind": "BuildConfig",
    "causes": [
      {
        "reason": "FieldValueInvalid",
        "message": "invalid value 'fedora', Details: ImageStreamTag object references must be in the form \u003cname\u003e:\u003ctag\u003e",
        "field": "triggers[0].from.name"
      }
    ]
  },
  "code": 422
}
```

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -199,8 +199,10 @@ class Openshift(object):
 
     def update_build_config(self, build_config_id, build_config_json, namespace=DEFAULT_NAMESPACE):
         url = self._build_url("namespaces/%s/buildconfigs/%s" % (namespace, build_config_id))
-        return self._put(url, data=build_config_json,
-                         headers={"Content-Type": "application/json"})
+        response = self._put(url, data=build_config_json,
+                             headers={"Content-Type": "application/json"})
+        check_response(response)
+        return response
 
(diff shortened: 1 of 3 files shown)


TomasTomecek:
LGTM

PR #243, opened 2015-09-07
Use service account token inside build pod

twaugh:
This allows authentication to OpenShift from within the build pod, by looking for the `/var/run/secrets/kubernetes.io/serviceaccount` path and using it if available.

It only does so if `use_auth` is not set to `False` and no other authentication credentials have been supplied.

This behaviour can be explicitly enabled or disabled by setting `builder_use_auth = false` in the configuration file.

Why might you explicitly enable it? For instance, if the only authorized users use client certificates (so `use_auth` must be off), the plugins will need to authenticate and the only way they can do so is by enabling `builder_use_auth`.

Why might you explicitly disable it? For instance, if the osbs-client version installed in the buildroot image is older than this pull request merge, plugins will not be able to authenticate without additional authentication credentials.

### DIFF
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -81,6 +81,8 @@ Some options are also mandatory.
 
 * `use_auth` (*optional*, `boolean`) — by default, osbs-client is trying to authenticate against OpenShift master to get OAuth token; you may disable the process with this option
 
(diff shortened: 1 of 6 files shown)


mmilata:
Looks good and seems to work fine, merging.

PR #246, opened 2015-09-15
Make sure we remove downloaded git repos after instantiating Dockerfi…

### LINKED ISSUE #200: osbs-client should clean up downloaded repositories
As it is right now, the code that checks out downloaded repositories (in `osbs/util.py`) doesn't explicitly clean up the repository. This is not a problem if osbs-client is invoked in a one-shot call (since the tmpdir gets removed by Python interpreter when it shuts down), but may become a problem in a long-lived process. Assuming a long-lived process would repeatedly use osbs-client, the number of repositories checked out in tmpdir would just grow until consuming all space.
The only thing we really the repository for is creating a `DockerfileParser` from `Dockerfile` inside the repo and then we can throw it away. So I think the best solution would be for `DockerfileParser` to cache the `Dockerfile` content when instantiated. This way, we could download the repo, create `DockerfileParser` instance and then remove the repo, leaving no unused/unmanaged files in tmpdir.
I'll open an issue for caching support at https://github.com/DBuildService/dockerfile-parse and when that's done, I'll implement the bits in osbs-client.

bkabrda:
…leParser. Fixes #200

### DIFF
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -7,8 +7,10 @@ of the BSD license. See the LICENSE file for details.
 """
 from __future__ import print_function, absolute_import, unicode_literals
 
+import contextlib
 import copy
 import os
+import shutil
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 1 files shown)


jpopelka:
LGTM

twaugh:
LGTM

PR #248, opened 2015-09-21
Configure sendmail plugin

bkabrda:
Depends on https://github.com/projectatomic/atomic-reactor/pull/282

### DIFF
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -89,6 +89,10 @@ Some options are also mandatory.
 
 * `pdc_secret` (*optional*, `string`) - name of kubernetes secret to use for plugins communicating with PDC
 
+* `pdc_uri` (*optional*, `string`) - URI of PDC server instance, e.g. `pdc.mycompany.com`
+
+* `smtp_uri` (*optional*, `string`) - URI of SMTP server, e.g. `smtp.mycompany.com`
+
 * `nfs_server_path` (*optional*, `string`) — NFS server and path to use for storing built image (it is passed to `mount` command)
 
 * `nfs_dest_dir` (*optional*, `string`) — directory to create on provided NFS server where image will be stored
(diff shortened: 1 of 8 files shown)


bkabrda:
One more TODO for this PR: set the `submitter` argument for the `sendmail` plugin.

bkabrda:
Everything should be ok now except the uri/url inconsistency. I'm really torn here, since it seems that we use "uri" everywhere in code, but "url" in plugin arguments... Any suggestions on how to best handle this? @TomasTomecek @twaugh

TomasTomecek:
For me, `url` is `https://host.example.com/`, while `uri` is `host.example.com`. Something like what you can see when tagging image before pushing it to registry -- you never put `http` there.

bkabrda:
@TomasTomecek then we have various parts of comments/code wrong:
https://github.com/projectatomic/osbs-client/commit/fd2b0485d7008ad02abd3a67b1f4be48f427ef85#diff-54c8f265fcc32ddf1df97395da4efe19R27 (`openshift_uri="https://localhost:8443/"`)
Just try to `grep -r openshift_uri` in osbs-client and you'll see that we have http/https used multiple times for `openshift_uri`, most of the time in tests.
Also, we never manipulate `openshift_uri` value in code (AFAICS), so it's always uri (no http/https) and then we put it in argument called `url` as it is.

So that's what's making me puzzled and that's what I was trying to cover in my previous comment.

Proposal: perhaps we should, from now on, call everything `uri` and leave the `openshift_uri`->`url` as a historical thing that we don't want to change?

TomasTomecek:
> then we have various parts of comments/code wrong

I'm not surprised :)

> Proposal: perhaps we should, from now on, call everything uri and leave the openshift_uri->url as a historical thing that we don't want to change?

+1

bkabrda:
Hmm, so using URIs instead of URLs everywhere introduces a different problem - one has to manually prepend scheme for both `urljoin` and `requests.get`. Maybe using url everywhere is better. This is getting really confusing, since AFAICS osbs client says it requires openshift uri, but in fact it requires openshift url :)

So... I'll just try to use "url" everywhere and see how that looks

bkabrda:
Ok, I did... something :) Please review

TomasTomecek:
The whole patch set looks fine to me.

twaugh:
LGTM

PR #249, opened 2015-09-21
improve `list-builds`

TomasTomecek:


twaugh:
This looks fantastic!

twaugh:
I like this a lot. Maybe the 'make filtering more generic' part could be more generic though: I can see it being useful to filter on particular values for particular fields.

`osbs list-builds image_id=...` would show the build that produced a given image.

`osbs list-builds user=alice user=bob` would list builds for both alice and bob (same key name so logical OR).

`osbs list-builds state=pending image=alice/image1 image=alice/image2` would list alice's pending builds for image1 and image2 (logical AND and ORs inferred like journalctl does).

What do you think?

TomasTomecek:
Tim, thanks sounds like an awesome idea. I would prefer not to implement it in this PR b/c it looks like a lot of work and that would make this PR open forever. Can you create a new issue please?

twaugh:
The reason I mentioned it now is that I don't want us to have to change the CLI in an incompatible way later.

I suppose it can be made compatible though: if a filter expression contains '=' then it would perform the matching I described above; otherwise it would search for a substring in the image name (i.e. it would be equivalent to adding image=... parameters for each substring that exists).

Just wanted to be clear on where we're going.

twaugh:
Filed enhanced filtering as #252.

TomasTomecek:
Ready for merge from my PoV.

twaugh:
https://github.com/twaugh/osbs-client/commit/c2351a85b6de34919162be9373fd2b8387417d7e has a test to verify the output of `TablePrinter` using `capsys`.

TomasTomecek:
@twaugh very nice!!

twaugh:
Feel free to cherry-pick it.

TomasTomecek:
I thought you already pushed that in this branch.

(wow, doesn't work on 3, since it looks like that python 3 division outputs floats, not ints; will fix)

twaugh:
Looks like you need `//` instead of `/`.

TomasTomecek:
indeed so, will use that instead

twaugh:
LGTM

PR #250, opened 2015-09-21
Fixes for problems spotted by landscape.io

twaugh:
Includes fixes for:
 * OSBS.update_labels_on_build() didn't return any value
 * OSBS.create_image_stream() didn't use namespace argument
 * CLI capture used wrong filenames
 * bad openshift_required_version config would result in wrong error

### DIFF
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -113,13 +113,7 @@ class Configuration(object):
                                     "openshift_required_version",
                                     can_miss=True)
         if verstring:
-            try:
-                openshift_required_version = [int(x)
-                                              for x in verstring.split('.')]
-            except ValueError:
-                pass
-
-            return openshift_required_version
+            return [int(x) for x in verstring.split('.')]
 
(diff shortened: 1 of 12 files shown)


jpopelka:
LGTM

osbs/conf.py
```diff
@@ -113,13 +113,7 @@ def get_openshift_required_version(self):
...
                                     can_miss=True)
         if verstring:
-            try:
-                openshift_required_version = [int(x)
-                                              for x in verstring.split('.')]
-            except ValueError:
-                pass
-
-            return openshift_required_version
+            return [int(x) for x in verstring.split('.')]
```
TomasTomecek: what if `verstring = "1.2.3-dev"

twaugh: Then it will fail with ValueError, whereas previously it would have failed with NameError.

Maybe it should use `pkg_resources.parse_version()` instead.

PR #232, opened 2015-08-26
Use dateutil to parse ISO8601 timestamps with Python 3

twaugh:
Provides robustness against changes in timestamp format.

Unfortunately I couldn't find a way to get this working with Python 2.6.

PR #251, opened 2015-09-21
Spell RFC3339 correctly

twaugh:
Let's at least spell RFC3339 correctly even if we can't get it working for any timezone but UTC (https://github.com/projectatomic/osbs-client/pull/232).


mmilata:
Merged, thanks:)

PR #253, opened 2015-09-22
Use pkg_resources.parse_version() for handling version numbers

twaugh:
The reason for this PR is to demonstrate a fix for the problem identified in https://github.com/projectatomic/osbs-client/pull/250#discussion_r40073359.

Does it make sense to do it this way?


TomasTomecek:
I like this.

twaugh:
Re-based against master.

PR #255, opened 2015-09-22
Add support for showing layering metadata (fixes #254)

### LINKED ISSUE #254: Show layering metadata
https://github.com/projectatomic/atomic-reactor/pull/331 adds layering metadata to builds. We should show that in the output when requested.

twaugh:
New column names: base_image, base_image_id, and image_id.


TomasTomecek:
...and I was just thinking that we should consider adding those new values to `osbs-client` and... it's already done. Good job, you!

twaugh:
Easy to do now it can format columns :-)

PR #258, opened 2015-09-23
utils: improve checkout_git_repo

TomasTomecek:
* get rid of .PIPE -- why pipe, when no one reads it?
 * since git removes empty dir if clone fails (which ends up as OSError
   from shutil.rmtree), we should clone inside a new dir in temp
   directory

### DIFF
--- a/osbs/cli/render.py
+++ b/osbs/cli/render.py
@@ -9,11 +9,9 @@ from __future__ import print_function, absolute_import, unicode_literals
 
 import sys
 import logging
-from subprocess import CalledProcessError
-try:
-    from subprocess import check_output
-except ImportError:
-    from osbs.utils import backported_check_output as check_output
+
+from osbs.exceptions import OsbsException
+from osbs.utils import run_command
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


bkabrda:
Not 100 % sure ATM, but if you don't use PIPE, the command's stdout/stderr will be redirected to osbs-client stdout/stderr... or so I seem to recall :)

TomasTomecek:
Yes, that's my point. Because when you pipe them and no one reads from the pipe, you are not able to access the output and therefore has no idea what's going on when something breaks.

To be honest, I would prefer piping into osbs' logger. Maybe we could create wrapper function with such functionality.

bkabrda:
Ok, now I get it. Then I'd suggest adding the output to the raised OsbsException (which is raised in case CalledProcessError is raised by the git subprocess).

twaugh:
I think CalledProcessError has an 'output' attribute or something like it, populated from .PIPE. Doesn't it?

bkabrda:
Yes, exactly: https://docs.python.org/2/library/subprocess.html#subprocess.CalledProcessError

TomasTomecek:
New in version 2.7.

:/

Will do.

TomasTomecek:
rebased

bkabrda:
LGTM

PR #260, opened 2015-09-24
Pause/resume scheduling of new builds

twaugh:
This admin function can be useful when maintenance needs to be performed on external systems such as Pulp, a Docker registry, NFS server, etc.

For 'pause-builds' the resource quota for the number of running pods is temporarily set to 0, preventing new builds from being scheduled, and running builds are left to finish.

For 'resume-builds' the resource quota of 0 running pods is lifted, leaving things the way they were before the pause and allowing new builds to be scheduled.


TomasTomecek:
This is mindblowing!

twaugh:
It is?! :-)

Let me see about adding some tests.

mmilata:
Looks good & works fine -> merging.

PR #261, opened 2015-09-25
Capture fixes

twaugh:
Here are some fixes for using the `--capture` option.

`response.json()` can raise a status exception, but the calling code won't be expecting it. Instead, decode the JSON from the content directly.

For Kubernetes endpoints, the API begins differently ("api" instead of "oapi"). Trim the beginning of those URLs as well as OpenShift ones.


mmilata:
LGTM, merged.

PR #265, opened 2015-09-25
conf: split usage of registry_uri and openshift_uri

### LINKED ISSUE #259: Configuration: some values are used on multiple places
...which makes testing (and some deployments) pretty much impossible. E.g. `openshift_uri`, `registry`. We should be able to override/change those for each specific usage

TomasTomecek:
* rename openshift_uri -> openshift_url (with deprecation warning)

Fixes #259


osbs/conf.py
```diff
@@ -270,6 +280,10 @@ def get_builder_use_auth(self):
                                default=self.get_use_auth(),
                                is_bool_val=True)
 
+    def get_builder_openshift_url(self):
+        """ url of OpenShift where builder will connect """
+        return self._get_value("builder_openshift_url", self.conf_section, "builder_openshift_url")
```
mmilata: Should we perhaps fall back to the value of `openshift_url` if `builder_openshift_url` is not defined?

mmilata:
Can you please document the changes in `docs/configuration_file.md`?

TomasTomecek:
@mmilata I knew I forgot something! Thanks

TomasTomecek:
done

osbs/conf.py
TomasTomecek: Okay, I think that's reasonable.

TomasTomecek:
fallback comment resolved

mmilata:
Merged, thanks!

PR #266, opened 2015-09-25
Make implicit labels conform to new standard

mmilata:
`Architecture` -> `architecture`
`Vendor` -> `vendor`
`Build_Host` -> `com.redhat.build-host`
`Authoritative_Registry` -> `authoritative-source`
New label: `distribution-scope`

Related:
* https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md
* projectatomic/atomic-reactor#116
* projectatomic/atomic-reactor#337


TomasTomecek:
looks fine

twaugh:
How will the atomic-reactor plugins know which set is in use? `add_dockerfile`, `assert_labels`, `bump_release`, and `tag_by_labels` all use labels from the old set.

mmilata:
Oops ... I wanted to change the `assert_labels` after osbs changes are deployed but forgot about the other plugins. We'll either have to support both sets for some time or convert old set to new set in one of the plugins. Which approach is better?

TomasTomecek:
Conversion could work, since most labels were just renamed anyway.

twaugh:
So `assert_labels` could have a module-level function that other plugins could call (like with `is_rebuild()`), to see which set of labels they should use.

mmilata:
I implemented a simple label conversion in projectatomic/atomic-reactor#342. The module-level function approach seems a bit too complicated for something that we'll (hopefully) only use temporarily.

TomasTomecek:
So plugins should now access `labels["name"]` instead of `labels["Name"]`, right?

Let's wait with merge for Tim.

mmilata:
Correct.
Yes, I'd also prefer if Tim looked at it.

mmilata:
See projectatomic/atomic-reactor#345 for the commit that switches to the new labels.

mmilata:
Damn, `add_labels_in_dockerfile` cannot be run before `bump_release`, because

* `bump_release` creates new commit therefore invalidating the value of the `vcs-ref` label, and
* in situation when Dockerfile has the old `Release` label, `add_labels_in_dockerfile` would create new `release` label, then `bump_release` would increment only the latter resulting in two labels with different values.

I'm thinking of running `add_labels_in_dockerfile` just after `bump_release` and modifying `bump_release` to use the best label available, i.e. `release`, or `Release` if `release` is not present.

TomasTomecek:
looks like we need `plugin_utils` for stuff like
```python
def get_label_name():
  labels.get("name", labels.get("Name"))
```

twaugh:
I agree with @TomasTomecek, it would be good to put that stuff in a single place.

mmilata:
Rebased on top of current master. I think that the label aliases in projectatomic/atomic-reactor#345 provide the single place for old&new labels. Probably doesn't make sense to merge this PR before the atomic-reactor one is merged.

TomasTomecek:
LGTM, let's merge this after the AR one

mmilata:
projectatomic/atomic-reactor#345 has been merged & this has been rebased on top of master. Please review.

twaugh:
LGTM

jpopelka:
LGTM

PR #272, opened 2015-10-13
Configure tag_and_push plugin

mmilata:
OSBS counterpart of projectatomic/atomic-reactor#349 (issue projectatomic/atomic-reactor#348).

### DIFF
--- a/inputs/prod_inner.json
+++ b/inputs/prod_inner.json
@@ -96,7 +96,12 @@
       "name": "tag_by_labels"
     },
     {
-      "name": "tag_and_push"
+      "name": "tag_and_push",
+      "args": {
+        "registries": {
+          "{{REGISTRY_URI}}": { "insecure": true }
+        }
+      }
     },
     {
       "name": "pulp_push",
--- a/inputs/simple_inner.json
+++ b/inputs/simple_inner.json
@@ -14,6 +14,14 @@
   ],
   "postbuild_plugins": [
     {
+      "name": "tag_and_push",
+      "args": {
+        "registries": {
+          "{{REGISTRY_URI}}": { "insecure": true }
+        }
+      }
+    },
+    {
       "args": {
         "image_id": "BUILT_IMAGE_ID"
       },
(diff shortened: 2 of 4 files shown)


bkabrda:
It'd be great to provide the registry from `spec.registry_uri` by explicitly adding it to `tag_and_push` `registries` argument with a special value and then replacing the special value if it's there. For example:

```
"registries": {
  "{{REGISTRY_URI}}": {"insecure": true},
  "some.other.registry": {"insecure": false}
}
```

This way user can:
- specify if the default registry should be used at all
- specify explicitly if the registry connection should be secure or not
- add more registries (IIUC the way it works now is that the code in `build_request.py` would just wipe out anything that was provided in `prod_inner`)

twaugh:
@bkabrda Is that for configuring a v2 registry and/or a v1 registry, or a separate idea?

We'll need to think of a way of configuring a v2 registry soon, as well as a way for osbs-client to know which registries support which API version.

bkabrda:
A separate idea, just to make configuring the plugin easier.

In term of V1/V2, I *think* we don't need to configure anything, since Docker client will first try pushing to V2 at given URI, then fallback to V1 and then fail if none is found. Therefore we just need to provide registry URIs and Docker client should do "the right thing" depending on which version of registry runs on the URI.

(I think we don't need to cover the case where both registries would be running on the same URI. I'm not sure that's even possible...)

twaugh:
Well, we do need to configure which registries to try pushing to, because that's what the `registry_api_versions` parameter is for.

bkabrda:
Hmm, that's right. So how about this:

```
"registries": {
  "foo": {
    "insecure": true,
    "version": 1
  },
  "bar": {
    "insecure": true,
    "version": 2
  }
}
```

Then the client would remove registries of versions that are not in `registry_api_versions`.

twaugh:
Yes, that looks good. And the `tag_and_push` plugin can just ignore `version`.

mmilata:
@bkabrda Updated with your suggestion.

bkabrda:
LGTM. @twaugh are you ok with merging or do you have additional comments?

twaugh:
LGTM

mmilata:
Did one cosmetic change in tests, sorry.

bkabrda:
Still LGTM, merging :+1:

mmilata:
Thanks:)

PR #273, opened 2015-10-14
Fix updating BuildConfigs. Fixes #244

### LINKED ISSUE #244: Updating BuildConfig does not remove e.g. secrets
When updating an existing BuildConfig, if the existing build JSON described a secret for pulp to use and the new build JSON does not, the secret remains in the BuildConfig after it is updated.

Shouldn't it be removed?

Slightly related to #241, discovered during investigation.

(@bkabrda)

bkabrda:


twaugh:
I get '409 Conflict' with this.

```json
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "buildconfig \"docker-hello-world-master\" cannot be updated: the object has been modified; please apply your changes to the latest version and try again",
  "reason": "Conflict",
  "details": {
    "name": "docker-hello-world-master",
    "kind": "buildconfig"
  },
  "code": 409
}
```

Do we need to whitelist metadata.resourceVersion or something?

bkabrda:
Yeah, we'll need to whitelist something. Let me rewrite the patch and test that it actually works...

bkabrda:
Basically, I think we just need to do the `pop` for non-existent keys in `new` while inside `strategy`. That should be general enough to not tie us to a specific openshift version and at the same time it should do all the changes we need (since the original problem can only happen for leftover configuration values for plugins, IIUC)

bkabrda:
Hopefully the second commit does the trick (worked for me while testing).

twaugh:
LGTM and works fine. Thanks!

PR #276, opened 2015-10-15
Use registry_api_versions to support v1 and/or v2

twaugh:
Use `registry_api_versions` to adjust build config as appropriate.


twaugh:
Ready for review.

TomasTomecek:
LGTM
(I feel like we are getting to a point, where we should lose the `ini` conf and do json instead)

twaugh:
Or YAML, yes. I would love that. The INI format is so restrictive.

twaugh:
Re-based against master.

PR #277, opened 2015-10-16
Extract v2 digests from BuildResponse

mmilata:
Related to projectatomic/atomic-reactor#339 / projectatomic/atomic-reactor#352.

Simply returns the list of the JSON objects, like the other methods that return pieces of `BuildResponse`. I can rework it to return list of actual python object that can e.g. return the string that can be used to docker-pull the image by its digest, if desired.

### DIFF
--- a/osbs/build/build_response.py
+++ b/osbs/build/build_response.py
@@ -161,3 +161,8 @@ class BuildResponse(object):
     def get_base_image_name(self):
         return graceful_chain_get(self.get_annotations_or_labels(),
                                   "base-image-name")
+
+    def get_digests(self):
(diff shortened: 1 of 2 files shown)


TomasTomecek:
For my information: why is it a list?

TomasTomecek:
LGTM

mmilata:
Correct, different tag means different digest, at least in the currently implemented v2. Might change with the v2.1/v3/? that's being discussed.

TomasTomecek:
Don't think so. It looks like that manifest will most likely stay. But we'll never know.

PR #280, opened 2015-10-20
Read the source_registry_uri from config/arguments

mmilata:
source_registry_uri introduced in 085e36b
Related to #259.


twaugh:
LGTM

twaugh:
Re-based.

PR #281, opened 2015-10-21
Add document about creating backups

bkabrda:


docs/backups.md
```diff
@@ -0,0 +1,23 @@
+# OSBS Backups
+
+## Creating Backups
+
+Creating OSBS backups is very simple. To backup all relevant data, one needs to:
+
+1. Pause OSBS by `osbs pause-builds`
+2. Wait for all running builds to finish
```
twaugh: `osbs pause-builds` already performs step 2. Do we just need to say "..., which will prevent new builds being scheduled and wait for running builds to finish" in step 1?

docs/backups.md
```diff
@@ -0,0 +1,23 @@
...
+3. Backup data using following commands.
+  * `oc export -o=json BuildConfig > buildconfigs.json`
```
twaugh: Nice! Although I tend to think `-o json` or `--output=json` looks better than `-o=json`. Is it just me?

mmilata: :+1:

docs/backups.md
bkabrda: You're right, I'll fix that.

docs/backups.md
bkabrda: I'll probably use `--output=json` to make the commands explicit, I guess that's best for documentation.

bkabrda:
Good catch, I just blindly rewrote these and didn't see that. I'll push --force this, so that we don't get bazilion unnecessary commits...

twaugh:
LGTM

PR #284, opened 2015-10-22
Validate git_branch and git_ref when triggers are enabled (fixes #271)

### LINKED ISSUE #271: Validate hash/branch
See https://github.com/projectatomic/atomic-reactor/issues/343, which is actually an osbs-client issue.

twaugh:


mmilata:
LGTM, merged!

osbs/utils.py
```diff
@@ -95,6 +96,10 @@ def checkout_git_repo(git_uri, git_ref, git_branch):
         shutil.rmtree(tmpdir)
 
 
+def looks_like_git_hash(git_ref):
```
TomasTomecek: The name made me me giggle :D

PR #285, opened 2015-10-23
spec: changes sugested in package review

jpopelka:
RHBZ#1273883


mmilata:
LGTM:) Merging.

I wonder if we should also run our test suite in the `%check` phase?

PR #286, opened 2015-10-28
Reduce complexity in `osbs.build.build_request`

twaugh:
The `build_request.py` module has some pretty large `render()` methods. This pull requests breaks them up into manageable chunks so the overall process is easier to read and understand.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -177,10 +177,7 @@ class CommonBuild(BuildRequest):
         logger.debug("setting params '%s' for %s", kwargs, self.spec)
         self.spec.set_params(**kwargs)
 
-    def render(self):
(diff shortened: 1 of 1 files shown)


TomasTomecek:
This is very nice! (and also hard to review -- I vote for merging if tests pass)

twaugh:
What I've done is move plugin-specific configuration out into separate methods with names including the plugin they are for, as well as moving out other related config updates such as resource limits.

This means the main `render()` method becomes much easier to read, as you can see very quickly which plugins are configured, and how that relates to other config changes.

I'll re-based it.

mmilata:
Nice, merged!

twaugh:
Thanks!

TomasTomecek:
> What I've done is move plugin-specific configuration out into separate methods

Very well done. This was totally needed.

PR #287, opened 2015-10-29
Enable metadata-only Content Generator imports when not using v1

twaugh:


mmilata:
LGTM, merged. Haven't really tried running any prod builds with the recent PRs, hope that's OK.

PR #289, opened 2015-10-29
Configure pulp_sync plugin

twaugh:
If we are using v1 we want to use `pulp_push`, otherwise we want `pulp_sync`. Only one of those plugins should be enabled, not both.


mmilata:
LGTM

PR #290, opened 2015-10-29
Check that at least one registry is provided

### LINKED ISSUE #268: validate registry_uri.value
```
Original traceback (most recent call last):
  File "osbs-client/osbs/api.py", line 33, in catch_exceptions
    return func(*args, **kwargs)
  File "osbs-client/osbs/api.py", line 285, in create_prod_build
    git_push_username=self.build_conf.get_git_push_username(),
  File "osbs-client/osbs/build/build_request.py", line 264, in set_params
    self.spec.set_params(**kwargs)
  File "osbs-client/osbs/build/spec.py", line 236, in set_params
    name_label)
  File "/usr/lib64/python2.7/posixpath.py", line 70, in join
    elif path == '' or path.endswith('/'):
AttributeError("'NoneType' object has no attribute 'endswith'",)

```

mmilata:
Closes #268.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -254,10 +254,13 @@ class CommonBuild(BuildRequest):
         self.template['spec']['source']['git']['uri'] = self.spec.git_uri.value
         self.template['spec']['source']['git']['ref'] = self.spec.git_ref.value
 
-        primary_registry_uri = self.spec.registry_uris.value[0].uri
-        tag_with_registry = '{0}/{1}'.format(primary_registry_uri,
-                                             self.spec.image_tag.value)
-        self.template['spec']['output']['to']['name'] = tag_with_registry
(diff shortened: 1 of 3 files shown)


twaugh:
Does this make sense for v1 though? In the v1 workflow we don't use the registry when pulp is in use do we?

mmilata:
Good point. It seems we always use it (even if a-r won't push there): https://github.com/projectatomic/osbs-client/blob/master/osbs/build/build_request.py#L257

I'll check if OpenShift can handle having no output registry, then we can set it only if there's at least one registry provided.

mmilata:
Never mind, we already do that when using pulp. I'll fix the PR ...

mmilata:
Updated.

twaugh:
Should we test it too?

mmilata:
Added test and rebased.

twaugh:
LGTM

PR #296, opened 2015-11-05
Add %check section and build also for Python3

jpopelka:
Looks like I turned Python3 builds off in 93d5854 (don't ask me why) and nobody has objected so far.
Let's turn them on again.

It'd also be nice to run tests when building the packages, so here's a `%check` section.
There are however some packages (e.g. `flexmock`) missing on `rhel`/`epel` so we can run the tests on Fedora only.
We also have to skip (if there's no internet connection) tests that try to access `httpbin.org`, because there's no connection in `koji` build environment. Here's example scratch-build [build.log](https://kojipkgs.fedoraproject.org//work/tasks/7898/11717898/build.log)

### DIFF
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -13,6 +13,7 @@ import pytest
 
 import osbs.http as osbs_http
 from osbs.http import parse_headers, HttpSession, HttpStream
+from osbs.exceptions import OsbsNetworkException
 
 from tests.fake_api import Connection, ResponseMapping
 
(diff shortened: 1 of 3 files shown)


tests/test_http.py
```diff
@@ -24,6 +25,15 @@ def s():
     return HttpSession(verbose=True)
 
 
+def has_connection():
+    # In case we run tests in an environment without internet connection.
+    try:
+        HttpStream("https://httpbin.org/get", "get")
+        return True
+    except OsbsNetworkException:
+        return False
```
twaugh: Nice! I'd been thinking of `-k not http` for that but this is automatic, much better idea.

twaugh:
LGTM

TomasTomecek:
LGTM

jpopelka:
rebased & merged

PR #297, opened 2015-11-05
Fix registry URI version parsing

twaugh:
The '/v2' suffix was not being taken into account correctly because the regular expression for stripping off the scheme from the URI was also stripping off the resource.

### DIFF
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -171,8 +171,8 @@ class CommonSpec(BuildTypeSpec):
         def ditch_http_prefix(val):
             if not val:
                 return val
-            # We only want the hostname[:port]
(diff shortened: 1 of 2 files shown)


jpopelka:
LGTM

TomasTomecek:
LGTM
sidenote: will this work correctly if someone is running registry on `registry.example.com:5000/my/custom/handle` so e.g. you can pull images like `docker pull registry.example.com:5000/my/custom/handle/image:v0.1`

twaugh:
No, it won't. I don't think you can even run the API that way.

TomasTomecek:
I think you're right, docker would have no idea, what's image name and what's uri suffix. Lgtm.

twaugh:
Re-based for merge.

PR #303, opened 2015-11-06
Spec macros

jpopelka:
These are not mandatory, but I needed to add them into `atomic-reactor.spec`, so to be consistent.


TomasTomecek:
LGTM, very nice!

jpopelka:
rebased & merged

PR #304, opened 2015-11-09
pulp_sync requires scheme part of docker registry URI

twaugh:
Pulp requires the scheme part of the docker registry URI, and so `pulp_sync` needs to be configured with it.

Don't strip out the scheme part when creating build parameters.
Strip out the scheme part for the source registry URI.
Strip out the scheme part for `tag_and_push` configuration.
Strip out the scheme part for `spec.output.to.name`.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -11,7 +11,6 @@ import json
 import logging
 import os
 from pkg_resources import parse_version
-import re
 
 try:
     # py2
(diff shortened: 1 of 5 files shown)


jpopelka:
LGTM

osbs/build/build_request.py
```diff
@@ -37,6 +38,15 @@ def register_build_class(cls):
     return cls
 
 
+def ditch_http_prefix(val):
+    if not val:
+        return val
+    # We don't want the scheme
+    return re.sub(r'^https?://(.*)$',
+                  lambda m: m.groups()[0],
+                  val)
```
mmilata: What about moving this code into the `RegistryURI` class? It would have three attributes - `uri`, `version` and something like `hostport` that would be used in places where `ditch_http_prefix(...)` is now.

twaugh: I've added a commit to do that. See what you think.

mmilata:
Oops ... I overlooked that we need to parse the source registry as well. I like the commit but would prefer if we used the same code to handle both source and target registries.

The last commit tries to do that - feel free to modify/drop it.

mmilata:
Updated to allow no source registry & addressed landscape issues.

twaugh:
Related: https://github.com/projectatomic/osbs-client/issues/308#issuecomment-156176676

mmilata:
Ready to merge?

twaugh:
Yes.

PR #293, opened 2015-11-02
[do not merge] Allow separate pulp instances for pulp_push and pulp_sync

twaugh:
Allow different pulp instances for `pulp_push` and `pulp_sync`. New configuration parameters `pulp_sync_secret` and `pulp_sync_registry_name` allow separate configuration for `pulp_sync`
plugin. If not set, it will use the `pulp_secret` and `pulp_registry_name` values.

PR #306, opened 2015-11-11
Allow setting secret for pulp_sync in sourceSecret

mmilata:
Whenever OpenShift < 1.0.6 is used.

This is osbs counterpart of projectatomic/atomic-reactor#370.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -348,8 +348,8 @@ class ProductionBuild(CommonBuild):
                 logger.debug("not setting secret for unused plugin %s",
                              plugin[1])
 
-        elif plugin[1] == 'pulp_push':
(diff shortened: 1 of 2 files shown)


twaugh:
I think we should probably verify that `sourceSecret` is not already set to some other value otherwise we'll overwrite `pulp_push`'s secret with `pulp_sync`'s if they are different. If they are different but we have to use `sourceSecret`, we should raise `OsbsValidationException` because we can't support more than one secret that way.

twaugh:
...and now I remember that they can't have different secrets unless https://github.com/projectatomic/osbs-client/pull/293 is merged, which it probably won't be.

So LGTM :-)

mmilata:
It won't hurt to add the check, just in case.

twaugh:
LGTM, merging.

PR #313, opened 2015-11-18
Use correct imagestream_url (fixes #308)

### LINKED ISSUE #308: ImageStream URL is first listed registry URI, should be source_registry_uri
When configuring the `import_image` plugin to create an ImageStream, it is told to fetch tags from the (first listed) registry URI.

In fact I think we want this to be the source registry URI. When using pulp, for instance, we want to fetch the repository tags back via crane and use that to trigger builds.

More specifically, I think it should be: the source registry URI, or if that is not set then it should be the first listed registry URI.

twaugh:


mmilata:
LGTM, merging.

PR #314, opened 2015-11-19
Don't request stop_autorebuild_if_disabled or sendmail plugins unless there are triggers

twaugh:


twaugh:
The `sendmail` plugin also needs to be removed. Working on that.

twaugh:
Investigating failure.

mmilata:
LGTM, merged.

PR #316, opened 2015-11-19
Revert implicit labels to the old standard

mmilata:
Partially reverts 36e1bef6c85f685484cc1c07db524555bebfc528. Related to #266.

When new labels were used with old a-r the result was mix of old (provided by a-r/Dockerfile) and new (provided by osbs-client) labels.

During the migration to the new set, we should:
- use the old set in osbs-client
- use the alias feature of atomic-reactor to add new-style alias for the
  old label.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -471,15 +471,15 @@ class ProductionBuild(CommonBuild):
 
     def render_add_labels_in_dockerfile(self):
         implicit_labels = {
-            'vendor': self.spec.vendor.value,
-            'com.redhat.build-host': self.spec.build_host.value,
-            'authoritative-source': self.spec.authoritative_registry.value,
+            'Vendor': self.spec.vendor.value,
+            'Build_Host': self.spec.build_host.value,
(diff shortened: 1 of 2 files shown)


twaugh:
That looks like a good approach.

If the base layer doesn't provide any of the implicit labels, osbs-client will add them with the old names and atomic-reactor will alias them to the new names.

If the base layer provides the implicit labels under the old names, osbs-client will only add whichever ones are missing, and atomic-reactor will alias all of them to the new names.

So the only issue would be how atomic-reactor handles the base layer only providing the implicit labels under the old names. I'm not clear on what will happen in that situation.

mmilata:
I think it should work fine even in the third case: https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/pre_add_labels_in_df.py#L131

I.e. the label provided by the base image will be `all_labels[new]`, `df_labels[new]` won't exist so on line 141 the new label will be added with the value of `all_labels[old]` which will have the value provided by osbs-client (line 127).

twaugh:
Thanks, I see now.

LGTM

PR #319, opened 2015-11-30
Add backup/restore functionality

mmilata:
Can be used for migrating OSBS instances.

Please test?

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -12,6 +12,9 @@ import json
 import logging
 
 from os import uname
+import codecs
+import time
+import os.path
 import sys
 import argparse
 from osbs import set_logging
@@ -19,10 +22,10 @@ from osbs.api import OSBS
 from osbs.cli.render import TablePrinter
 from osbs.conf import Configuration
 from osbs.constants import (DEFAULT_CONFIGURATION_FILE, DEFAULT_CONFIGURATION_SECTION,
-                            CLI_LIST_BUILDS_DEFAULT_COLS, PY3)
+                            CLI_LIST_BUILDS_DEFAULT_COLS, PY3, BACKUP_RESOURCES)
 from osbs.exceptions import OsbsNetworkException, OsbsException, OsbsAuthException, OsbsResponseException
 from osbs.cli.capture import setup_json_capture
-from osbs.utils import strip_registry_from_image
+from osbs.utils import strip_registry_from_image, paused_builds, TarReader, TarWriter
 
 logger = logging.getLogger('osbs')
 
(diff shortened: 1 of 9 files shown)


osbs/cli/main.py
```diff
@@ -285,6 +289,38 @@ def cmd_resume_builds(args, osbs):
...
 
+def cmd_backup(args, osbs):
+    dirname = "osbs-backup-%s-%s-%s" % (args.instance, args.namespace, datetime.now().isoformat())
+    try:
+        mkdir(dirname, 0o700)
+    except OSError:
+        logger.error("Backup directory %s already exists, not overwriting", dirname)
+        raise
+
+    with paused_builds(osbs, args.namespace):
```
twaugh: Nice!

osbs/cli/main.py
```diff
@@ -416,6 +452,19 @@ def cli():
                                           description='allow builds to be scheduled again after pause-builds')
     resume_builds.set_defaults(func=cmd_resume_builds)
 
+    backup_builder = subparsers.add_parser(str_on_2_unicode_on_3('backup-builder'),
+                                           help='dump builder data (admin)',
+                                           description='create backup of all OSB data')
```
twaugh: Typo

TomasTomecek:
Oh mai, this is very impressive.

mmilata:
Fixed the typo, let me know when I should squash the commits.

twaugh:
Tested on a system with imagestreams, builds, and buildconfigs with triggers on the imagestreams.

Worked fine, everything restored with no superfluous builds.

twaugh:
Would it be nicer if it sent/received a tarfile over standard output? That way you could do things like:

```
osbs --instance=old backup-builder | osbs --instance=new restore-builder
```

It also avoids the cumbersome directory name (I tried using rsync on it and it got confused by the ':'s in the timestamp).

osbs/core.py
```diff
@@ -595,6 +595,26 @@ def import_image(self, name, namespace=DEFAULT_NAMESPACE):
...
+    def restore_resource(self, resource_type, resource, namespace=DEFAULT_NAMESPACE):
+        url = self._build_url("namespaces/%s/%s" % (namespace, resource_type))
+        response = self._post(url, data=json.dumps(resource),
+                              headers={"Content-Type": "application/json"})
+
+        ## replace instead of create - fails on already existing resources
+        #name = resource['metadata']['name']
+        #url = self._build_url("namespaces/%s/%s/%s" % (namespace, resource_type, name))
+        #response = self._put(url, data=json.dumps(resource),
+        #                     headers={"Content-Type": "application/json"})
```
twaugh: Can we remove the commented-out bits?

osbs/utils.py
```diff
@@ -64,6 +66,20 @@ def graceful_chain_get(d, *args):
     return t
 
 
+def graceful_chain_del(d, *args):
+    if not d:
+        return
+    for arg in args[:-1]:
+        try:
+            d = d[arg]
+        except (AttributeError, KeyError): #when do we get AttributeError? IndexError for lists instead?
```
twaugh: I can't see why we'd get anything but KeyError or IndexError there.

mmilata:
Thanks for testing. Addressed the inline comments, will look into the tar output later.

mmilata:
Rebased on top of master due to conflicts. PTAL

Changes: backup now creates a tarball so now you can do:
```
osbs -i old backup-builder -f - | osbs -i new restore-builder -
```

Default file name also no longer uses colons.

twaugh:
Works well!

Do we need to think about backing up resourcequota? I'm thinking not, because osbs doesn't create those itself except for the pause/resume one.

LGTM

osbs/cli/main.py
```diff
@@ -291,6 +295,50 @@ def cmd_resume_builds(args, osbs):
...
+    asciireader = codecs.getreader('ascii')
+
+    with paused_builds(osbs, args.namespace):
+        for f in TarReader(infile):
+            resource_type = os.path.basename(f.filename).split('.')[0]
+            if resource_type not in BACKUP_RESOURCES:
+                logger.warning("Unknown resource type for %s, skipping", f.filename)
+                continue
+
+            logger.info("restoring %s" % resource_type)
```
twaugh: Silly nitpick: this, and also line 311, could be written `logger.info("... %s", resource_type)` (no `%`).

mmilata:
Thanks! I wouldn't backup resourcequotas, I would only backup data created by OSBS (possibly indirectly). I'd consider the resourcequota created by ansible-osbs to be configuration and therefore out of scope.

Let me just at least some tests before merging.

twaugh:
Yes, agreed.

osbs/cli/main.py
mmilata: Good catch, will force-push fixed commit.

mmilata:
Added tests, fixed the string formatting.

twaugh:
LGTM

PR #321, opened 2015-12-04
Fix authentication for Python 3

mmilata:
Using auth on python 3 resulted in OsbsException: KeyError(b'access_token',)

Essentially reverts f7aafa6 since the headers are unicode/str after the http refactoring in 61f327b.


TomasTomecek:
Tested with `list-builds` command on 2 and 3. Works, merging.

PR #322, opened 2015-12-14
import_image: report whether new tags were imported

twaugh:


mmilata:
LGTM, merged (github doesn't think so, though?).

mmilata:
Looks like I forgot to do `git pull` ... the result looks the same so let me close it.

twaugh:
It's just that I rebased it for the Kerberos ccache fix. It's fine now. Thanks!

PR #323, opened 2015-12-15
Fix kerberos_ccache test cases

twaugh:
Some of the `pytest` expectations were failing because of extra keyword arguments.


twaugh:
@mmilata, what do you think?

twaugh:
Example failure from before this PR: https://travis-ci.org/projectatomic/osbs-client/builds/96783835

mmilata:
I vaguely recall having a hard time using pytest expectations - maybe not passing the implicit kwargs was the reason? Anyway the test pass both on Travis and my machine so LGTM.

PR #324, opened 2015-12-17
BuildResponse: only accept JSON in constructor

mmilata:
Fixes a bug where two builds were started the first time we created a BuildConfig.

Another thing we could do is to have single argument constructor where the code would detect `HttpResponse`s and automagically call `.json()` on them. I slightly prefer doing it explicitly.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -90,19 +90,19 @@ class OSBS(object):
         serialized_response = response.json()
         build_list = []
         for build in serialized_response["items"]:
-            build_list.append(BuildResponse(None, build))
+            build_list.append(BuildResponse(build))
         return build_list
 
     @osbsapi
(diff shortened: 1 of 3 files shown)


osbs/api.py
```diff
@@ -243,9 +243,10 @@ def _create_build_config_and_build(self, build_request, namespace):
                         raise OsbsException(
                             self._panic_msg_for_more_running_builds(build_config_name, builds))
                     else:
-                        build = builds[0].request
```
mmilata: When `BuildConfig` gets autoinstantiated, `.request` is `None` resulting in explicitly starting another build.

twaugh:
Nice catch! Reviewing...

twaugh:
LGTM!

PR #332, opened 2016-01-07
Use watch API to wait for autoinstantiated builds

### LINKED ISSUE #216: BuildConfig instantiation polling not quite right
For BuildConfigs we expect to instantiate builds automatically we're currently waiting up to 60 seconds for *running* builds, but we have no guarantee the builds will get scheduled in that time. All we need to wait for is *a* build to show up.

Additionally, we shouldn't really need to poll for this, just `watch` (like `OpenShift.wait()` does).

(Aside: I've been trying tests with a trigger listed in the JSON, and trying to work out how to automate tests for this.)

mmilata:
Closes #216.

There's no timeout, we don't have it for any other watch methods either - should we?

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -171,21 +171,6 @@ class OSBS(object):
                 running.append(br)
         return running
 
-    def _poll_for_builds_from_buildconfig(self, build_config_id, namespace=DEFAULT_NAMESPACE):
-        # try polling for 60 seconds and then fail if build doesn't appear
-        deadline = int(time.time()) + 60
-        while int(time.time()) < deadline:
-            logger.debug('polling for build from BuildConfig "%s"',
-                         build_config_id)
-            builds = self._get_running_builds_for_build_config(build_config_id, namespace)
-            if len(builds) > 0:
-                return builds
-            # wait for 5 seconds before trying again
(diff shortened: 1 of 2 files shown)


TomasTomecek:
Polling had 60 seconds, that sounds reasonable. Not sure about implementation, does pycurl support timeout? (or do we have to start new thread with timer?)

mmilata:
The last time I tested this the 60 seconds were sometimes not enough. PyCURL seems to [support timeouts](http://curl.haxx.se/libcurl/c/CURLOPT_TIMEOUT.html). Should I open a separate ticket for this?

TomasTomecek:
I would stuff it into this one. But to be honest, I'm not sure we need the timeout (the build should be scheduled no matter what, right?). @twaugh Tim, what do you think?

twaugh:
I don't think it requires a timeout really.

twaugh:
LGTM, and it works fine here.

twaugh:
Also FWIW it took 117s for a build to get scheduled here just now.

TomasTomecek:
> 117s for a build to get scheduled here just now.

oh my, `+1` for not doing timeout then

PR #334, opened 2016-01-11
Refactor: don't pass namespace as an argument

### LINKED ISSUE #333: refactoring: namespace as a property of OSBS instance
Currently a lot of methods of `class OSBS` and `class OpenShift` take optional parameter `namespace`. I propose to make it property of those classes and initialize it once in constructor so we don't have to pass it around everywhere.

It's part of our python API but koji-containerbuild doesn't use it. Is it worthwhile?

mmilata:
Passing it as a parameter everywhere clutters the code and serves no
purpose. It is initialized in OSBS constructor and lives as a property of
class Openshift now.

Breaks our Python API in case non-default namespace is used.

Closes #333.

---

* I took the liberty of removing support for listing builds in all namespaces. It wasn't working for some time now. I can make it work if anybody thinks it's useful.
* The `osbsapi` decorator checks if the methods are called with `namespace` kwarg and prints warning if it's the case. This means that we don't break compatibility with API users that use the default namespace. With little bit more work we can probably unbreak it for API users that use other namespace, though I don't think there are any.
* If this PR is merged I'll fix the osbs-client usage in atomic-reactor.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -12,13 +12,14 @@ import logging
 import os
 import sys
 import time
+import warnings
 from functools import wraps
 
 from .constants import SIMPLE_BUILD_TYPE, PROD_WITHOUT_KOJI_BUILD_TYPE, PROD_WITH_SECRET_BUILD_TYPE
 from osbs.build.build_request import BuildManager
 from osbs.build.build_response import BuildResponse
 from osbs.build.pod_response import PodResponse
-from osbs.constants import DEFAULT_NAMESPACE, PROD_BUILD_TYPE
+from osbs.constants import PROD_BUILD_TYPE
 from osbs.core import Openshift
 from osbs.exceptions import OsbsException, OsbsValidationException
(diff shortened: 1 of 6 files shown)


TomasTomecek:
Very nice! Not sure if we could use [`warnings`](https://docs.python.org/2/library/warnings.html) module to indicate that `namespace` argument is no longer supported.

mmilata:
Good point. According to https://docs.python.org/2/howto/logging.html#when-to-use-logging:

> warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning

> logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted

Will update to use warnings.

mmilata:
Updated.

TomasTomecek:
LGTM

mmilata:
Rebased.

twaugh:
LGTM

PR #336, opened 2016-01-15
OSBS.create_build(): increase API compatibility

twaugh:
Shouldn't need to provide git_branch=None or target=None.


mmilata:
LGTM

Issue #339, opened 2016-02-09
osbs-client times out (and gives traceback) when watching a paused build


twaugh:
I paused the builder and submitted a job, and waited for an hour.

```
[...]
2016-02-09 16:13:28,476 - osbs.core - INFO - object has changed: 'modified', status: 'Pending', name: 'docker-hello-world-master-21'
2016-02-09 16:13:28,476 - osbs.core - INFO - matching build found
2016-02-09 16:13:28,476 - osbs.core - DEBUG - is 'pending' in ['failed', 'complete', 'error', 'cancelled', 'running']?
2016-02-09 16:13:28,476 - osbs.core - DEBUG - No, build is not in the state I'm waiting for.
2016-02-09 17:13:28,459 - osbs.http.curl - DEBUG - Closing connection 0
2016-02-09 17:13:28,459 - osbs.http - DEBUG - end of the stream
2016-02-09 17:13:28,459 - osbs.http - DEBUG - cleaning up
2016-02-09 17:13:28,459 - osbs.core - ERROR - build 'docker-hello-world-master-21' was not found during wait
Traceback (most recent call last):
  File "/usr/lib/python3.4/site-packages/osbs_client-0.16-py3.4.egg/osbs/api.py", line 37, in catch_exceptions
  File "/usr/lib/python3.4/site-packages/osbs_client-0.16-py3.4.egg/osbs/api.py", line 399, in wait_for_build_to_get_scheduled
  File "/usr/lib/python3.4/site-packages/osbs_client-0.16-py3.4.egg/osbs/core.py", line 471, in wait_for_build_to_get_scheduled
  File "/usr/lib/python3.4/site-packages/osbs_client-0.16-py3.4.egg/osbs/core.py", line 456, in wait
NameError: name 'response' is not defined

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/osbs", line 9, in <module>
    load_entry_point('osbs-client==0.16', 'console_scripts', 'osbs')()
  File "/usr/lib/python3.4/site-packages/osbs_client-0.16-py3.4.egg/osbs/cli/main.py", line 553, in main
  File "/usr/lib/python3.4/site-packages/osbs_client-0.16-py3.4.egg/osbs/cli/main.py", line 194, in cmd_build
  File "/usr/lib/python3.4/site-packages/osbs_client-0.16-py3.4.egg/osbs/api.py", line 48, in catch_exceptions
[...]

PR #340, opened 2016-02-09
watch_resource: reconnect when connection closed (#339)

twaugh:
To test: pause a builder and submit a job, then wait for an hour.


twaugh:
Tested against Origin 1.0.8. Fixes #339.

mmilata:
LGTM, merged!

mmilata:
Thinking about it, shouldn't we have a short sleep between the retries, or perhaps a limited number of retries?

twaugh:
Hmm, yes.

Issue #342, opened 2016-02-11
pause-builds then backup-builder causes builds to be resumed


twaugh:
I think it would be good for **backup-builder** to use a different resourcequota name than pause-builds. That way this could work correctly:

```
osbs pause-builds
osbs backup-builder -f ...
osbs resume-builds
```

If the builder is paused, we don't we **backup-builder** to resume it.

PR #343, opened 2016-02-11
Different resource quota name for backup pause

mmilata:
This should solve #342. I'm wondering if the `resume-builds` subcommand should also take the resource quota as a parameter to facilitate resuming builds after interrupted backup command?

### DIFF
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -195,14 +195,14 @@ def checkout_git_repo(git_uri, git_ref, git_branch=None):
 
 
 @contextlib.contextmanager
-def paused_builds(osbs):
-    logger.info("pausing builds")
-    osbs.pause_builds()
+def paused_builds(osbs, quota_name=None):
     try:
+        logger.info("pausing builds")
+        osbs.pause_builds(quota_name=quota_name)
         yield osbs
     finally:
         logger.info("resuming builds")
-        osbs.resume_builds()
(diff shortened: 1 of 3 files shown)


osbs/utils.py
```diff
@@ -195,14 +195,14 @@ def checkout_git_repo(git_uri, git_ref, git_branch=None):
 
 
 @contextlib.contextmanager
-def paused_builds(osbs):
+def paused_builds(osbs, quota_name=None):
     logger.info("pausing builds")
-    osbs.pause_builds()
+    osbs.pause_builds(quota_name=quota_name)
```
twaugh: I'm wondering if we should move the `osbs.pause_builds()` call inside the `try` block. We can `resume_builds()` without having to have paused them in the first place without error, and the pause call will put the resource quota in place and then wait for builds to finish.

If we interrupt it during waiting, we'll leave the resource quota in place. I think that's what must have happened when we saw this acting up previously.

twaugh:
Yes, it wouldn't hurt to let it have an optional parameter for the resource quota name I suppose.

osbs/utils.py
mmilata: Oh, good point!

mmilata:
Updated with your suggestions. Hopefully I didn't overlook some case where `paused_builds` does not behave as expected when exception is raised somewhere.

twaugh:
LGTM, thanks!

PR #344, opened 2016-02-12
setup.py: package all jsons in inputs/

mmilata:


twaugh:
LGTM

twaugh:
When `setup.py install` is run with the `--root ...` option it installs the `data_files` into the data dir in that build root.

Note: when running it without `--root` it only installs the data files in the egg, not in the filesystem.

Issue #252, opened 2015-09-22
Field-based filtering for `list-builds`


twaugh:
It would be useful to be filter on particular values for particular fields.

`osbs list-builds image_id=...` would show the build that produced a given image.

`osbs list-builds user=alice user=bob` would list builds for both alice and bob (same key name so logical OR).

`osbs list-builds state=pending image=alice/image1 image=alice/image2` would list alice's pending builds for image1 and image2 (logical AND and ORs inferred like journalctl does).

`osbs list-builds substring` would list builds whose image names match `substring`, as currently.

TomasTomecek:
Excellent idea. As you said, this won't even break/change existing CLI interface.

PR #345, opened 2016-02-18
Improve performance for `list-builds --running`

twaugh:
Previously, all builds were fetched and then each build's status was examined.

Now, the fieldSelector is set to match status=running in the builds query when the --running option is passed to the CLI tool. This means only running builds will be returned.

API changes:
- OSBS.list_builds() now accepts a 'field_selector' keyword argument

This also brings us closer to an implementation for #252.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -90,8 +90,15 @@ class OSBS(object):
         return self._bm
 
     @osbsapi
-    def list_builds(self):
-        response = self.os.list_builds()
+    def list_builds(self, field_selector=None):
+        """
+        List builds with matching fields
+
+        :param field_selector: str, field selector for Builds
+        :return: BuildResponse list
+        """
+
+        response = self.os.list_builds(field_selector=field_selector)
         serialized_response = response.json()
         build_list = []
         for build in serialized_response["items"]:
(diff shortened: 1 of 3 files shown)


TomasTomecek:
they have finally made this in openshift? oh, the joy!

twaugh:
I think the fieldSelector part has been there for ages (ever?). This doesn't limit the *amount* of data reported for each build, only *which* builds are reported.

twaugh:
Oh, wait, it needs to find the pending jobs as well.

twaugh:
Hmm, this query seems to return all the builds:
```
GET /oapi/v1/namespaces/default/builds/?fieldSelector=status%21%3Dfailed%2Cstatus%21%3Dcomplete%2Cstatus%21%3Derror%2Cstatus%21%3Dcancelled HTTP/1.1
```

from some code I'm testing locally:
```
       field_selector = ",".join(["status!={status}".format(status=status)
                                  for status in BUILD_FINISHED_STATES])
       kwargs['field_selector'] = field_selector
```

twaugh:
The problem was that the status to match against is eg "Completed", not "completed".

TomasTomecek:
@_@

TomasTomecek:
this thing is mind-blowing - helped me a lot; merging

PR #350, opened 2016-02-22
New list-builds column 'duration'

twaugh:


TomasTomecek:
very nice! (some unit test would be very neat)

twaugh:
I'm going to close this because more comprehensive performance metrics reporting is being worked on over in https://github.com/twaugh/osbs-metrics.

PR #353, opened 2016-03-03
README: osbs build requires -b param

vrutkovs:


TomasTomecek:
+1

I've also experienced that `$ osbs build` isn't very user friendly.

PR #357, opened 2016-03-07
Make it possible to specify buildroot to be used

### LINKED ISSUE #356: Should be able to specify which buildroot to use
Currently the buildroot container to use is set to `buildroot`. There should be a command line option to set it to something else.

mmilata:
Closes #356.
I'm not quite sure where to put the default value, whether into *spec.py*, *conf.py*, or both.

### DIFF
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -14,7 +14,7 @@ import logging
 import datetime
 import os
 import re
-from osbs.constants import DEFAULT_GIT_REF
+from osbs.constants import DEFAULT_GIT_REF, DEFAULT_BUILD_IMAGE
 from osbs.exceptions import OsbsValidationException
 from osbs.utils import (get_imagestreamtag_from_image,
                         git_repo_humanish_part_from_uri,
@@ -147,6 +147,7 @@ class CommonSpec(BuildTypeSpec):
     name = BuildIDParam()
     yum_repourls = BuildParam("yum_repourls")
     use_auth = BuildParam("use_auth", allow_none=True)
+    build_image = BuildParam('build_image')
 
     def __init__(self):
         self.required_params = [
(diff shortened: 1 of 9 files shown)


twaugh:
Nice!

I think setting the default value in spec.py should be sufficient. There is already precedent for this with `git_ref`, and `BuildParam`'s constructor takes a `default` keyword argument to allow it. Also, this parameter is specifically about a value in the build JSON, rather than something more general.

osbs/build/spec.py
```diff
@@ -147,6 +147,7 @@ class CommonSpec(BuildTypeSpec):
     name = BuildIDParam()
     yum_repourls = BuildParam("yum_repourls")
     use_auth = BuildParam("use_auth", allow_none=True)
+    build_image = BuildParam('build_image')
```
twaugh: Could use 'default=' here, for consistency with how `git_ref` is handled. (Or is `git_ref` the one that needs bringing into line?)

twaugh:
Thinking more about how koji-containerbuild plugin will use this, it might be quite useful to have an optional keyword argument `build_image` in `create_prod_build()` etc. That way it could direct builds to diferent buildroots without needing to re-init OSBS. What do you think?

TomasTomecek:
> That way it could direct builds to diferent buildroots without needing to re-init OSBS. What do you think?

I agree.

Not sure if at some point we'll need multiple build images for multiple streams (koji targets) to handle 1.10 etc.

twaugh:
It's possible, but the main reason for doing this is the ability to reproduce the build environment after the fact IMHO.

osbs/build/spec.py
mmilata: It appears to me that the defafault= value is overwritten by the first call to `set_params`, namely to `None` if no value is passed.

mmilata:
Yeah, it makes sense from koji-containerbuild perspective. Added commit does just that. But it can still be set in the configuration (like most everything else?).

twaugh:
LGTM!

PR #358, opened 2016-03-08
new call: "get-quota": get or list resource quota

TomasTomecek:
* api
* cli

should I also add tests?


mmilata:
LGTM, merged, thanks!

PR #360, opened 2016-03-09
New OSBS.get_compression_extension() method

twaugh:
This allows clients to work out what filename extension to expect for the 'docker save' output file.


twaugh:
coveralls is in read-only mode at the moment, but running `py.test-3 --cov=osbs.api --cov-report=annotate` shows only one case not handled by testing, which is when the compress plugin's 'method' argument is set but is neither 'gzip' nor 'lzma'.

twaugh:
I've added more tests now, covering the case above.

mmilata:
Nice, merged!

Issue #187, opened 2015-07-31
Wrong error message if connection is closed in lower layer than application one


pbabinca:
When I receive streamed logs from OpenShift with `get_build_logs(follow=True)` the connection to OpenShift is sometimes closed due https://github.com/openshift/origin/issues/2348. This is done on lower levels of networking stack than application one.

osbs-client doesn't check this and if `pycurl.error` exception occur it just prints last message `(18, 'NSS: client certificate not found (nickname not specified)')` which isn't error just a warning.

TomasTomecek:
So, what do you propose then? Am not sure I follow precisely.

pbabinca:
Try to query pycurl what was the cause of the lost connection and if it isn't on application level don't print that message.

mmilata:
The error code `18` is `CURLE_PARTIAL_FILE` which corresponds to the message *Transferred a partial file*, which is what happened here. The error message displayed doesn't correspond to the code, which suggests this might be pycurl bug.

Unfortunately pycurl doesn't provide means of translating the code to the human readable string. Looking further into this ...

PR #361, opened 2016-03-15
HttpStream: don't raise for E_PARTIAL_FILE

twaugh:
This allows the connection to end normally when closed by the server.

### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -258,7 +258,10 @@ class HttpStream(object):
 
         if err_list:
             err_obj = err_list[0]
-            raise OsbsNetworkException(self.url, err_obj[2], err_obj[1])
+
+            # For anything but the connection being closed, raise
+            if err_obj[1] != pycurl.E_PARTIAL_FILE:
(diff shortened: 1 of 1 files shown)


twaugh:
Fixes #187.

twaugh:
From libcurl-errors(3):

```
       CURLE_PARTIAL_FILE (18)
              A file transfer was shorter or larger than expected.  This  hap‐
              pens  when  the  server first reports an expected transfer size,
              and then delivers data that doesn't match the  previously  given
              size.
```

twaugh:
How to reproduce:

1. Edit /etc/origin/master/master-config.yaml and change the servingInfo.requestTimeoutSeconds parameter (last line) to 60
2. Restart origin-master
3. pause builds with `osbs pause-builds` or just set a resource quota by hand:
```
echo '{"kind":"ResourceQuota","apiVersion":"v1","metadata":{"name":"pause"},"spec":{"hard":{"pods":"0"}}}' | oc create -f -
```
4. submit a build

TomasTomecek:
I can confirm this patch closes the connection in a sane way; I'm still trying without the patch.

PR #363, opened 2016-03-15
(do not merge) cp_built_image_to_nfs: changes to allow for separate nodes

twaugh:


twaugh:
Is the pod's `status.hostIP` sufficient, or should we maybe use the pod's `spec.nodeName` and look up the node's `status.addresses` of type Hostname or ExternalIP.

twaugh:
Dropping.

PR #373, opened 2016-03-31
Add test harness for measuring build throughput

mmilata:


tests/throughput-test-harness.py
```diff
@@ -0,0 +1,172 @@
...
+    logging.info("running: %s", " ".join(args))
+
+    p = subprocess.Popen(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    (out, err) = p.communicate()
+    if out:
+        logging.info("stdout:\n%s", out.rstrip())
+    if err:
+        logging.info("stderr:\n%s", err.rstrip())
+    if p.returncode != 0:
+        raise RuntimeError("Subprocess failed w/ return code {0}".format(p.returncode))
```
vrutkovs: Not sure if we should throw exception here - a single failure will require a test restart.

What if we log the exception and show a list of failed builds after the full cycle is done?

mmilata: I'd rather fail by default, especially for the git commands, to avoid working on a wrong branch if a checkout fails or pushing something that is not in some expected state. It probably makes sense for starting builds, though - I'll update it.

vrutkovs: Okay, makes sense

mmilata:
@vrutkovs, added new commit, what do you think?

vrutkovs:
@mmilata yeah, that should address that concern I had in mind

twaugh:
I re-based against master and added a --repo-url parameter for start-builds.

vrutkovs:
LGTM

tests/throughput-test-harness.py
```diff
@@ -0,0 +1,214 @@
...
+import argparse
+import subprocess
+
+from dockerfile_parse import DockerfileParser
+
+DEFAULT_BRANCH_PREFIX = "branch"
+DEFAULT_BRANCH_COUNT = 100
+DEFAULT_STAGGER_NUMBER = 10
+DEFAULT_STAGGER_WAIT = 60
+DEFAULT_KOJI_BIN = "brew-test"
```
twaugh: I guess this should be 'koji'.

mmilata: Fixed.

PR #375, opened 2016-04-01
Allow user to specify OAuth 2.0 token to use

twaugh:


vrutkovs:
LGTM, works on OpenShift with googleauth enabled

PR #376, opened 2016-04-01
Don't set arg defaults for config items

twaugh:
Since args override configuration file values, setting a default value for an argument will always mask the configuration file value.


mmilata:
D'oh ... LGTM

PR #377, opened 2016-04-01
Configurable secret for koji ssl certs

mmilata:
@maxamillion tested this and it seems to work: http://koji.stg.fedoraproject.org/koji/getfile?taskID=90071405&name=build.log (the auth, at least)

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -766,7 +766,7 @@ class ProductionBuild(CommonBuild):
                           self.spec.pdc_secret.value,
 
                           ('exit_plugins', 'koji_promote', 'koji_ssl_certs'):
-                          self.spec.koji_certs_secret})
+                          self.spec.koji_certs_secret.value})
 
         if self.spec.pulp_secret.value:
             # Don't push to docker registry, we're using pulp here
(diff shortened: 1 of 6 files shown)


twaugh:
How about a mention in the docs?

osbs/build/spec.py
```diff
@@ -258,6 +260,7 @@ def set_params(self, sources_command=None, architecture=None, vendor=None,
         self.koji_target.value = koji_target
         self.kojiroot.value = kojiroot
         self.kojihub.value = kojihub
+        self.koji_certs_secret = koji_certs_secret
```
twaugh: This should be `self.koji_certs_secret.value =`.

osbs/build/build_request.py
```diff
@@ -762,7 +763,10 @@ def render(self, validate=True):
                           self.spec.pulp_secret.value,
 
                           ('exit_plugins', 'sendmail', 'pdc_secret_path'):
-                          self.spec.pdc_secret.value})
+                          self.spec.pdc_secret.value,
+
+                          ('exit_plugins', 'koji_promote', 'koji_ssl_certs'):
+                          self.spec.koji_certs_secret})
```
twaugh: This should be `self.koji_certs_secret.value`.

twaugh:
LGTM, merging.

Issue #370, opened 2016-03-30
Don't update BuildConfig for scratch builds


twaugh:
When submitting a scratch build, ideally we would create a stand-alone 'Build' object instead of updating any existing BuildConfig and instantiating it.

By doing this, we can avoid scratch builds from interfering with chain rebuilds.

vrutkovs:
Fixed via #452

PR #378, opened 2016-04-01
Enable Content Generators

maxamillion:


osbs/build/spec.py
```diff
@@ -258,6 +260,7 @@ def set_params(self, sources_command=None, architecture=None, vendor=None,
         self.koji_target.value = koji_target
         self.kojiroot.value = kojiroot
         self.kojihub.value = kojihub
+        self.koji_certs_secret = koji_certs_secret
```
twaugh: Here's the corresponding bug we expected earlier. This needs to be

```
self.koji_certs_secret.value = koji_certs_secret
```

twaugh: I just understood what you told me earlier - this comes from another PR. I've added comments there about the two instancesides of missing `.value`, and you'll be able to re-base once they're fixed.

In the meantime you could add the fixes to this pull request as a separate commit that will be easy to skip on re-base, just for the purpose of getting the Travis testing to pass.

twaugh:
Now that https://github.com/projectatomic/osbs-client/pull/377 is merged, please re-base against master.

twaugh:
Also:
* please squash together related commits
* the `or None` change (c038b02) is no longer needed.

maxamillion:
Rebased and removed change from https://github.com/projectatomic/osbs-client/commit/c038b025d92e2a83a8ed1e8b1ca758cadf72ec09

osbs/build/build_request.py
```diff
@@ -473,6 +484,21 @@ def adjust_for_triggers(self):
...
+        Remove koji Content Generator related plugins if no triggers set in
+        order to hadle the "scratch build" scenario
+        """
+        if self.scratch:
+            # Note: only one for now, but left in a list like other adjust_for_
+            # functions in the event that this needs to be expanded
+            for when, which in [
+                ("exit_plugins", "koji_promote"),
+            ]:
+                logger.info("removing %s from request because there are no triggers",
```
twaugh: This should be a message about scratch builds, not triggers, I think?

osbs/build/build_request.py
```diff
@@ -748,6 +774,7 @@ def render(self, validate=True):
             del self.template['spec']['triggers']
 
         self.adjust_for_triggers()
+        self.adjust_for_scratch()
```
twaugh: If we ran `adjust_for_scratch()` before `adjust_for_triggers()`, we could actually remove any triggers for this build. We never want to have automated rebuilds for scratch builds.

(Note: this adjusts the BuildConfig, so kills any existing triggers... fixing that is separate issue, https://github.com/projectatomic/osbs-client/issues/370, which we won't worry about here)

twaugh:
Did some changes get lost? This looks like all it does is deal with scratch builds.

twaugh:
I think the missing parts are:
* don't remove koji_promote when there are no triggers
* don't run the cp_built_image_to_nfs plugin at all (`render_cp_built_image_to_nfs` can go, and the references to it in `inputs/*.json` can go as well)

maxamillion:
I don't know, I rebased as requested and this is what I got ... let me inspect git history and figure out what happened.

twaugh:
I added another commit on https://github.com/twaugh/osbs-client/tree/content-generator to fix up the tests, based on what you had before.

So what's left is: run adjust_for_scratch() before adjust_for_triggers(); fix adjust_for_scratch log message.

twaugh:
Oh, and you'll need to cherry-pick that change I made (it's 7106f3a).

maxamillion:
@twaugh - Thank you, I missed that ... been badly multi-tasking all morning which is effectively done nothing good for anyone. Apologies.

twaugh:
The scratch/triggers message, and removing triggers for scratch builds, can be dealt with later.

PR #380, opened 2016-04-05
Two CLI options for oauth-enabled installations

vrutkovs:
Includes #375

 * print-token-url: print an auth page url to get token
 * get-serviceaccount-token: prints tokens set for serviceaccount

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -23,11 +23,18 @@ from osbs.build.build_response import BuildResponse
 from osbs.cli.render import TablePrinter
 from osbs.conf import Configuration
 from osbs.constants import (DEFAULT_CONFIGURATION_FILE, DEFAULT_CONFIGURATION_SECTION,
-                            CLI_LIST_BUILDS_DEFAULT_COLS, PY3, BACKUP_RESOURCES, DEFAULT_NAMESPACE,
+                            CLI_LIST_BUILDS_DEFAULT_COLS, PY3, BACKUP_RESOURCES,
                             BUILD_FINISHED_STATES)
 from osbs.exceptions import OsbsNetworkException, OsbsException, OsbsAuthException, OsbsResponseException
 from osbs.cli.capture import setup_json_capture
 from osbs.utils import strip_registry_from_image, paused_builds, TarReader, TarWriter
+
+try:
+    # py2
+    from urlparse import urljoin
+except ImportError:
+    # py3
+    from urllib.parse import urljoin
 
 logger = logging.getLogger('osbs')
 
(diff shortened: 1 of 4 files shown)


osbs/cli/main.py
```diff
@@ -349,6 +356,17 @@ def cmd_restore(args, osbs):
     logger.info("backup recovery complete!")
 
 
+def cmd_oauth_token(args, osbs):
+    uri = urljoin(osbs.os_conf.get_openshift_base_uri(), "/token/request")
```
twaugh: Should this be `/oauth/token/request`?

vrutkovs: Indeed, gonna fix that, although my test instance somehow managed to redirect correctly

osbs/cli/main.py
```diff
@@ -499,6 +517,17 @@ def cli():
                                  help="don't stop when restoring a resource fails")
     restore_builder.set_defaults(func=cmd_restore)
 
+    oauth2_builder = subparsers.add_parser(str_on_2_unicode_on_3('print-oauth-token-url'),
```
twaugh: We ought to have some consistency about how we refer to tokens. Here it's 'oauth-token', elsewhere it's 'oauth2-token' and 'token'.

Should we just use 'token' throughout?  e.g. osbs --token ..., "token = " in osbs.conf, --print-token-url, get-serviceaccount-token (as now).

vrutkovs: Okay, lets use 'token' (the auth page also uses that)

twaugh:
Looks really nice.

I'd like to use this PR to also figure out a way of storing the service account token separately from the config file. Maybe a 'token_file' configuration parameter?

Then I think 'get-serviceaccount-token' could output just the token on its own without the secret's name, so we could do this:

```
osbs --token $ADMIN_USER_TOKEN get-serviceaccount-token $SERVICEACCOUNT > sa-token
osbs --token-file sa-token ...
```

vrutkovs:
get-serviceaccount-token may return several tokens, a simple
redirection will not work in some cases.

What if we add '--save-in-config' to overwrite osbs.conf param with
first found token?

twaugh:
OK, so perhaps let's just implement a `--token-file` argument/config param, so the service account token can be stored outside of osbs.conf.

vrutkovs:
Done, now we can read token from file too.

Reverted 'deprecate ```use_auth``` and look for ```use_httpd_auth```' change - its still required for oauth to work

twaugh:
OK, figured out what wasn't working for me: we need to base64-decode the data in the token secret before displaying it.

twaugh:
LGTM

twaugh:
It needs docs, but I'll file a separate issue for that.

PR #382, opened 2016-04-06
Run compress plugin even for v2-only builds

twaugh:
We will always use the'save' tarball, even for v2-only builds.


mmilata:
LGTM

PR #384, opened 2016-04-07
Tokens fixes

twaugh:
The only functional change is to make sure osbs.conf-specific token (whether 'token' or 'token_file') can be overridden.


vrutkovs:
LGTM

PR #385, opened 2016-04-08
only set Build_Host label if specified in config

twaugh:
See also https://github.com/projectatomic/atomic-reactor/pull/447


mmilata:
LGTM, wanna squash the commits?

twaugh:
Sure.

PR #386, opened 2016-04-08
We should use build_image from config rather than create_build kwarg

twaugh:
This removes the build_image keyword argument from the create_build API, and instead gets it from configuration/CLI.


mmilata:
Oops, stupid mistake ... merged, thank you!

PR #389, opened 2016-04-12
Use git reset --hard && git checkout

### LINKED ISSUE #388: Git reset --hard 'commit' is wrong
We should use 'git reset --hard && git checkout <commit>' in ```utils.py``` as I see a lot of 
```
OsbsException(u"Command [u'git', u'reset', u'--hard', u'master'] returned 128\n\nfatal: ambiguous argument 'master': unknown revision or path not in the working tree.\nUse '--' to separate paths from revisions, like this:\n'git <command> [<revision>...] -- [<file>...]'\n",)
```

vrutkovs:
Fixed #388

### DIFF
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -186,7 +186,8 @@ def checkout_git_repo(git_uri, git_ref, git_branch=None):
 
         # Find the specific ref we want
         try:
-            run_command(['git', 'reset', '--hard', git_ref], cwd=repo_path)
+            run_command(['git', 'reset', '--hard'], cwd=repo_path)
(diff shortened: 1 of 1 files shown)


TomasTomecek:
wild idea: we could also clean everything by `git clean -df`, not sure if that's needed

PR: lgtm, tested locally similar scenario

vrutkovs:
```git clean``` might dangerous, although we can fall back to it if ```checkout``` has failed

mmilata:
Merged, thank you! If you can reliably reproduce this it might be worth investigating whether it's a git bug, because this error is _really_ weird.

PR #391, opened 2016-04-14
Add back build_image keyword arg for create_build()

twaugh:
This is needed for allowing e.g. the koji-containerbuild plugin to set the build_image.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -247,6 +247,7 @@ class OSBS(object):
                           user, component,
                           target,      # may be None
                           architecture=None, yum_repourls=None,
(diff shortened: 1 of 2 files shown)


mmilata:
Can't koji-containerbuild pass it to Configuration object like this? https://github.com/release-engineering/koji-containerbuild/blob/master/koji_containerbuild/plugins/builder_containerbuild.py#L336

The configuration is getting a bit messy ...

twaugh:
Oh... right. I got lost in the code. Thanks for pointing out that way of doing it.

PR #392, opened 2016-04-14
Inject proxy in repos using 'proxy' config setting

vrutkovs:
See https://github.com/projectatomic/atomic-reactor/pull/461

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -212,6 +212,10 @@ class CommonBuild(BuildRequest):
             self.dj.dock_json_set_arg('prebuild_plugins',
                                       "add_yum_repo_by_url", "repourls",
                                       self.spec.yum_repourls.value)
+            if self.spec.proxy.value:
+                self.dj.dock_json_set_arg('prebuild_plugins',
+                                          "add_yum_repo_by_url", "inject_proxy",
+                                          self.spec.proxy.value)
 
     def render_check_and_set_rebuild(self, use_auth=None):
         if self.dj.dock_json_has_plugin_conf('prebuild_plugins',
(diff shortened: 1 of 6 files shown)


osbs/build/build_request.py
```diff
@@ -505,6 +505,13 @@ def render_add_labels_in_dockerfile(self):
                                     "add_labels_in_dockerfile",
                                     "labels", implicit_labels)
 
+    def render_add_yum_repo_by_url(self):
+        super(ProductionBuild, self).render_add_yum_repo_by_url()
+        if self.spec.proxy.value:
+            self.dj.dock_json_set_arg('prebuild_plugins',
+                                      "add_yum_repo_by_url", "inject_proxy",
+                                      self.spec.proxy.value)
```
mmilata: Please move this under `CommonBuild` (in spec.py too).

osbs/conf.py
```diff
@@ -356,6 +356,9 @@ def get_git_push_username(self):
     def get_build_image(self):
         return self._get_value("build_image", self.conf_section, "build_image")
 
+    def get_proxy(self):
+        return self._get_value("proxy", self.conf_section, "proxy")
+
```
mmilata: Please add to command-line options.

tests/build/test_build_request.py
```diff
@@ -265,6 +270,9 @@ def test_render_prod_request_with_repo(self, architecture, build_image):
         assert 'sourceSecret' not in build_json["spec"]["source"]
         assert plugin_value_get(plugins, "prebuild_plugins", "add_yum_repo_by_url",
                                 "args", "repourls") == ["http://example.com/my.repo"]
+        if proxy:
+            assert plugin_value_get(plugins, "prebuild_plugins", "add_yum_repo_by_url",
+                                    "args", "inject_proxy") == proxy
```
mmilata: Maybe also test here that the value doesn't get set when not using proxy?

mmilata:
Some remarks inline, otherwise looks good. I'm wondering whether we should call the parameter something else than "proxy" to avoid confusion in case we use some other kind of proxy in the future? Something like "yum_proxy"?

Github says there are conflicts, can you please rebase?

tests/build/test_build_request.py
vrutkovs: Agreed

vrutkovs:
>Something like "yum_proxy"?

Yeah, sound a bit more accurate

mmilata:
LGTM, merged.

PR #395, opened 2016-04-20
Allow specifying imagestream instead of buildimage docker reference in prod builds

### LINKED ISSUE #394: Support using ImageStreamTag as build image
This would require another config setting and a different customStrategy.kind if its set

vrutkovs:
Fixes #394

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -770,6 +770,11 @@ class ProductionBuild(CommonBuild):
             self.template['spec']['output']['to']['name'] = \
                 self.spec.image_tag.value
 
+        if self.spec.build_imagestream.value:
+            self.template['spec']['strategy']['customStrategy']['from']['kind'] = 'ImageStreamTag'
+            self.template['spec']['strategy']['customStrategy']['from']['name'] = \
+                self.spec.build_imagestream.value
+
         use_auth = self.spec.use_auth.value
         self.render_add_labels_in_dockerfile()
         self.render_koji()
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -283,8 +283,13 @@ class CommonBuild(BuildRequest):
             if 'sourceSecret' in self.template['spec']['source']:
                 del self.template['spec']['source']['sourceSecret']
 
-        self.template['spec']['strategy']['customStrategy']['from']['name'] = \
-            self.spec.build_image.value
+        if self.spec.build_imagestream.value:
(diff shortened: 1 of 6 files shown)


osbs/build/spec.py
```diff
@@ -190,6 +190,7 @@ def set_params(self, git_uri=None, git_ref=None,
 class ProdSpec(CommonSpec):
     git_branch = BuildParam('git_branch', allow_none=True)
     trigger_imagestreamtag = BuildParam('trigger_imagestreamtag')
+    build_imagestream = BuildParam('build_imagestream')
```
mmilata: Please move the imagestream code under `CommonSpec` so that simple builds can use it too (we should probably remove them in the future but that's another story ...).

mmilata: Also, `build_image` needs to be removed from `CommonSpec.required_params`.

osbs/conf.py
```diff
@@ -356,6 +356,9 @@ def get_git_push_username(self):
     def get_build_image(self):
         return self._get_value("build_image", self.conf_section, "build_image")
 
+    def get_build_imagestream(self):
+        return self._get_value("build_imagestream", self.conf_section, "build_imagestream")
```
mmilata: We also need `--build-imagestream` cmdline argument in `cli/main.py`. Perhaps make it [mutually exclusive](https://docs.python.org/2/library/argparse.html?highlight=argparse#mutual-exclusion) with `--build-image`?

osbs/build/build_request.py
```diff
@@ -770,6 +770,11 @@ def render(self, validate=True):
             self.template['spec']['output']['to']['name'] = \
                 self.spec.image_tag.value
 
+        if self.spec.build_imagestream.value:
+            self.template['spec']['strategy']['customStrategy']['from']['kind'] = 'ImageStreamTag'
+            self.template['spec']['strategy']['customStrategy']['from']['name'] = \
+                self.spec.build_imagestream.value
```
mmilata: I'd move this near the code that sets `build_image` so that it's clear in the code that the options are mutually exclusive.

mmilata:
I like it but see couple of issues, commented inline.

osbs/build/spec.py
vrutkovs: Okay, agreed, was not sure if we want this for simple build also

osbs/conf.py
vrutkovs: Sure, will do

mmilata:
Merged (after squashing), thanks!

PR #403, opened 2016-04-29
Set koji_promote.target so the build gets tagged

twaugh:
Requires https://github.com/projectatomic/atomic-reactor/pull/471; please do not merge before that PR.


twaugh:
This is still undergoing testing.

twaugh:
Tested and ready for review.

twaugh:
Caution: do not merge until atomic-reactor has been released with a koji_promote plugin that can accept `target`. https://github.com/projectatomic/atomic-reactor/pull/471

twaugh:
The actual condition we want to make sure of is that osbs-client isn't *released* with this change before atomic-reactor is. Seeing as we're going to release both within the next week or so, I think it is fine to merge this now if everyone is happy.

PR #404, opened 2016-04-29
New koji_task_id parameter for create_prod_build()

twaugh:
This is stored in the kojitask label of the BuildConfig.


mmilata:
LGTM, merged.

PR #405, opened 2016-05-02
Limit length of build names

lcarva:


osbs/api.py
```diff
@@ -189,19 +188,42 @@ def _panic_msg_for_more_running_builds(self, build_config_name, builds):
...
+        queries = (
+            (self.os.get_build_config_by_labels, git_labels),
+            (self.os.get_build_config, name),
+        )
+
+        existing_bc = None
+        for func, arg in queries:
+            try:
+                existing_bc = func(arg)
+                # build config found
```
twaugh: What happens if there is a build config matching the name we created, but it has conflicting git-repo-name and/or git-branch labels? i.e. if we created a name for a build config that already exists

osbs/build/build_request.py
```diff
@@ -732,6 +737,11 @@ def render(self, validate=True):
             self.spec.validate()
         super(ProductionBuild, self).render()
 
+        repo_name = git_repo_humanish_part_from_uri(self.spec.git_uri.value)
+        # What about forked repos?
```
twaugh: This relates to the build config name collision issue I mentioned above. Options are: pick a new name; error out.

tests/build/test_build_request.py
```diff
@@ -57,6 +58,21 @@ def test_build_request_isnt_auto_instantiated(self):
...
+        build_json = copy.deepcopy(TEST_BUILD_JSON)
+        br = BuildRequest('something')
+        flexmock(br).should_receive('template').and_return(build_json)
+        assert br.template['metadata'].get('labels') is None
+
+        br.set_label('label-1', 'value-1')
+        br.set_label('label-2', 'value-2')
+        br.set_label('label-3', 'value-3')
+        assert br.template['metadata']['labels'] == {
+                'label-1': 'value-1',
```
twaugh: 'pep8' doesn't like this indent:
`E126 continuation line over-indented for hanging indent`

osbs/build/build_request.py
lcarva: In the case of a forked repository, the repo_name will be the same. The only way to distinguish them would be to store the whole git uri as a label as opposed to just the repository name.

tests/build/test_spec.py
```diff
@@ -55,3 +55,23 @@ def test_registry_uris_param_v2(self):
         assert registry.uri == 'http://registry.example.com:5000'
         assert registry.docker_uri == 'registry.example.com:5000'
         assert registry.version == 'v2'
+
```
twaugh: pep8 wants 2 blank lines here.

osbs/api.py
lcarva: The build config name is always preserved from the existing build config (https://github.com/projectatomic/osbs-client/pull/405/files#diff-034b3f8b7a38c1ab83d3dc6cce5f2683R252).

osbs/build/build_request.py
twaugh: Oh, I see.

We don't really want to store the whole URI in the label, as labels also have a length limit we might hit.

Instead, we could say that different buildconfigs can share the same labels but will be distinguished by their spec.source.git.uri attributes.

I don't think this is a use-case we'll actually need to implement right now, so I'm fine with erroring out if there is already a build config with the same labels.

Later on if we need to we could handle the case that the labels query returns several build configs that match, and we have to further match on the spec.source.git.uri attribute.

What do you think?

lcarva: I think deferring this use case is valid at this point, especially because the branch name is taken into account. As a side note, we could always hash the URI and branch together and use that as a label, or maybe even as the build config name.

twaugh: That's true. I was hoping to keep things human readable if possible, even labels.

tests/build/test_spec.py
lcarva: Done.

tests/build/test_build_request.py
lcarva: done.

osbs/api.py
lcarva: Added `_verify_labels_match` to handle the case where a build config may have been matched by name but the git labels do not match.

lcarva:
I messed up the rebase command and caused unrelated commits to be re-done. Closing this PR, please follow up with #408

twaugh:
FWIW pull requests always follow the symbolic branch name they were opened with, so you can fix up your local tree and 'git push -f' that branch, and the PR will be fixed up too.

PR #406, opened 2016-05-03
change order of exit plugins

twaugh:
The store_metadata plugin now uses the result from the koji_promote plugin, so koji_promote needs to run before store_metadata.


mmilata:
LGTM!

PR #408, opened 2016-05-04
Limit length of build names

lcarva:


osbs/build/build_request.py
```diff
@@ -732,6 +737,11 @@ def render(self, validate=True):
             self.spec.validate()
         super(ProductionBuild, self).render()
 
+        repo_name = git_repo_humanish_part_from_uri(self.spec.git_uri.value)
+        # What about forked repos?
```
twaugh: Want to update this comment to describe the scheme we talked about in the original PR? i.e. in the future we could take spec.source.git.uri into account as well as labels, and so allow more than one build config with the same set of labels.

twaugh:
LGTM! Nice tests.

osbs/api.py
```diff
@@ -189,19 +191,57 @@ def _panic_msg_for_more_running_builds(self, build_config_name, builds):
...
+            (self.os.get_build_config, name),
+        )
+
+        existing_bc = None
+        for func, arg in queries:
+            try:
+                existing_bc = func(arg)
+                # build config found
+                break
+            except OsbsException:
```
mmilata: Isn't this exception too general? I think it might mask things like server 500s or network errors which we probably want to know about.

mmilata:
Commented inline, otherwise LGTM, impressive testing:)

osbs/api.py
lcarva: That's true. I could instead use OsbsResponseException, but that would also mask 500s.

How about in addition to only catching OsbsResponseException, also log the exception value?

mmilata: I was thinking of using new exception, something like `OsbsBuildConfigNotFound`, but now I see we must also handle 404 (or something) from `get_build_config`. At least logging the exception sounds reasonable.

lcarva: Added logging for exception.

mmilata:
Merged, thanks!

PR #410, opened 2016-05-05
Fix backup processing

### LINKED ISSUE #407: restore-builder: fix logic that prepares resources before import
Don't strip status.lastVersion from buildconfigs: causes build numbers for each restored buildconfig to start from 1, which then clashes with name of a restored build.

Typo: lastTrigerredImageID -> lastTriggeredImageID.

Review why are we stripping spec.triggers.*.imageChange.lastTriggeredImageID.

mmilata:
* Don't drop `status.lastVersion` from buildconfigs as this causes name collisions of newly created builds with restored builds.
* Don't drop `lastTriggeredImageID` because empty value may cause a rebuild from the same base image.

Closes #407.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -542,13 +542,6 @@ class OSBS(object):
     def _prepare_resource(resource_type, resource):
         utils.graceful_chain_del(resource, 'metadata', 'resourceVersion')
 
-        if resource_type == 'buildconfigs':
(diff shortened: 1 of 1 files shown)


vrutkovs:
LGTM

lcarva:
LGTM

PR #414, opened 2016-05-11
create_prod_build: raise exception for missing Name label (fixes #413)

### LINKED ISSUE #413: Show a better error message for a missing Name label
```
  File "/usr/lib/python2.6/site-packages/koji/daemon.py", line 1161, in runTask
      response = (handler.run(),)\n  File "/usr/lib/python2.6/site-packages/koji/tasks.py", line 158, in run
          return koji.util.call_with_argcheck(self.handler, self.params, self.opts)
            File "/usr/lib/python2.6/site-packages/koji/util.py", line 154, in call_with_argcheck
                return func(*args, **kwargs)\n  File "/usr/lib/koji-builder-plugins/builder_containerbuild.py", line 373, in handler
                    **create_build_args\n  File "/usr/lib/python2.6/site-packages/osbs/api.py", line 37, in catch_exceptions
                        return func(*args, **kwargs)\n  File "/usr/lib/python2.6/site-packages/osbs/api.py", line 358, in create_build
                            return self.create_prod_build(**kwargs)\n  File "/usr/lib/python2.6/site-packages/osbs/api.py", line 48, in catch_exceptions
                                raise OsbsException(cause=ex, traceback=sys.exc_info()[2])
                                OsbsException: KeyError(u\'Name\',)
```

twaugh:


mmilata:
I'm wondering if we can make switching to https://github.com/projectatomic/ContainerApplicationGenericLabels possible without needing to update osbs-client. Should we try looking at `Name` and if it's missing then 'name'? Or make the label name configurable?

twaugh:
Good point. At the moment, though, atomic-reactor requires 'Name' (e.g. post_tag_by_labels), so if osbs-client lets 'name' through when there is no 'Name', the build will ultimately fail.

mmilata:
Fair enough, let's merge it as it is and I'll have to look at the labels in a-r and o-c later.

PR #415, opened 2016-05-11
watch-builds: support json output

vrutkovs:
This output would be used by zabbix-metrics script as its simplier and 
the output doesn't get cached


mmilata:
LGTM, merged!

PR #417, opened 2016-05-13
Require OpenShift Origin >= 1.0.6 by default

twaugh:
Change default for openshift_required_version.

This causes BuildConfig JSON to be created that is only handled correctly by OpenShift Origin 1.0.6 and later.


vrutkovs:
Looks good

PR #418, opened 2016-05-17
rename kojitask label 'koji-task-id'

twaugh:
This is safe right now as it is not yet used.


vrutkovs:
Looks better than 'kojitask'

mmilata:
:+1: merged

PR #419, opened 2016-05-18
Fix several small problems found by quantifiedcode.com

vrutkovs:
See my fork's page at https://www.quantifiedcode.com/app/project/bd1a52c2e00c4cff90df9252c23a1328/

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -615,7 +615,7 @@ class OSBS(object):
 
     # implements subset of OpenShift's export logic in pkg/cmd/cli/cmd/exporter.go
     @staticmethod
-    def _prepare_resource(resource_type, resource):
+    def _prepare_resource(resource):
         utils.graceful_chain_del(resource, 'metadata', 'resourceVersion')
 
     @osbsapi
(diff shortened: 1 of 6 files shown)


osbs/utils.py
```diff
@@ -87,7 +87,7 @@ def write_file(self, name, content):
 
 
 class TarReader(object):
-    TarFile = namedtuple('TarFile', ['filename', 'fileobj'])
+    tar_file = namedtuple('tar_file', ['filename', 'fileobj'])
```
twaugh: Not sure I agree with this one. TarFile makes it clearer that it's a defined class/type, not an instance or method. If it really can't be a class member it could perhaps move to module scope, but that seems to give it more scope that required.

osbs/api.py
```diff
@@ -615,7 +615,7 @@ def resume_builds(self, quota_name=None):
 
     # implements subset of OpenShift's export logic in pkg/cmd/cli/cmd/exporter.go
     @staticmethod
-    def _prepare_resource(resource_type, resource):
+    def _prepare_resource(resource):
```
twaugh: I wasn't sure if this was intended for future use, but I guess it's not an API so can be removed.

osbs/utils.py
vrutkovs: Right, I shouldn't have blindly followed QC's advice on this one

mmilata:
LGTM, thank you!

PR #420, opened 2016-05-18
New optional koji_task_id parameter for list_builds.

twaugh:
This allows clients to filter listed builds by their koji-task-id label.

CC @lcarva, @breillyrh

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -420,19 +420,31 @@ class Openshift(object):
             return response.iter_lines()
         return response.content
 
-    def list_builds(self, build_config_id=None, field_selector=None):
+    def list_builds(self, build_config_id=None, koji_task_id=None,
+                    field_selector=None):
         """
         List builds matching criteria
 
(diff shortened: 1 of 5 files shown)


lcarva:
LGTM

osbs/core.py
```diff
@@ -420,19 +420,31 @@ def logs(self, build_id, follow=False, build_json=None, wait_if_missing=False):
             return response.iter_lines()
         return response.content
 
-    def list_builds(self, build_config_id=None, field_selector=None):
+    def list_builds(self, build_config_id=None, koji_task_id=None,
+                    field_selector=None):
```
breillyrh: Slightly concerned about changing the order of arguments, but it looks like optional arguments are always specified. LGTM.

twaugh: This is for the internal method. The API method for the OSBS class, the one you actually call, preserves the order (new parameter goes at the end).

twaugh: And yes, for internal usage the build_config_id parameter is always specified by name when used (1 place).

PR #423, opened 2016-05-20
run compress plugin after pulp_push

twaugh:
Now that pulp_push is smart enough to strip layers from the tarball, it is more efficient to let it do that manipulation before the tarball we'll upload to Koji is compressed.

This saves a round of decompression.

### DIFF
--- a/inputs/prod_inner.json
+++ b/inputs/prod_inner.json
@@ -80,13 +80,6 @@
   ],
   "postbuild_plugins": [
     {
-      "name": "compress",
-      "args": {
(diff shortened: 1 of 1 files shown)


vrutkovs:
+1 on this one

lcarva:
LGTM

mmilata:
LGTM

I think I've merged the old version of the commit:( Was there anything different?

twaugh:
No, just a rebase. Sorry, I didn't let you know that was coming...

PR #422, opened 2016-05-19
Add build custom base image functionality

lcarva:
Requires corresponding release of atomic-reactor with changes from: https://github.com/projectatomic/atomic-reactor/pull/493

PR #424, opened 2016-05-20
Labels parameter

lcarva:
Overlapping commit from #422 needed to avoid unit testing conflicts. Merging this will also merge #422 .

### DIFF
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -9,6 +9,7 @@ from __future__ import print_function, absolute_import, unicode_literals
 
 import logging
 import os
+import json
 import warnings
 from pkg_resources import parse_version
 
@@ -361,6 +362,15 @@ class Configuration(object):
 
     def get_proxy(self):
         return self._get_value("yum_proxy", self.conf_section, "yum_proxy")
+
+    def get_labels(self):
+        labels_list = self._get_value("labels", self.conf_section, "labels")
+        if labels_list is None:
+            return {}
+        # When read from config file, it needs to be deserialized
+        if type(labels_list) is not list:
+            labels_list = json.loads(labels_list)
+        return dict(labels_list)
 
     def get_oauth2_token(self):
         # token overrides token_file
--- a/tests/build/test_build_request.py
+++ b/tests/build/test_build_request.py
@@ -1201,6 +1201,34 @@ class TestBuildRequest(object):
         mount_path = koji_certs_secret[0]['mountPath']
         assert get_plugin(plugins, 'exit_plugins', 'koji_promote')['args']['koji_ssl_certs'] == mount_path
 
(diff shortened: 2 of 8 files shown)


tests/build/test_build_request.py
```diff
@@ -1206,6 +1234,8 @@ def test_prod_is_custom_base_image(self, tmpdir, base_image, is_custom):
         build_request.set_params(**kwargs)
         build_json = build_request.render()
 
+        phase = 'prebuild_plugins'
+        plugin = 'add_labels_in_dockerfile'
```
mmilata: These seem to be unused.

osbs/conf.py
```diff
@@ -362,6 +363,15 @@ def get_build_imagestream(self):
     def get_proxy(self):
         return self._get_value("yum_proxy", self.conf_section, "yum_proxy")
 
+    def get_labels(self):
+        labels_list = self._get_value("labels", self.conf_section, "labels")
+        if labels_list is None:
+            return {}
+        # When read from config file, it needs to be deserialized
+        if type(labels_list) is not list:
+            labels_list = json.loads(labels_list)
```
mmilata: Does this mean the config syntax is something like `labels = [("label1", "value1"), ("label2", "value2"), ("label3", "value3")]`?

Looks pretty weird to me:) But then again, we probably don't expect users to set labels in the configuration, do we? Should we just not allow setting this in the config?

tests/build/test_build_request.py
```diff
@@ -1245,3 +1188,105 @@ def test_render_prod_request_with_koji_secret(self, tmpdir):
...
+        # Safe to call prior to build image being set
+        assert build_request.is_custom_base_image() is False
+
+        kwargs = get_sample_prod_params()
+        kwargs['base_image'] = base_image
+        build_request.set_params(**kwargs)
+        build_json = build_request.render()
+
+        phase = 'prebuild_plugins'
+        plugin = 'add_labels_in_dockerfile'
```
mmilata: These seem to be unused.

mmilata:
Two comments inline, otherwise LGTM.

osbs/conf.py
lcarva: Yes. It does look odd, but I don't suppose anyone would use it in the config. We could enhance the Configuration class to control the sources of certain item, but as it stands it does not support this behavior. Is it worth adding it?

tests/build/test_build_request.py
lcarva: Removed. Good catch!

osbs/conf.py
mmilata: I think we can live with this:)

mmilata:
LGTM

twaugh:
LGTM

PR #425, opened 2016-05-22
Add a .landscape.yml to add checks in Python 3 mode

twaugh:


twaugh:
The webhook doesn't seem to have linked up right, but it did actually run:
https://landscape.io/github/projectatomic/osbs-client/850

PR #426, opened 2016-05-23
[WIP] prod_inner: use default dont_load setting for squash plugin

vrutkovs:
Currently image-id stored in annotations doesn't match the pulp image id
as we're not loading the squashed image back to Docker


mmilata:
LGTM!

mmilata:
Can you please open new PR for this? I don't think this one can be reopened.

PR #427, opened 2016-05-23
prod_inner: use default squash plugin settings

vrutkovs:
Currently image-id stored in annotations doesn't match the pulp image id
as we're not loading the squashed image back to Docker

A duplicate of #426

### DIFF
--- a/inputs/prod_inner.json
+++ b/inputs/prod_inner.json
@@ -71,11 +71,7 @@
   ],
   "prepublish_plugins": [
     {
-      "name": "squash",
-      "args": {
-        "remove_former_image": false,
-        "dont_load": true
-      }
+      "name": "squash"
     }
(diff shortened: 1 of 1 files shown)


twaugh:
I just realised: we should also remove the `remove_former_image` parameter as well so we don't end up accumulating unsquashed images in the docker engine.

twaugh:
...which means we can also remove `args`, i.e.:

```
  "prepublish_plugins": [
    {
      "name": "squash"
    }
  ],
```

vrutkovs:
>we should also remove the remove_former_image parameter

That didn't work for me - docker complained that this image is used. I hope this is fixed in latest docker-squash

twaugh:
LGTM, tested here fine.

PR #429, opened 2016-05-25
Don't crash while checking for labels in inexisting buildconfig

vrutkovs:
Reproduced on scratchbuild server

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -226,7 +226,7 @@ class OSBS(object):
         - metadata.labels.git-repo-name AND metadata.labels.git-branch are equal
         """
 
-        git_labels = [(key, build_config['metadata']['labels'][key])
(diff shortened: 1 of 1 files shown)


twaugh:
This looks like the wrong fix. `build_config` at that point is one we just rendered, so it ought to have those labels set.

vrutkovs:
Right, the title is wrong - on scratch server we always create a new buildconfig - and ```_create_build_config_and_build``` takes build config from build request

mmilata:
The simple build also checks for existing buildconfig if i'm reading the code correctly. I've created PR #435 that attempts to fix this by adding the labels for simple builds as well.

I'm working on removing the build types entirely (#129) but not sure when it's going to be ready, would be nice if simple builds worked in the meantime.

vrutkovs:
Closed as #435 is merged

PR #432, opened 2016-06-01
compress: don't load the saved image back - it should be handled by squash

vrutkovs:


twaugh:
Actually the parameter is named misleadingly -- setting it to True actually means less to do (it uses the 'docker save' tarball produced by squash).

So it's fine as it is.

PR #434, opened 2016-06-02
hide logs and packages in get-build output if those are unavailable

### LINKED ISSUE #433: Show output of 'build-logs' in 'get-build'
Since we don't store logs in annotations anymore we should either show logs (same as in 'build-logs') or remove this section in 'get-build' output

vrutkovs:
Fixes #433

### DIFF
--- a/osbs/build/build_response.py
+++ b/osbs/build/build_response.py
@@ -97,7 +97,7 @@ class BuildResponse(object):
         """
         logs = graceful_chain_get(self.get_annotations_or_labels(), "logs")
         if not logs:
-            logger.error("no logs")
+            logger.debug("no logs found in annotations")
             return ""
         if decode_logs:
             output = []
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -171,8 +171,8 @@ def cmd_get_build(args, osbs):
         print_json_nicely(build_json)
     elif args.output == 'text':
         repositories_dict = build.get_repositories()
-        repositories_str = None
-        if repositories_dict is not None:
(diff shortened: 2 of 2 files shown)


osbs/cli/main.py
```diff
@@ -206,15 +223,7 @@ def cmd_get_build(args, osbs):
...
-
-            BUILD LOGS
-
-            {logs}
-
-            PACKAGES
-
-            {packages}
-
+            {logs}{packages}
```
lcarva: Are we missing newlines here? `"""\` won't add a new line.

vrutkovs: I'm not sure why but it seems each empty line is rendered with ```\n```, see [sample output](https://gist.github.com/vrutkovs/dffd745242230137e0e3346b12a08259).

osbs/build/build_response.py
```diff
@@ -97,7 +97,7 @@ def get_logs(self, decode_logs=True):
         """
         logs = graceful_chain_get(self.get_annotations_or_labels(), "logs")
         if not logs:
-            logger.error("no logs")
+            logger.warn("no logs found in annotations")
```
mmilata: If we expect this to normally happen with future builds then let's use some other log level - debug?

mmilata:
Otherwise LGTM

osbs/build/build_response.py
vrutkovs: Agreed, will change that

mmilata:
LGTM, merged, thank you.

PR #435, opened 2016-06-03
Don't crash when attempting scratch/simple build

mmilata:
Related to #429

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -300,6 +300,14 @@ class CommonBuild(BuildRequest):
             self.template['spec']['strategy']['customStrategy']['from']['name'] = \
                 self.spec.build_image.value
 
+        repo_name = git_repo_humanish_part_from_uri(self.spec.git_uri.value)
+        # NOTE: Since only the repo name is used, a forked repos will have
+        # the same git-repo-name tag. This is a known limitation. If this
+        # use case must be handled properly, the git URI must be taken into
+        # account.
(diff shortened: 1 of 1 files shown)


osbs/build/build_request.py
```diff
@@ -794,12 +802,6 @@ def render(self, validate=True):
             self.spec.validate()
         super(ProductionBuild, self).render()
 
-        repo_name = git_repo_humanish_part_from_uri(self.spec.git_uri.value)
-        # NOTE: Since only the repo name is used, a forked repos will have
-        # the same git-repo-name tag. This is a known limitation. If this
-        # use case must be handled properly, the git URI must be taken into
-        # account.
-        self.set_label('git-repo-name', repo_name)
         self.set_label('git-branch', self.spec.git_branch.value)
```
lcarva: Why not move this one?

mmilata: Branch is only mandatory for production builds, though I can't remember why it's the case, maybe we can make it mandatory for all builds?

twaugh:
We've talked about getting rid of build types before. I wonder if we should change Simple to Scratch?

mmilata:
@twaugh I'm working on that but it's not ready yet - I realized we probably need to distinguish build types at least for the purpose of input validation (e.g. prod build needs the sources command but simple build doesn't). Or should this be the responsibility of koji-containerbuild?

This patch is supposed to un-break simple builds in the meantime.

PR #439, opened 2016-06-08
new 'labels' optional parameter for create_prod_build()

twaugh:
This parameter provides overrides for the labels in the Dockerfile. It cannot be set in the configuration file but can be provided at the command line.

Not providing configuration file compatibility because (a) it was never documented, and (b) config values would not get merged with CLI/kwarg values.

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -278,8 +278,6 @@ def cmd_build(args, osbs):
         target=osbs.build_conf.get_koji_target(),
         architecture=osbs.build_conf.get_architecture(),
         yum_repourls=osbs.build_conf.get_yum_repourls(),
-        build_image=osbs.build_conf.get_build_image(),
-        build_imagestream=osbs.build_conf.get_build_imagestream(),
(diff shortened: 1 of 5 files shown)


lcarva:
LGTM

osbs/cli/main.py
```diff
@@ -263,6 +263,11 @@ def cmd_cancel_build(args, osbs):
 
 
 def cmd_build(args, osbs):
+    if args.labels:
+        labels = dict(args.labels)
+    else:
+        labels = None
```
vrutkovs: Nitpick: wouldn't label var have a wrong var scope here? I guess we should set it to None first and then make a dict if ```args.labels``` is set

lcarva: Python doesn't have the concept of block scoping. In this case it's function scope.

vrutkovs: Thanks, these statements always confuse me after several years of Ruby

vrutkovs:
LGTM, I tested a similar patch. Merging

PR #441, opened 2016-06-10
Enable tag_from_config plugin

lcarva:
This can be merged, but a new release would require a release of atomic-reactor containing the following changes: https://github.com/projectatomic/atomic-reactor/pull/515

### DIFF
--- a/inputs/prod_inner.json
+++ b/inputs/prod_inner.json
@@ -85,6 +85,9 @@
       "name": "tag_by_labels"
     },
     {
+      "name": "tag_from_config"
+    },
+    {
       "name": "tag_and_push",
       "args": {
         "registries": {


vrutkovs:
LGTM

mmilata:
I think atomic-reactor won't fail when build json contains unknown plugin ... let me check.

mmilata:
Ok, it's not the case, whole build fails when unknown plugin is encountered.

I'd wait for a-r release and then merge it. Or we can enable the plugin conditionally based on osbs.conf. Ideas?

lcarva:
I think it's safe to merge, it's just the release that must be synchronized properly.

vrutkovs:
Agreed, its safe to merge (just to be sure we can make an interim release of atomic-reactor)

mmilata:
Alright, merged.

PR #440, opened 2016-06-10
Support setting email/login/email for registries in tag_and_push plugin

vrutkovs:
Required change to enable https://github.com/projectatomic/atomic-reactor/pull/466

This also enables setting several secrets for a plugin

* [x] Add tests

PR #443, opened 2016-06-17
Support setting email/login/email for registries in tag_and_push plugin

lcarva:
This is the same PR as https://github.com/projectatomic/osbs-client/pull/440 except it's been rebased so it can be merged.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -170,12 +170,16 @@ class BuildRequest(object):
             placeholder = '{{REGISTRY_URI}}'
 
             if placeholder in registries:
-                for registry in self.spec.registry_uris.value:
+                for i, registry in enumerate(self.spec.registry_uris.value):
                     if not registry.uri:
                         continue
 
                     regdict = registries[placeholder].copy()
                     regdict['version'] = registry.version
+                    registry_values = self.spec.registry_secrets.value
+                    if isinstance(registry_values, list) and i < len(registry_values):
+                        regdict['secret'] = os.path.join(SECRETS_PATH, registry_values[i])
+
                     registries[registry.docker_uri] = regdict
 
                 del registries[placeholder]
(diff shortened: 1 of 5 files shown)


vrutkovs:
+1, lets merge this one

twaugh:
Is there a reason not to use a "kubernetes.io/dockercfg"-type secret?

The advantage of the dockercfg secret type is that Kubernetes already knows about it, so we could use "oc secrets new-dockercfg ..." to create it, and "oc describe secret ..." to see it.

vrutkovs:
That's the type we should be using (at least, I tested it in previous PR). Haven't tested if this works yet though

osbs/build/build_request.py
```diff
@@ -728,7 +738,10 @@ def render(self, validate=True):
                           self.spec.koji_certs_secret.value,
 
                           ('prebuild_plugins', 'add_filesystem', 'koji_ssl_certs_dir'):
-                          self.spec.koji_certs_secret.value})
+                          self.spec.koji_certs_secret.value,
+
+                          ('postbuild_plugins', 'tag_and_push', 'secret'):
```
twaugh: But tag_and_push doesn't have a keyword argument named 'secret'.

twaugh: Oh, I see, it's not added to the configuration for 'tag_and_push'. I think we should restructure that to make it more explicit, maybe by setting the argument name to None instead of 'secret', and skipping the dock_json_set_arg call in set_secret_for_plugin in that case.

twaugh: ...and adding a comment here explaining what is and isn't being done.

twaugh: Cherry-pickable commit showing what I mean:
https://github.com/projectatomic/osbs-client/commit/2128704b51cb68a26f751cba565ceda183bfc178

vrutkovs:
Closing this one as #445 was merged

PR #444, opened 2016-06-21
buildresponse: return a meaningful error message

vrutkovs:
This is handy for koji-containerbuild to return a more descriptive error message

### DIFF
--- a/osbs/build/build_response.py
+++ b/osbs/build/build_response.py
@@ -119,6 +119,22 @@ class BuildResponse(object):
         else:
             return logs
 
+    def get_error_message(self):
+        """
+        Return an error message based on atomic-reactor's metadata
+        """
+        try:
+            str_metadata = graceful_chain_get(self.get_annotations_or_labels(), "plugins-metadata")
+            metadata_dict = json.loads(str_metadata)
+            plugin, error_message = list(metadata_dict['errors'].items())[0]
+            if error_message:
+                # Plugin has non-empty error description
+                return "Error in plugin %s: %s" % (plugin, error_message)
+            else:
+                return "Error in plugin %s" % plugin
+        except Exception:
+            return None
+
     def get_commit_id(self):
         return graceful_chain_get(self.get_annotations_or_labels(), "commit_id")
 
--- a/tests/build/test_build_response.py
+++ b/tests/build/test_build_response.py
@@ -6,6 +6,7 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 import json
+import pytest
 from osbs.build.build_response import BuildResponse
 
 class TestBuildResponse(object):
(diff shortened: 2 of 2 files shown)


osbs/build/build_response.py
```diff
@@ -119,6 +119,19 @@ def get_logs(self, decode_logs=True):
         else:
             return logs
 
+    def get_error_message(self):
+        try:
+            str_metadata = graceful_chain_get(self.get_annotations_or_labels(), "plugins-metadata")
+            metadata_dict = json.loads(str_metadata)
+            error = metadata_dict['errors'].items()[0]
```
twaugh: How about `plugin, error = ` since we know items() returns 2-tuples? It might make the next few lines easier to read.

vrutkovs: Good catch, updating

osbs/build/build_response.py
```diff
@@ -119,6 +119,19 @@ def get_logs(self, decode_logs=True):
...
+            if error[1]:
+                # Plugin has non-empty error description
+                return "Error in plugin %s: %s" % (error[0], error[1])
+            else:
+                return "Error in plugin %s" % (error[0])
+        except:
+            return ""
```
twaugh: Should we return `None` when there is no error?

Would be good to have a docstring for the method too.

mmilata: What's the point of masking all errors, shouldn't this be done by the caller?

If we decide to keep it I'd suggest at least using `except Exception` so as to not block things like signals.

vrutkovs: >Should we return None when there is no error?

Not sure, I'd rather have error handled here than in koji-containerbuild

osbs/build/build_response.py
```diff
@@ -119,6 +119,22 @@ def get_logs(self, decode_logs=True):
...
             return logs
 
+    def get_error_message(self):
+        """
+        Return an error message based on atomic-reactor's metadata
+        """
+        try:
+            str_metadata = graceful_chain_get(self.get_annotations_or_labels(), "plugins-metadata")
+            metadata_dict = json.loads(str_metadata)
+            plugin, error_message = list(metadata_dict['errors'].items())[0]
```
lcarva: why only choose the first one?

twaugh: It's reasonable I think. The first error changed the build from succeeding to failing.

vrutkovs: We don't collect error messages from non-critical plugins, so the first error would be the only one

lcarva: It's pulling the first one from a `dict`, not `OrderedDict`. It's not necessarily the first error recorded.

Also I think conversion to `list` is redundant in this case.

vrutkovs: Convertion to ```list``` is required for Py3 - ```items()``` is a generator there

vrutkovs:
Tested on a local koji instance

lcarva:
LGTM

twaugh:
LGTM

PR #445, opened 2016-06-21
Support setting email/login/email for registries in tag_and_push plugin

vrutkovs:
This also enables setting several secrets for a plugin and adds test.

Supersedes #443, as it contains suggestions from Tim and Luiz

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -16,9 +16,11 @@ from pkg_resources import parse_version
 try:
     # py2
     import urlparse
+    from itertools import izip_longest as zip_longest
 except ImportError:
     # py3
     import urllib.parse as urlparse
+    from itertools import zip_longest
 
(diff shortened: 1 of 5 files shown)


osbs/build/build_request.py
```diff
@@ -234,7 +240,9 @@ def set_secret_for_plugin(self, plugin, secret):
                         'mountPath': secret_path,
                     })
 
-                self.dj.dock_json_set_arg(*(plugin + (secret_path,)))
+                # tag_and_push sets secret path for each registry separately
```
twaugh: Just spotted this comment. Should probably be 'if no argument name supplied, leave config alone'.

vrutkovs: Agreed

twaugh:
LGTM, just a stale comment, noted inline.

twaugh:
LGTM

PR #446, opened 2016-06-21
Configure pulp_sync with registry secret if provided

twaugh:
Depends on https://github.com/projectatomic/atomic-reactor/pull/521 so once merged we should wait until atomic-reactor is released with that change before releasing osbs-client.


lcarva:
LGTM

PR #448, opened 2016-06-27
Set git branch to empty string instead of None

vrutkovs:
This unblocks builds using rhpkg, which doesn't pass git branch parameter

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -151,6 +151,8 @@ class BuildRequest(object):
         return False
 
     def set_label(self, name, value):
+        if not value:
+            value = ''
         self.template['metadata'].setdefault('labels', {})
         self.template['metadata']['labels'][name] = value
 
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -694,7 +694,10 @@ class BuildRequest(object):
         # use case must be handled properly, the git URI must be taken into
         # account.
         self.set_label('git-repo-name', repo_name)
-        self.set_label('git-branch', self.spec.git_branch.value)
+        if self.spec.git_branch.value:
+            self.set_label('git-branch', self.spec.git_branch.value)
+        else:
+            self.set_label('git-branch', 'unknown')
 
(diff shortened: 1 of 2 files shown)


twaugh:
I think it should default to 'unknown' instead of the empty string, and the equivalent code in osbs/utils.py:make_name_from_git() should be removed.

Is testing this hard?

vrutkovs:
Sure, sounds like a good idea to me. Testing it is fairly easy on Jenkins.

I wonder if using something like ```git branch --contains <commit id>``` is a good idea. What do you guys think?

twaugh:
We looked at this a while ago, and there's basically no way to infer the branch name in the general case. For example: imagine when two branches are at the same commit.

vrutkovs:
Yes, this and introduction of python git bindings are my biggest concerns.

I'll set this to 'unknown' for now

osbs/utils.py
```diff
@@ -311,7 +311,7 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
     """
 
     repo = git_repo_humanish_part_from_uri(repo)
-    branch = branch or 'unknown'
+    branch = branch
```
twaugh: no-op :-)

twaugh:
FWIW, if we can't get it working so that triggers are removed correctly when no branch is passed in, and easy alternative fix would be to do the equivalent default 'unknown' substitution when setting the annotation.

vrutkovs:
Right, moved this part to the ```set_label``` part, otherwise autorebuild is being triggered.

Tested on local koji instance

twaugh:
Can we make the test_render_prod_request_trigger_missing_param() test-case verify that the git-branch label is never set to None?

twaugh:
LGTM

lcarva:
I wonder if `set_label` should be modified to never allow `None` values.

twaugh:
Couldn't hurt.

twaugh:
LGTM

lcarva:
LGTM

mmilata:
LGTM

PR #451, opened 2016-06-30
Remove labels overwrite functionality

lcarva:


twaugh:
This removes the `labels` parameter that koji-containerbuild is currently using, so applying this change on a Koji builder would result in failures.

Don't we need to adjust koji-containerbuild first, or else accept the `labels` parameter and ignore it?

lcarva:
I'll do some additional testing, but k-cb uses the create_build API method
which takes arbitrary keyword arguments. It should simply ignore unknown
ones. Also added a unit test to verify this behavior.

twaugh:
Oh, good point.

tests/test_api.py
```diff
@@ -68,6 +68,34 @@ def test_get_pod_for_build(self, osbs):
...
+            'git_branch': TEST_GIT_BRANCH,
+            'user': TEST_USER,
+            'component': TEST_COMPONENT,
+            'target': TEST_TARGET,
+            'architecture': TEST_ARCH,
+            'yum_repourls': None,
+            'koji_task_id': None,
+            'scratch': False,
+            # Stuff that should be ignored and not cause erros
+            'labels': {'Release': 'bacon'},
```
twaugh: OK, I didn't spot this before.

twaugh:
LGTM

PR #452, opened 2016-07-04
Create a new Build for scratch builds

vrutkovs:
This creates a new build on existing BuildConfig (if found) instead of updating BuildConfig

TODO:
 * [x] Add a new config option for scratch builds (after build_types were removed)
 * [x] Add tests
 * [x] Verify on several existing projects

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -12,6 +12,7 @@ import logging
 import os
 import sys
 import warnings
+import datetime
 from functools import wraps
 
 from osbs.build.build_request import BuildRequest
@@ -259,6 +260,19 @@ class OSBS(object):
                 msg = self._panic_msg_for_more_running_builds(build_config_name, running_builds)
             raise OsbsException(msg)
 
+    def _create_scratch_build(self, build_request):
+        logger.debug(build_request)
+        build_json = build_request.render()
+        build_json['kind'] = 'Build'
+        if 'spec' not in build_json.keys():
+            build_json['spec'] = {}
+        build_json['spec']['serviceAccount'] = 'builder'
+        build_json['metadata']['labels']['scratch'] = 'true'
+        build_config_name = 'scratch-%s' % datetime.datetime.now().strftime('%Y%m%d%H%M%S')
+        logger.debug('starting scratch build %s', build_config_name)
+        build_json['metadata']['name'] = build_config_name
+        return BuildResponse(self.os.create_build(build_json).json())
+
     def _create_build_config_and_build(self, build_request):
         build = None
 
@@ -314,7 +328,7 @@ class OSBS(object):
                           target,      # may be None
                           architecture=None, yum_repourls=None,
                           koji_task_id=None,
-                          scratch=False,
+                          scratch=None,
                           **kwargs):
         """
         Create a production build
@@ -381,10 +395,13 @@ class OSBS(object):
             git_push_url=self.build_conf.get_git_push_url(),
             git_push_username=self.build_conf.get_git_push_username(),
             builder_build_json_dir=self.build_conf.get_builder_build_json_store(),
-            scratch=scratch,
+            scratch=self.build_conf.get_scratch(scratch),
         )
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
-        response = self._create_build_config_and_build(build_request)
+        if build_request.scratch:
+            response = self._create_scratch_build(build_request)
+        else:
+            response = self._create_build_config_and_build(build_request)
         logger.debug(response.json)
         return response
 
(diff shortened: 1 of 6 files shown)


twaugh:
Why look for an existing BuildConfig in that situation? We already have everything we need to submit a new build don't we?

twaugh:
I think it might be better if `create_prod_build()` only called `self.os.create_build()` when scratch is set, and `_create_build_config_and_build()` otherwise.

vrutkovs:
> We already have everything we need to submit a new build don't we?

Indeed, if we specify BuildConfig there the build will be displayed as belonging to this buildconfig. Otherwise it will create a new buildconfig

OTOH I'm not entirely sure about this - wouldn't it be better to create a new ```scratchbuilds``` buildconfig in case we'll decide to use the same namespace for scratch builds?

tests/test_api.py
```diff
@@ -722,3 +729,41 @@ def test_create_build_config_auto_start(self):
...
+        updated_build_json['spec'] = {}
+        updated_build_json['spec']['serviceAccount'] = 'builder'
+        build_name = 'scratch-%s' % datetime.datetime.now().strftime('%Y%m%d%H%M%S')
+        updated_build_json['metadata']['name'] = build_name
+
+        (flexmock(osbs.os)
+            .should_receive('create_build')
+            .with_args(updated_build_json)
+            .once()
+            .and_return(flexmock(json=lambda: {'spam': 'maps'})))
```
twaugh: Could add a flexmock(osbs.os).should_receive('create_build_config').never() to make that explicit.

vrutkovs: Good catch, added

osbs/api.py
```diff
@@ -381,10 +394,13 @@ def create_prod_build(self, git_uri, git_ref,
             git_push_url=self.build_conf.get_git_push_url(),
             git_push_username=self.build_conf.get_git_push_username(),
             builder_build_json_dir=self.build_conf.get_builder_build_json_store(),
-            scratch=scratch,
+            scratch=self.build_conf.get_scratch(),
```
twaugh: Careful here. This changes means that create_prod_build(..., scratch=True) will no longer have any effect. Instead, the scratch parameter would have to be passed to the Configuration() constructor, or put in a config file.

We could actually solve this problem by having get_scratch() take a parameter to use as the default, and giving it the 'scratch' we were passed here as its parameter.

osbs/api.py
```diff
@@ -259,6 +260,18 @@ def _verify_no_running_builds(self, build_config_name):
                 msg = self._panic_msg_for_more_running_builds(build_config_name, running_builds)
             raise OsbsException(msg)
 
+    def _create_scratch_build(self, build_request):
+        build_json = build_request.render()
+        build_json['kind'] = 'Build'
+        if 'spec' not in build_json.keys():
+            build_json['spec'] = {}
+        build_json['spec']['serviceAccount'] = 'builder'
+        build_json['metadata']['labels']['scratch'] = 'true'
```
twaugh: +1

osbs/build/build_request.py
```diff
@@ -50,7 +50,7 @@ def __init__(self, build_json_store):
         self._resource_limits = None
         self._openshift_required_version = parse_version('1.0.6')
         # For the koji "scratch" build type
-        self.scratch = False
+        self.scratch = None
```
twaugh: Curious -- why the change here?

twaugh: Oh, I get it, I think. Don't provide a value, so as to avoid overriding it if it is set elsewhere.

vrutkovs: This is done to make sure config param doesn't get overwritten by default params.

tests/test_api.py
```diff
@@ -722,3 +729,49 @@ def test_create_build_config_auto_start(self):
 
         build_response = osbs._create_build_config_and_build(build_request)
         assert build_response.json == {'spam': 'maps'}
+
+    def test_scratch_build_config(self):
```
twaugh: Could we add a test for passing 'scratch=True' to the create_build() API method?

vrutkovs: Added ```test_scratch_param_to_create_build``` test case

twaugh:
LGTM

osbs/cli/main.py
```diff
@@ -566,7 +566,7 @@ def cli():
     build_parser.add_argument("--storage-limit", action='store', required=False,
                               help="storage limit")
     build_parser.add_argument("--scratch", action='store_true', required=False,
-                              default=False, help="perform a scratch build")
```
lcarva: You could remove the `default` param as its default value is `None`.

lcarva:
LGTM

PR #453, opened 2016-08-15
prod_inner: run all_rpm_packages plugin before pushing image

twaugh:


lcarva:
LGTM
Why have we not done this before!?

PR #455, opened 2016-09-09
Add support for unique_tag_only config

lcarva:
When `unique_tag_only` is set to True, atomic-reactor will be configured to ensure built image is only tagged with its unique timestamp tag.

If `unique_tag_only` is either not set or set to False, behavior remains unchanged. This should allow using an older atomic-reactor version than the one containing these changes: https://github.com/projectatomic/atomic-reactor/pull/554

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -396,6 +396,7 @@ class OSBS(object):
             git_push_username=self.build_conf.get_git_push_username(),
             builder_build_json_dir=self.build_conf.get_builder_build_json_store(),
             scratch=self.build_conf.get_scratch(scratch),
+            unique_tag_only=self.build_conf.get_unique_tag_only(),
         )
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
         if build_request.scratch:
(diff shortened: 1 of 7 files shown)


vrutkovs:
Would it be a good idea to set this option to True if `scratch = True` is set in the config also? 

As those option seem to be here just for scratch builds - wouldn't it be simplier to merge those anyway?

lcarva:
Although they're only used for scratch builds now, there's nothing that prevents it from using it on regular builds. I'd rather not couple the two together unless it's necessary. 

As part of integration test, I also tested a non-scratch build with unique_tag_only set to True.

vrutkovs:
Right, agreed. We shouldn't be adding way too many options - that's my concern here.

Anyway, LGTM

lcarva:
Another point to my previous comment is that fedora uses a completely different registry for scratch builds, and they will likely choose to not use `unique_tag_only` and simply have all tags applied to a scratch build.

This was enough for me to think these should be disparate features.

twaugh:
LGTM

PR #421, opened 2016-05-18
handle connection being closed while streaming logs

twaugh:
When the connection is closed, if the time since the last activity is less than a minute, finish.

Otherwise, ask again for logs starting from just after we last saw activity.

PR #458, opened 2016-09-21
Rewrite http.py to use python-requests

### LINKED ISSUE #51: Nondescript IOError Exception

puiterwijk:
This will work around an issue when pycurl is used in the same context as
rpm is, as used by koji-containerbuild.

Fixes #51

Signed-off-by: Patrick Uiterwijk puiterwijk@redhat.com

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -11,7 +11,6 @@ import os
 import numbers
 import time
 import base64
-import pycurl
 
 import logging
 from osbs.kerberos_ccache import kerberos_ccache_init
@@ -21,8 +20,8 @@ from osbs.constants import WATCH_MODIFIED, WATCH_DELETED, WATCH_ERROR
 from osbs.constants import (SERVICEACCOUNT_SECRET, SERVICEACCOUNT_TOKEN,
                             SERVICEACCOUNT_CACRT)
 from osbs.exceptions import (OsbsResponseException, OsbsException,
-                             OsbsWatchBuildNotFound, OsbsAuthException,
-                             OsbsNetworkException)
+                             OsbsWatchBuildNotFound, OsbsAuthException)
+from osbs.http import decoded_json
 from osbs.utils import graceful_chain_get
 
 try:
@@ -47,7 +46,7 @@ def check_response(response):
         if hasattr(response, 'content'):
             content = response.content
         else:
-            content = ''.join(response.iter_lines())
+            content = ''.join(decoded_json(response.iter_lines()))
 
         logger.error("[%d] %s", response.status_code, content)
         raise OsbsResponseException(message=content, status_code=response.status_code)
@@ -408,17 +407,13 @@ class Openshift(object):
             try:
                 response = self._get(buildlogs_url, stream=1,
                                      headers={'Connection': 'close'})
-            except OsbsNetworkException as ex:
-                # pycurl reports 'empty reply from server' as
-                # FOLLOWLOCATION. Handle this as though there were no
-                # lines returned.
-                if ex.status_code != pycurl.FOLLOWLOCATION:
-                    raise
-            else:
                 check_response(response)
-                for line in response.iter_lines():
+
+                for line in decoded_json(response.iter_lines()):
                     last_activity = time.time()
                     yield line
+            except httplib.IncompleteRead:
+                pass
 
             idle = time.time() - last_activity
             logger.debug("connection closed after %ds", idle)
@@ -554,7 +549,7 @@ class Openshift(object):
         while True:
             with self._get(url, stream=True, headers={'Connection': 'close'}) as response:
                 check_response(response)
-                for line in response.iter_lines():
+                for line in decoded_json(response.iter_lines()):
                     logger.debug(line)
                     try:
                         j = json.loads(line)
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -7,94 +7,35 @@ of the BSD license. See the LICENSE file for details.
 
 
 abstraction on top of http api calls
-
-use pycurl (can handle chunked response properly)
-
-chunked implementation for pycurl taken from:
-  http://stackoverflow.com/a/21809888/909579
 """
 
 from __future__ import print_function, absolute_import, unicode_literals
 
-import re
 import sys
 import json
-import time
-import codecs
 import logging
-from io import BytesIO
-
-import pycurl
 
 from osbs.exceptions import OsbsException, OsbsNetworkException, OsbsResponseException
 
+import requests
 try:
-    # py2
-    import httplib
+    from requests_kerberos import HTTPKerberosAuth
 except ImportError:
-    # py3
-    import http.client as httplib
-
+    HTTPKerberosAuth = None
 
 logger = logging.getLogger(__name__)
 
-SELECT_TIMEOUT = 9999
-PYCURL_NETWORK_CODES = [pycurl.E_BAD_CONTENT_ENCODING,
-                        pycurl.E_BAD_DOWNLOAD_RESUME,
-                        pycurl.E_CONV_FAILED,
-                        pycurl.E_CONV_REQD,
-                        pycurl.E_COULDNT_CONNECT,
-                        pycurl.E_COULDNT_RESOLVE_HOST,
-                        pycurl.E_COULDNT_RESOLVE_PROXY,
-                        pycurl.E_FILESIZE_EXCEEDED,
-                        pycurl.E_HTTP_POST_ERROR,
-                        pycurl.E_HTTP_RANGE_ERROR,
-                        pycurl.E_HTTP_RETURNED_ERROR,
-                        pycurl.E_LOGIN_DENIED,
-                        # old pycurl: E_OPERATION_TIMEOUTED, new pycurl: E_OPERATION_TIMEDOUT
-                        getattr(pycurl, "E_OPERATION_TIMEDOUT", "E_OPERATION_TIMEOUTED"),
-                        pycurl.E_PARTIAL_FILE,
-                        pycurl.E_READ_ERROR,
-                        pycurl.E_RECV_ERROR,
-                        pycurl.E_REMOTE_FILE_NOT_FOUND,
-                        pycurl.E_SEND_ERROR,
-                        pycurl.E_SSL_CACERT,
-                        pycurl.E_SSL_CERTPROBLEM,
-                        pycurl.E_SSL_CIPHER,
-                        pycurl.E_SSL_CONNECT_ERROR,
-                        pycurl.E_SSL_PEER_CERTIFICATE,
-                        pycurl.E_SSL_SHUTDOWN_FAILED,
-                        pycurl.E_TOO_MANY_REDIRECTS,
-                        pycurl.E_UNSUPPORTED_PROTOCOL,
-                        pycurl.E_WRITE_ERROR]
 
-PYCURL_NETWORK_CODES = [x for x in PYCURL_NETWORK_CODES if x is not None]
+def decoded_json(iterable):
+    """Return str objects for each JSON string in iterable
 
-
-def parse_headers(all_headers):
-    # all_headers contains headers from all responses - even without FOLLOWLOCATION there
-    # might be multiple sets of headers due to 401 Unauthorized. We only care about the last
-    # response.
-    try:
-        raw_headers = all_headers.split(b"\r\n\r\n")[-2]
-    except IndexError:
-        logger.warning('Incorrectly terminated http headers')
-        raw_headers = all_headers
-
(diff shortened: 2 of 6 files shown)


osbs/http.py
```diff
@@ -35,66 +33,15 @@
     # py3
     import http.client as httplib
```
lcarva: This is only used at the end of this file to access status code constants.
You can remove `httplib` import and instead use:

```python
requests.codes.OK
requests.codes.CREATED
```

lcarva:
`tests/test_http.py` needs to be adjusted as well

vrutkovs:
We're also gonna need to update the spec file to require `python-requests-kerberos`

twaugh:
I love how much code this removes...

lcarva:
I think we need to add `requests` and `requests-kerberos` to requirements.txt

twaugh:
Good point -- and remove pycurl.

lcarva:
pycurl is still being used in core.py, that should probably change too since it's expecting http.py to use pycurl

vrutkovs:
Agree with Lui: `core.py` (and other parts probably) should be updated to trap `requests` exception

puiterwijk:
The other files are now also updated to no longer depend on pycurl.

puiterwijk:
@vrutkovs Are you sure that this was with the latest version of my PR?
As I had indeed hit upon this issue, and this is fixed in one of the more recent revisions by a manual encoding in the python3 case:

```
+        if isinstance(self.content, bytes):
+            self.content = self.content.decode('utf-8') 
```

vrutkovs:
@puiterwijk I am, so the symptoms I see are:
1) If I use `build-logs -f` it crashes - as `watch_resource` uses streaming output, so `response` there is `HttpStream`, not `HttpResponse`.
2) If I use `build-logs` (without `--follow`) the logs are printed as one large bytes string

vrutkovs:
Two commits to fix this:
* https://github.com/vrutkovs/osbs-client/commit/3ffd5d3329dc1894a592d61c688a5d7de6a12083
* https://github.com/vrutkovs/osbs-client/commit/de690e332901c24e2b2c292fec380e5a95fe0e14

puiterwijk:
@vrutkovs Thanks for those fixes!

vrutkovs:
@twaugh @lcarva please, review and retest so we could finally merge this.

Patrick, thanks!

osbs/core.py
```diff
@@ -456,6 +456,8 @@ def logs(self, build_id, follow=False, build_json=None, wait_if_missing=False):
         buildlogs_url = self._build_url("builds/%s/log/" % build_id)
         response = self._get(buildlogs_url, headers={'Connection': 'close'})
         check_response(response)
+        if isinstance(response.content, bytes):
+            return response.content.decode('utf-8')
```
twaugh: How do we want to handle errors? What if the logs contain invalid UTF-8?

puiterwijk:
@vrutkovs I had to revert a small part of your patch, since the HttpRersponse class still exports it as .content to not break the "external" API.

vrutkovs:
@puiterwijk agreed, was about to suggest the same idea :)

puiterwijk:
@vrutkovs hah. Please tell me how you got a time machine!
"vrutkovs approved these changes 31 seconds from now " :-)

osbs/core.py
```diff
@@ -405,20 +404,12 @@ def stream_logs(self, build_id):
...
-                # lines returned.
-                if ex.status_code != pycurl.FOLLOWLOCATION:
-                    raise
-            else:
-                check_response(response)
-                for line in response.iter_lines():
-                    last_activity = time.time()
-                    yield line
+            response = self._get(buildlogs_url, stream=1,
+                                 headers={'Connection': 'close'})
```
twaugh: The code this replaces took a while to get right. The particular case where we needed to catch the FOLLOWLOCATION error was when handling chunked data where the server only ever sends headers and finally closes the connection.

This can happen when e.g. watching logs on a OpenShift Build, when no logs are produced for 300 seconds.
https://github.com/projectatomic/osbs-client/pull/421

Does `requests` not treat this as an error?

twaugh:
Do we know if this works using Python 2.6?

osbs/core.py
twaugh: I just tested this. `requests` *does* treat this as an error:
```
  File "/usr/lib/python2.7/site-packages/requests/api.py", line 71, in get
    return request('get', url, params=params, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/api.py", line 57, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 475, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 617, in send
    r.content
  File "/usr/lib/python2.7/site-packages/requests/models.py", line 741, in content
    self._content = bytes().join(self.iter_content(CONTENT_CHUNK_SIZE)) or bytes()
  File "/usr/lib/python2.7/site-packages/requests/models.py", line 667, in generate
    raise ChunkedEncodingError(e)
requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(0 bytes read)', IncompleteRead(0 bytes read))
```

You can test it yourself using this:
```
from http.server import HTTPServer, BaseHTTPRequestHandler

class ChunkServer(BaseHTTPRequestHandler):
    def do_GET(self):
        self.send_response(200)
        self.send_header('Transfer-Encoding', 'chunked')
        self.end_headers()

chunk = HTTPServer(('', 5700), ChunkServer)
chunk.serve_forever()
```

We need to make sure this case is handled correctly and that osbs-client will re-connect in this situation.

twaugh:
I guess osbs-client could do with a unit test for handling chunk decode failures gracefully.

osbs/core.py
vrutkovs: Fixed in https://github.com/vrutkovs/osbs-client/commit/a93fe430fce96f39e7fb7201b7dd1fb54a6d1f2c

vrutkovs:
Tested in py2.6 (in centos6 container) - `build`, `get-build`, `watch-builds` and `build-logs` worked fine

twaugh:
Still needs the error handling fix to actually be in this PR. Patrick, could you please cherry pick it (vrutkovs@a93fe43)?

twaugh:
I think this is ready to merge. Anyone have objections?

osbs/http.py
```diff
@@ -160,203 +79,68 @@ def __init__(self, url, method, data=None, kerberos_auth=False,
...
-        if 'content-type' in self.headers:
-            content_type = self.headers['content-type'].lower()
-            match = re.search(r'charset=(\S+)', content_type)
-            if match:
-                encoding = match.group(1)
-        if encoding is None:
-            encoding = 'utf-8'  # assume utf-8
-
-        return encoding
+        return self.req.iter_lines(decode_unicode=True)
```
twaugh: This returns `bytes` not `str`, because the headers returned in the response include:
```
Transfer-Encoding: chunked, chunked
Content-Type: application/json
```

i.e. the charset is not specified. This leads to an exception when using "build-logs -f".

RFC 4627 defines a method for determining the encoding of an application/json response by examining the first four bytes, and this is implemented in `requests.utils.guess_json_utf()`.

It seems that it is not exactly http.py's job to do this decoding as part of the request object, but I think it would be useful for it to provide a function to do it, like this:
```
def decoded_json(iterable):
    """Return str objects for each JSON string in iterable

    See https://tools.ietf.org/html/rfc4627#section-3
    """
    for line in iterable:
        yield line.decode(requests.utils.guess_json_utf(line)
```
(needs a test case too)

and then core.py should use this function whenever it uses `iter_lines()` expecting JSON strings, like this:
```
for line in decoded_json(response.iter_lines()):
```

tests/test_core.py
```diff
@@ -92,7 +90,6 @@ def test_stream_logs(self, openshift):
         (flexmock(openshift)
             .should_receive('_get')
              # First: timeout in response after 100s
-            .and_raise(ex)
```
twaugh: I just spotted this - any way to restore the "timeout" case to this test?

twaugh: Turns out we can: https://github.com/projectatomic/osbs-client/commit/3323f30fb63c1b027f9327e799064ea7aeba69fe, and the test fails.

twaugh:
While we're at it, can we review all the tests that are removed in this PR and check they really need to be removed instead of being adapted?

osbs/core.py
```diff
@@ -405,20 +404,15 @@ def stream_logs(self, build_id):
         while True:
             buildlogs_url = self._build_url("builds/%s/log/" % build_id,
                                             **kwargs)
+            response = self._get(buildlogs_url, stream=1,
```
twaugh: Why is this outside the try:...except: block?

twaugh:
This looks ready to (squash commits and) merge, just don't want to destabilise things for the upcoming release. Maybe we should merge after that.

vrutkovs:
Agree, marking it as '[Merge after release]'

twaugh:
@puiterwijk would be great if you could squash these into a single commit (or however many make sense).

puiterwijk:
@twaugh Sure, I'm fine with doing that. Do you want me to retain the commits from you and @vrutkovs separate, or do you want me to squash those too?

twaugh:
I'm happy for mine to be squashed.

puiterwijk:
This PR is now squashed to a single PR. If @vrutkovs prefers not to get his commits squashed as well, I can pull those back as their own commits.

vrutkovs:
Integration tests failed on a rebased version of this PR: for some reason `wait_for_build_to_get_scheduled(build_id)` returns when the build is complete, not scheduled.

I'll investigate this further

vrutkovs:
Turns out that we can't use this version in our brew builders - RHEL6 has outdated `python-requests-2.6.0-3.el6.noarch`, which doesn't seem to watch resources properly. An update to 2.13 via pip fixes it.

vrutkovs:
Not sure about a particular bug in python-requests, but is seems `iter_lines` is hanging with `stream=True`. I've reworked this function in https://github.com/vrutkovs/osbs-client/commit/b900def9c4b0e607437704669e35c05540baf861 and it worked! The code is probably too ugly to include

@puiterwijk @twaugh would you mind reviewing it so we could merge this PR?

puiterwijk:
@vrutkovs I think that this skips the unicode decoding, but other than that, it looks OK to me. (but as you said, not pretty).

I'm wondering if we can detect broken requests versions, and only use this path when those are detected?

vrutkovs:
@puiterwijk yes, I think its feasible, I'll take a look onto that

puiterwijk:
Also, my personal opinion is that if python-requests in RHEL6 is broken, we should file a bug and get it fixed.
Since this is part of the standard, documented API.

vrutkovs:
@puiterwijk agree, although the oldest working version is 2.11.1 - so the diff between 2.6.0 and 2.11.1 would be pretty large.

I'll check which is easier - conditionally include a function in osbs-client or make a patch for rhel6

vrutkovs:
Final version: https://github.com/vrutkovs/osbs-client/commit/52e777de9fcc873c1151b7871cecda369d4d7a1b

I don't think backporting this part is feasible for rhel6 - its much simpler to update the version all together. In any case rhel6 package update is not going to happen soon

vrutkovs:
After a more careful investigation we've found out that no python-requests update is required, just a small fix to set default `chunk_size` to 1 is required to make it work.

Thanks and sorry it took so long to merge this

Issue #393, opened 2016-04-17
CLI: login


twaugh:
There should be a `login` verb which saves the command-line-provided token to ~/.osbs/{instance}.token.

The default for `token_file` should be that location if it exists.

PR #464, opened 2016-10-05
Add login subcommand

lcarva:
Resolves: #393

Early preview, needs unit testing.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -10,6 +10,8 @@ from __future__ import print_function, unicode_literals, absolute_import
 import json
 import logging
 import os
+import os.path
+import stat
 import sys
 import warnings
 import datetime
@@ -539,6 +541,27 @@ class OSBS(object):
     @osbsapi
     def get_token(self):
         return self.os.get_oauth_token()
+
+    @osbsapi
+    def login(self, token):
+        token_file = utils.get_instance_token_file_name(self.os_conf.conf_section)
+        token_file_dir = os.path.dirname(token_file)
+
+        if not os.path.exists(token_file_dir):
+            os.makedirs(token_file_dir)
+
+        # Inspired by http://stackoverflow.com/a/15015748/5998718
+        # For security, remove file with potentially elevated mode
+        if os.path.exists(token_file):
+            os.remove(token_file)
+
+        # Open file descriptor
+        fdesc = os.open(token_file,
+                        os.O_WRONLY | os.O_CREAT | os.O_EXCL,
+                        stat.S_IRUSR | stat.S_IWUSR)
+
+        with os.fdopen(fdesc, 'w') as f:
+            f.write(token)
 
     @osbsapi
     def get_user(self, username="~"):
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -335,6 +335,11 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
     return repo + separator + branch
 
 
+def get_instance_token_file_name(instance):
+    """Return the token file name for the given instance."""
+    return '{0}/.osbs/{1}.token'.format(os.path.expanduser('~'), instance)
+
+
 def run_command(*popenargs, **kwargs):
     """
     Run command with arguments and return its output as a byte string.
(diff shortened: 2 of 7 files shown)


twaugh:
Yes, that looks great.

lcarva:
Unit tests added.

osbs/api.py
```diff
@@ -529,6 +530,15 @@ def get_token(self):
         return self.os.get_oauth_token()
 
     @osbsapi
+    def login(self, token):
+        token_file = utils.get_instance_token_file_name(self.os_conf.conf_section)
+        token_file_dir = os.path.dirname(token_file)
+        if not os.path.exists(token_file_dir):
+            os.makedirs(token_file_dir)
+        with open(token_file, 'w') as f:
+            f.write(token)
```
twaugh: We should take some precautions with the permissions of the file or its directory.

osbs/utils.py
```diff
@@ -335,6 +335,10 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
     return repo + separator + branch
 
 
+def get_instance_token_file_name(instance):
+    """Return the token file name for the given instance."""
+    return '{0}/.osbs/{1}.token'.format(os.path.expanduser('~'), instance)
+
```
twaugh: Needs additional blank line for PEP8?

osbs/api.py
lcarva: This resolves to a sub path of the user's home directory, which should be writable by the user.
What precautions would you recommend?

osbs/utils.py
lcarva: Added.

osbs/api.py
lcarva: I see what you mean.
Modified to set read/write permission only for file owner on token file.

lcarva:
PTAL @twaugh

osbs/api.py
twaugh: We can't create the file and set its permissions afterwards; we have to do it in one step (otherwise someone could open the file in the interim and hold onto the file descriptor).

I think you can only do this in Python with `os.open`.

lcarva: Fixed by using `os.fchmod` and passing the file descriptor from file object within context.

osbs/api.py
```diff
@@ -541,6 +542,16 @@ def get_token(self):
         return self.os.get_oauth_token()
 
     @osbsapi
+    def login(self, token):
+        token_file = utils.get_instance_token_file_name(self.os_conf.conf_section)
+        token_file_dir = os.path.dirname(token_file)
+        if not os.path.exists(token_file_dir):
+            os.makedirs(token_file_dir)
+        with open(token_file, 'w') as f:
+            os.fchmod(f.fileno(), 0o600)
```
twaugh: This is still creating the file first and then changing its permissions.

twaugh: See e.g. http://stackoverflow.com/questions/5624359/write-file-with-specific-permissions-in-python

osbs/api.py
```diff
@@ -541,6 +543,31 @@ def get_token(self):
...
 
     @osbsapi
+    def login(self, token):
+        token_file = utils.get_instance_token_file_name(self.os_conf.conf_section)
+        token_file_dir = os.path.dirname(token_file)
+
+        if not os.path.exists(token_file_dir):
+            os.makedirs(token_file_dir)
+
+        # Shameless copy from http://stackoverflow.com/a/15015748/5998718
```
twaugh: Do we need to include a copy of the MIT license in that case?

osbs/api.py
```diff
@@ -541,6 +543,31 @@ def get_token(self):
...
+        # For security, remove file with potentially elevated mode
+        if os.path.exists(token_file):
+            os.remove(token_file)
+
+        # Open file descriptor
+        umask_original = os.umask(0)
```
twaugh: I'm actually not so sure we need to reset the umask. If the user wants to set their umask to something so restrictive they can't write to their own files, we should probably honour that.

lcarva: Removed umask setting.

PR #465, opened 2016-10-06
Make --quiet flag work in CLI

csomh:
Output will be quite (WARNING level) if -q, --quite option is present.

Verbose output (DEBUG level) will be produced if
    --verbose option present
    OR
    verbose is set to true (1) in <config>/general.

Otherwise output has INFO level.

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -466,6 +466,8 @@ def cli():
         description="OpenShift Build Service client"
     )
     exclusive_group = parser.add_mutually_exclusive_group()
+    # FIXME: default=None is needed to indicate for osbs.conf.Configuration
+    # that the option was not specified
     exclusive_group.add_argument("--verbose", action="store_true", default=None)
     exclusive_group.add_argument("-q", "--quiet", action="store_true")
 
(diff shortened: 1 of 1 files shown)


osbs/cli/main.py
```diff
@@ -684,13 +684,13 @@ def main():
         logger.error("Configuration error: %s", ex.message)
         return -1
 
-    is_verbose = os_conf.get_verbosity()
+    is_verbose = os_conf.get_verbosity() or args.verbose
```
vrutkovs: This would disable `--verbose` option

csomh: `args.verbose` is already handled by osbs.conf.Configuration (not very obvious), so I removed that, and limited the change to prioritize the `--quiet` flag correctly.

osbs/cli/main.py
```diff
@@ -466,6 +466,8 @@ def cli():
         description="OpenShift Build Service client"
     )
     exclusive_group = parser.add_mutually_exclusive_group()
+    # FIXME: default=None is needed to indicate for osbs.conf.Configuration
```
lcarva: This seems reasonable to me. @twaugh ?

twaugh: And to me.

PR #466, opened 2016-10-11
Allow build_imagestream in conjuction with scratch=true

twaugh:
When creating a Build directly OpenShift does not convert ImageStreamTag references to DockerImage references. In this circumstance, osbs-client needs to do this itself.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -268,6 +268,18 @@ class OSBS(object):
             build_json['spec'] = {}
         build_json['spec']['serviceAccount'] = 'builder'
         build_json['metadata']['labels']['scratch'] = 'true'
+
+        builder_img = build_json['spec']['strategy']['customStrategy']['from']
+        kind = builder_img['kind']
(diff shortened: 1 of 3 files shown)


osbs/api.py
```diff
@@ -268,6 +268,18 @@ def _create_scratch_build(self, build_request):
...
+
+        builder_img = build_json['spec']['strategy']['customStrategy']['from']
+        kind = builder_img['kind']
+        if kind == 'ImageStreamTag':
+            # Only BuildConfigs get to specify an ImageStreamTag. When
+            # creating Builds directly we need to specify a
+            # DockerImage.
+            response = self.get_image_stream_tag(builder_img['name'])
+            ref = response.json()['image']['dockerImageReference']
+            builder_img['kind'] = 'DockerImage'
```
lcarva: I'm wondering if we should just delete `builder_img['kind']` since that attribute is not set when using an image reference.

PR #470, opened 2016-10-24
configure pulp_pull plugin

twaugh:
Only run this when Pulp is used and we are producing v2 images.

Depends on https://github.com/projectatomic/atomic-reactor/pull/574.

### DIFF
--- a/inputs/prod_inner.json
+++ b/inputs/prod_inner.json
@@ -124,6 +124,10 @@
       }
     },
     {
+      "name": "pulp_pull",
+      "args": {}
+    },
+    {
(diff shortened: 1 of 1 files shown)


twaugh:
Although there is technically no need to load the squashed image back in if the pulp_pull plugin is used, doing so will avoid having to transfer the blobs back in from the registry which may be an expensive network operation.

lcarva:
Atomic reactor's PR has been merged, if this is now ready for merge, please update PR's title.

twaugh:
It's ready.

PR #473, opened 2016-10-31
Remove support for OpenShift v1.0.5 and below

lcarva:


twaugh:
How about docs/secret.md?

lcarva:
Updated docs/secret.md

twaugh:
How about `tests/mock_jsons/0.5.4`?

lcarva:
I think ideally, we want to regenerate those jsons (0.5.4 and 1.0.4) for newer versions (1.0.6 and 1.3.3).
I went down this road by running `mock_jsons.sh` which had to be modified slightly, but it only produces some of the files needed.
I can address those, but it'll make this patch, and story, much larger. Wasn't sure if we wanted to focus mainly on doc and code updates for now.

twaugh:
Fair enough.

PR #474, opened 2016-11-04
docs: update development setup

TomasTomecek:
Just let me know if the proposed changes are too drastic.

Also am not sure if the dockerfile should also be placed to a-r.

Signed-off-by: Tomas Tomecek <ttomecek@redhat.com>


docs/development-setup.md
```diff
@@ -1,73 +1,137 @@
...
+$ docker build --no-cache --tag=buildroot .
 ```
 
-Then you can start OpenShift with one simple command:
+If you are using `osbs-client` from git master branch, you should also install
+git master of atomic-reactor:
 
 ```
-$ docker-compose up -d
+$ docker build --no-cache --tag=buildroot --build-arg LATEST_CODE=true .
```
csomh: Shouldn't this have `REACTOR_SOURCE=git` instead of `LATEST_CODE=true`?

For those who are not masters of docker yet, would be nice to include who this step can be verified (`docker images` and look for *buildroot*) - error messages are not always obvious and can be easily missed.

docs/development-setup.md
```diff
@@ -1,73 +1,137 @@
...
 ```
 
 
-## setup without docker-compose
+## Building images, finally
+
+In order to submit a build, you need to have a permission. If you started
+OpenShift with `oc cluster up`, there's a user `developer` set up with
+namespace `myproject`, this is what we'll be using.
```
csomh: I know this might be considered trivial, but the step to `oc login` mentioned explicitly here would be nice.

docs/development-setup.md
```diff
@@ -1,73 +1,137 @@
...
-    registry_uri = <registry:5000>
-    vendor = <this_company>
-    authoritative_registry = <registry.example.com>
-    build_host = <builder.example.com>
-    architecture = x86_64
-    verify_ssl = false
-    builder_use_auth = false
-    use_auth=false
-    ```
+#### Registry
```
csomh: Shouldn't this be moved after the getting the token step? The registry IP should be used in `osbs.conf`.

docs/development-setup.md
```diff
@@ -1,73 +1,137 @@
...
+
+`oc cluster up` method deploys registry by default. You can figure out the IP
+address of the registry by issuing following command:
+
+```
+$ oc describe service docker-registry
```
csomh: I could not make this work, and couldn't figure it out what is missing.
I've did `oc login -u developer`.
`oc get svc` gives me nothing and `oc status` gives me this:
```
In project My Project (myproject) on server https://192.168.122.37:8443

You have no services, deployment configs, or build configs.
Run 'oc new-app' to create an application.
```

docs/development-setup.md
TomasTomecek: you're correct (and I thought I fixed all of those); I changed the variable name and semantics halfway through and apparently forgot to update all past references

docs/development-setup.md
TomasTomecek: Good catch.

docs/development-setup.md
TomasTomecek: To me, that's kind of redundant, but I certainly don't mind adding such info.

docs/development-setup.md
TomasTomecek: You're absolutely correct. I've just realized that this is wrong (I deployed another registry by accident).

TomasTomecek:
Thanks for review. Great suggestions!

docs/development-setup.md
```diff
@@ -1,73 +1,158 @@
...
+    rules bufore starting the cluster with `iptables -F`)
+
+
+## Getting `osbs-client`
+
+For local development, we advise to clone this git repo:
 
 ```
 $ git clone git@github.com:projectatomic/osbs-client.git
 $ cd osbs-client
```
lcarva: Maybe add `python setup.py develop` for functional parity with `dnf` approach below?

TomasTomecek: Great suggestion! I actually have my own aliases for this problem. That's why I forgot about the CLI when just cloning the repo.

docs/development-setup.md
```diff
@@ -28,6 +28,9 @@ For local development, we advise to clone this git repo:
...
+$ pip3 install -r ./requirements.txt
+$ pip3 install -r ./requirements-py3.txt
+$ python3 ./setup.py develop
```
lcarva: Shouldn't this line take care of the requirements?

TomasTomecek:
Is any of the issues still unresolved? If not, can we have this merged?

TomasTomecek:
I think I resolved most of the issues. Only one remains: pushing images.

Build container needs to use `builder` service account for authentication. This can be done via mounting dockercfg secret in the build container (when present in home, no changes are needed in a-r).

I'm assuming this could be done by adding new configuration option to bind the secret in the pod. Preferably to `~/.docker/config.json`.

Honestly, I won't have time to implement that. If you have time to do that, great! If not, I remove the registry part from the docs.

twaugh:
Doesn't the `builder` service account get used by default for builds? I don't think we have to do anything special for that.

TomasTomecek:
This is about actual image push. In order to do that successfully, you need to authenticate with the registry. OpenShift is able then authorize your request, if you authenticate with correct credentials. Hence you need the credentials in build container (maybe they are mounted there by default, haven't checked this).

We would have to add auth bits [here](https://github.com/projectatomic/atomic-reactor/blob/13d0e3e867752c1715e4360a6aab6cd5d12dd965/atomic_reactor/core.py#L481).

KarstenHopp:
Two observations from reading development-setup.md:
'Obtaining token' should probably go to the 'config file' section with an explanation that the placeholder token in your example needs to be replaced with it.

The 'registry' section is in the wrong order imho. First there is the 'oc describe service docker-registry' command, which will fail as you're logged in as developer after installation.
So the correct order would be 
 oc login -u system:admin
 oc project default
 oc describe service docker-registry
and then (missing)
 oc login -u developer

docs/development-setup.md
```diff
@@ -1,73 +1,164 @@
...
+
+ 1. If you're on Fedora, you can install package origin:
+    ```
+    $ dnf install -y origin
+    ```
+    Now you have command `oc` available on your system
+
+ 2. You can start a single-node cluster like this
+    ```
+    $ oc cluster up
```
lcarva: In addition, the following is required to allow builder ServiceAccount to perform operations from within builder container:
`oc adm policy add-role-to-user admin system:serviceaccount:myproject:builder`

For instance, without this, the plugin `store_metadata_in_osv3` fails with 403 error:  `User "system:serviceaccount:myproject:builder" cannot get builds in project "myproject"`,

twaugh: Does the 'edit' role work for this or is 'admin' required?

TomasTomecek:
Hey guys, I would love to update the PR this week, so hopefully we can merge before Christmas or in early January.

docs/development-setup.md
TomasTomecek: Luiz, you're correct! And Tim as well, `edit` is sufficient.

TomasTomecek:
This should be ready for review now. I think I addressed all the comments. Unfortunately I had to "cut" the registry part b/c it just didn't work and would require some non-trivial code changes.

Thank you to everyone who helped with reviews: @KarstenHopp @lcarva @twaugh @csomh

TomasTomecek:
Luiz, thanks for your thorough review!

I agree with all your points, I bumped into 3) 4) (and probably 2) but haven't noticed it).

lcarva:
@TomasTomecek tried fixing conflict in README.md file, but github still complains. Could you rebase so we can merge this?

PR #475, opened 2016-11-10
first pass at enabling site-customizations

maxamillion:

### DIFF
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -110,3 +110,53 @@ Some options are also mandatory.
 * `memory_limit` (*optional*, `string`) — memory limit to apply to build (for more info, see [documentation for resources](https://github.com/projectatomic/osbs-client/blob/master/docs/resource.md)
 
 * `storage_limit` (*optional*, `string`) — storage limit to apply to build (for more info, see [documentation for resources](https://github.com/projectatomic/osbs-client/blob/master/docs/resource.md)
+
+
+## Build JSON Templates
+
+In the `build_json_dir` there must be `prod.json` and `prod_inner.json` which
+defines the [OpenShift Build](https://docs.openshift.org/latest/dev_guide/builds.html)
+specification that will be used to enable the specific
+[atomic-reactor](https://github.com/projectatomic/atomic-reactor) plugins that
+should be enabled and the config values for each.
+
+There is also a third file that is optional that can exist along side the
+previous two in `build_json_dir` which is `prod_customize.json` and it will
+provide the ability to set site-specific customizations such as removing,
+plugins, adding plugins, or modifying arguments passed to existing plugins.
+
+The syntax of `prod_customize.json` is as follows:
+
+```json
+{
+    "disable_plugins": [
+        {
+            "plugin_type": "foo_type",
+            "plugin_name": "foo_name"
+        },
+        {
+            "plugin_type": "bar_type",
+            "plugin_name": "bar_name"
+        }
+    ],
+
+    "enable_plugins": [
+        {
+            "plugin_type": "foo_type",
+            "plugin_name": "foo_name",
+            "plugin_args": {
+                "foo_arg1": "foo_value1",
+                "foo_arg2": "foo_value2"
+            }
+        }
+    ]
+}
+```
+
+Such that:
+
+* `disable_plugins` will define a list of lists that define the plugin type of the plugin that is to be removed (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`) and the name of the plugin.
+
+* `enable_plugins` is used to add plugins or modify already enabled plugins, these must be defined as key-value pairs as illustrated above.
+
+
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -123,7 +123,7 @@ should be enabled and the config values for each.
 There is also a third file that is optional that can exist along side the
 previous two in `build_json_dir` which is `prod_customize.json` and it will
 provide the ability to set site-specific customizations such as removing,
-plugins, adding plugins, or modifying arguments passed to existing plugins.
+plugins, adding plugins, or overriding arguments passed to existing plugins.
 
 The syntax of `prod_customize.json` is as follows:
 
@@ -157,6 +157,6 @@ Such that:
 
 * `disable_plugins` will define a list of lists that define the plugin type of the plugin that is to be removed (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`) and the name of the plugin.
 
-* `enable_plugins` is used to add plugins or modify already enabled plugins, these must be defined as key-value pairs as illustrated above.
+* `enable_plugins` is used to add plugins or modify already enabled plugins by overriding args passed to the plugin, these must be defined as key-value pairs as illustrated above. It should be noted that plugins added here will be executed at the end of the list of plugins in that particular `plugin_type` (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`).
 
 
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -155,8 +155,8 @@ The syntax of `prod_customize.json` is as follows:
 
 Such that:
 
-* `disable_plugins` will define a list of lists that define the plugin type of the plugin that is to be removed (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`) and the name of the plugin.
+* `disable_plugins` will define a list of dicts that define the plugin type of the plugin that is to be removed (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`) and the name of the plugin.
 
-* `enable_plugins` is used to add plugins or modify already enabled plugins by overriding args passed to the plugin, these must be defined as key-value pairs as illustrated above. It should be noted that plugins added here will be executed at the end of the list of plugins in that particular `plugin_type` (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`), unless the plugin has already been previously added and this setting is only being used to override args. In the case of arg override, the plugin order execution will not change.
+* `enable_plugins` will define a list of dicts that is used to add plugins or modify already enabled plugins by overriding args passed to the plugin, these must be defined as key-value pairs as illustrated above. It should be noted that plugins added here will be executed at the end of the list of plugins in that particular `plugin_type` (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`), unless the plugin has already been previously added and this setting is only being used to override args. In the case of arg override, the plugin order execution will not change.
 
 
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -157,6 +157,6 @@ Such that:
 
 * `disable_plugins` will define a list of lists that define the plugin type of the plugin that is to be removed (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`) and the name of the plugin.
 
-* `enable_plugins` is used to add plugins or modify already enabled plugins by overriding args passed to the plugin, these must be defined as key-value pairs as illustrated above. It should be noted that plugins added here will be executed at the end of the list of plugins in that particular `plugin_type` (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`).
+* `enable_plugins` is used to add plugins or modify already enabled plugins by overriding args passed to the plugin, these must be defined as key-value pairs as illustrated above. It should be noted that plugins added here will be executed at the end of the list of plugins in that particular `plugin_type` (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`), unless the plugin has already been previously added and this setting is only being used to override args. In the case of arg override, the plugin order execution will not change.
 
 
--- a/inputs/prod_customize.json
+++ b/inputs/prod_customize.json
@@ -1,0 +1,7 @@
+{
+    "disable_plugins": [
+    ],
+
+    "enable_plugins": [
+    ]
+}
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -680,7 +680,7 @@ class BuildRequest(object):
                         plugin_dict['plugin_name']
                     )
                     logger.debug(
-                        "site-specific plugin disabled -> Type:{} Name:{}".format(
+                        "site-specific plugin disabled -> Type:{0} Name:{1}".format(
                             plugin_dict['plugin_type'],
                             plugin_dict['plugin_name']
                         )
@@ -701,7 +701,7 @@ class BuildRequest(object):
                         plugin_dict['plugin_args']
                     )
                     logger.debug(
-                        "site-specific plugin enabled -> Type:{} Name:{} Args:".format(
+                        "site-specific plugin enabled -> Type:{0} Name:{1} Args: {2}".format(
                             plugin_dict['plugin_type'],
                             plugin_dict['plugin_name'],
                             plugin_dict['plugin_args']
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -24,7 +24,7 @@ except ImportError:
 
 from osbs.build.manipulate import DockJsonManipulator
 from osbs.build.spec import BuildSpec
-from osbs.constants import SECRETS_PATH, DEFAULT_OUTER_TEMPLATE, DEFAULT_INNER_TEMPLATE
+from osbs.constants import SECRETS_PATH, DEFAULT_OUTER_TEMPLATE, DEFAULT_INNER_TEMPLATE, DEFAULT_CUSTOMIZE_CONF
 from osbs.exceptions import OsbsException, OsbsValidationException
 from osbs.utils import looks_like_git_hash, git_repo_humanish_part_from_uri
 
@@ -46,6 +46,7 @@ class BuildRequest(object):
         self.build_json = None       # rendered template
         self._template = None        # template loaded from filesystem
         self._inner_template = None  # dock json
+        self._customize_conf = None  # site customize conf for _inner_template
         self._dj = None
         self._resource_limits = None
         self._openshift_required_version = parse_version('1.0.6')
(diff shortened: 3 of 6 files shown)


osbs/build/build_request.py
```diff
@@ -136,6 +137,20 @@ def inner_template(self):
...
     @property
+    def customize_conf(self):
+        if self._customize_conf is None:
+            path = os.path.join(self.build_json_store, DEFAULT_CUSTOMIZE_CONF)
+            logger.debug("loading inner template from path %s", path)
+            try:
+                with open(path, "r") as fp:
+                    self._customize_conf= json.load(fp)
+            except IOError:
+                # File not found, which is perfectly fine. Just leave it as None
```
lcarva: If the file is not found, it'll try to read it from disk every time this property is accessed. Maybe set it to `{}` ?

osbs/build/build_request.py
```diff
@@ -791,6 +832,9 @@ def render(self, validate=True):
         self.render_koji_promote(use_auth=use_auth)
         self.render_sendmail()
 
+
+        self.render_customizations()
```
lcarva: Shouldn't this be done at the beginning of `render`? Disabling and enabling a certain plugin in some cases requires other modifications.

osbs/build/build_request.py
```diff
@@ -648,6 +663,32 @@ def render_import_image(self, use_auth=None):
...
                                           'insecure_registry', True)
 
+    def render_customizations(self):
+        """
+        Customize prod_inner for site specific customizations
+        """
+
+        try:
+            for plugin_tuple in self.customize_conf['disable_plugins']:
+                self.dj.remove_plugin(*plugin_tuple)
```
lcarva: Maybe check if plugin is actually enabled first?

osbs/build/build_request.py
```diff
@@ -648,6 +663,32 @@ def render_import_image(self, use_auth=None):
...
+        except TypeError:
+            # None provided to disable, we can ignore this as that's a valid use
+            # case
+            pass
+
+        try:
+            for plugin in self.customize_conf['enable_plugins']:
+                self.dj.add_plugin(
```
lcarva: Check if plugin is not enabled first?

osbs/build/manipulate.py
```diff
@@ -50,6 +50,14 @@ def remove_plugin(self, plugin_type, plugin_name):
                 self.dock_json[plugin_type].remove(p)
                 break
 
+    def add_plugin(self, plugin_type, plugin_name, args_dict):
+        """
+        if config contains plugin, remove it
+        """
+        self.dock_json[plugin_type].setdefault(plugin_name, dict)
+        for arg_key, arg_val in args_dict:
```
lcarva: `args_dict.items()` ?

osbs/build/build_request.py
maxamillion: Ah, good call. I'll need to change the exception handling for `render_customizations` to match as well. Thanks!

osbs/build/build_request.py
maxamillion: I wasn't sure if it should be done at the beginning or at the end because if someone sets args for a site customizations, won't some of these other `render_` function calls potentially overwrite those customizations? I could split it out to be `self.render_customizations_remove()` and `self.render_customizations_enable()` with one at the beginning and the other at the end.

osbs/build/manipulate.py
maxamillion: I don't follow, why would that be better?

lcarva:
```python
>>> args_dict = {'one': 'um', 'two': 'dois'}

>>> for arg_key, arg_val in args_dict:
...     print arg_key, arg_val
ValueError: too many values to unpack
too many values to unpack

>>> for x, y, z in args_dict:
...     print x, y, z
t w o
o n e

>>> for arg_key, arg_val in args_dict.items():
...     print arg_key, arg_val
two dois
one um
```

maxamillion: :+1:

osbs/build/build_request.py
twaugh: I think it should be done at the start, i.e. it describes changes that you could make to the prod_inner.json file.

If there are changes that a site needs to make that get overwritten by the way osbs-client transforms the plugins during rendering, we need to handle that in osbs-client either by changing its default behaviour or by making it configurable.

docs/configuration_file.md
```diff
@@ -110,3 +110,47 @@ Some options are also mandatory.
...
+plugins, adding plugins, or modifying arguments passed to existing plugins.
+
+The syntax of `prod_customize.json` is as follows:
+
+```json
+{
+    "disable_plugins": [
+        ["plugin_type1", "plugin_name1"],
+        ["plugin_type2", "plugin_name2"]
+    ],
```
csomh: Would require some additional handling in code, but wouldn't be better to make this consistent with `enable_plugins`, and have
```json
"disable_plugins": [
    {
        "plugin_type1": "foo_type",
        "plugin_name1": "foo_name"
    },
    {
        "plugin_type2": "bar_type",
        "plugin_name2": "bar_name"
    }
]
```

osbs/build/build_request.py
maxamillion: @twaugh moved to the start

docs/configuration_file.md
maxamillion: +1 Updated

maxamillion:
I cleaned this up a bit, fixed a bug and added tests cases

docs/configuration_file.md
```diff
@@ -110,3 +110,53 @@ Some options are also mandatory.
...
+                "foo_arg2": "foo_value2"
+            }
+        }
+    ]
+}
+```
+
+Such that:
+
+* `disable_plugins` will define a list of lists that define the plugin type of the plugin that is to be removed (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`) and the name of the plugin.
```
lcarva: "list of dicts"

docs/configuration_file.md
```diff
@@ -110,3 +110,53 @@ Some options are also mandatory.
...
+In the `build_json_dir` there must be `prod.json` and `prod_inner.json` which
+defines the [OpenShift Build](https://docs.openshift.org/latest/dev_guide/builds.html)
+specification that will be used to enable the specific
+[atomic-reactor](https://github.com/projectatomic/atomic-reactor) plugins that
+should be enabled and the config values for each.
+
+There is also a third file that is optional that can exist along side the
+previous two in `build_json_dir` which is `prod_customize.json` and it will
+provide the ability to set site-specific customizations such as removing,
+plugins, adding plugins, or modifying arguments passed to existing plugins.
```
lcarva: "modifying arguments" - it does this by overriding what's already there. Maybe this can be clarified here.

osbs/build/build_request.py
```diff
@@ -649,10 +664,63 @@ def render_import_image(self, use_auth=None):
...
+                        )
+                    )
+                except KeyError:
+                    # Malformed config
+                    logger.debug("Invalid custom configuration found for disable_plugins")
+        except TypeError:
+            # None provided to disable, we can ignore this as that's a valid use
+            # case
+            logger.debug("No site-specific plugins to disable")
+            pass
```
lcarva: Unnecessary `pass` here.

osbs/build/build_request.py
```diff
@@ -649,10 +664,63 @@ def render_import_image(self, use_auth=None):
                 self.dj.dock_json_set_arg('postbuild_plugins', 'import_image',
                                           'insecure_registry', True)
 
+    def render_customizations(self):
+        """
+        Customize prod_inner for site specific customizations
+        """
+
+        try:
+            for plugin_dict in self.customize_conf['disable_plugins']:
```
lcarva: It might be more "pythonic" to remove one-level of nested code by replacing the outer try/except with `self.customize_conf.get('disable_plugins', [])`

This will return an empty list of customize_conf does not provide any 'disable_plugins'.

same for the `self.customize_conf['enable_plugins']` statement below.

osbs/build/build_request.py
```diff
@@ -136,6 +137,20 @@ def inner_template(self):
...
+    def customize_conf(self):
+        if self._customize_conf is None:
+            path = os.path.join(self.build_json_store, DEFAULT_CUSTOMIZE_CONF)
+            logger.debug("loading customize conf from path %s", path)
+            try:
+                with open(path, "r") as fp:
+                    self._customize_conf= json.load(fp)
+            except IOError:
+                # File not found, which is perfectly fine. Set to empty string
+                self._customize_conf = ""
```
lcarva: Can this default to an empty `dict` instead? It should only ever be `None` or a `dict`.

docs/configuration_file.md
```diff
@@ -110,3 +110,53 @@ Some options are also mandatory.
...
+
+## Build JSON Templates
+
+In the `build_json_dir` there must be `prod.json` and `prod_inner.json` which
+defines the [OpenShift Build](https://docs.openshift.org/latest/dev_guide/builds.html)
+specification that will be used to enable the specific
+[atomic-reactor](https://github.com/projectatomic/atomic-reactor) plugins that
+should be enabled and the config values for each.
+
+There is also a third file that is optional that can exist along side the
```
lcarva: Please also clarify that added plugins will be executed last for the corresponding plugin type.

osbs/build/manipulate.py
```diff
@@ -50,6 +50,17 @@ def remove_plugin(self, plugin_type, plugin_name):
                 self.dock_json[plugin_type].remove(p)
                 break
 
+    def add_plugin(self, plugin_type, plugin_name, args_dict):
+        """
+        if config has plugin, override it, else add it
```
lcarva: In the case of modifying a plugin, I think it'd make sense to replace it in place, instead of appending to end of list.

lcarva:
btw unit tests failed on python 2.6 due to unsupported string formatting syntax.

docs/configuration_file.md
maxamillion: @lcarva +1

docs/configuration_file.md
maxamillion: @lcarva +1 - oversight.

osbs/build/build_request.py
maxamillion: +1

osbs/build/build_request.py
maxamillion: @lcarva +1

osbs/build/manipulate.py
maxamillion: +1

inputs/prod_customize.json
```diff
@@ -0,0 +1,7 @@
+{
+    "disable_plugins": [
+    ],
+
+    "enable_plugins": [
+    ]
+}
```
twaugh: To be clear: this is an example customize file and will not be shipped in the package/RPM, is that correct? That way makes sense to me.

osbs/build/build_request.py
```diff
@@ -649,10 +664,60 @@ def render_import_image(self, use_auth=None):
                 self.dj.dock_json_set_arg('postbuild_plugins', 'import_image',
                                           'insecure_registry', True)
 
+    def render_customizations(self):
+        """
+        Customize prod_inner for site specific customizations
+        """
+
+        disable_plugins = self.customize_conf.get('disable_plugins', [])
+        if len(disable_plugins) == 0:
```
twaugh: Maybe this could be `if not disable_plugins:`?

osbs/build/build_request.py
```diff
@@ -649,10 +664,60 @@ def render_import_image(self, use_auth=None):
...
+                    )
+                    logger.debug(
+                        "site-specific plugin disabled -> Type:{0} Name:{1}".format(
+                            plugin_dict['plugin_type'],
+                            plugin_dict['plugin_name']
+                        )
+                    )
+                except KeyError:
+                    # Malformed config
+                    logger.debug("Invalid custom configuration found for disable_plugins")
```
twaugh: Does this want to include the `plugin_type` and `plugin_name` it is complaining about?

osbs/build/build_request.py
```diff
@@ -649,10 +664,60 @@ def render_import_image(self, use_auth=None):
...
+
+        enable_plugins = self.customize_conf.get('enable_plugins', [])
+        if len(enable_plugins) == 0:
```
twaugh: `if not enable_plugins:`?

osbs/build/build_request.py
```diff
@@ -649,10 +664,60 @@ def render_import_image(self, use_auth=None):
...
+                    logger.debug(
+                        "site-specific plugin enabled -> Type:{0} Name:{1} Args: {2}".format(
+                            plugin_dict['plugin_type'],
+                            plugin_dict['plugin_name'],
+                            plugin_dict['plugin_args']
+                        )
+                    )
+                except KeyError:
+                    # Malformed config
+                    logger.debug("Invalid custom configuration found for enable_plugins")
```
twaugh: Could include the `plugin_type` and `plugin_name` it didn't find.

tests/build/test_build_request.py
```diff
@@ -1348,3 +1348,85 @@ def test_prod_non_custom_base_image(self, tmpdir):
         pull_base_image_plugin = get_plugin(
             plugins, 'prebuild_plugins', 'pull_base_image')
         assert pull_base_image_plugin is not None
+
+    def test_render_prod_custom_site_plugin_enable(self):
+        """
+        Test to make sure that when we attempt to enable a plugin, it is
+        actually enabled in the JSON for the build_request post-rendering
```
twaugh: 'post-rendering' here seems like we're expecting something different than what is implemented.

tests/build/test_build_request.py
```diff
@@ -1348,3 +1348,85 @@ def test_prod_non_custom_base_image(self, tmpdir):
...
+
+        assert {
+                "name": plugin_name,
+                "args": plugin_args
+        } in build_request.dj.dock_json[plugin_type]
+
+    def test_render_prod_custom_site_plugin_disable(self):
+        """
+        Test to make sure that when we attempt to disable a plugin, it is
+        actually disabled in the JSON for the build_request post-rendering
```
twaugh: 'post-rendering' seems like it expects the custom rendering to happen at the end instead of at the start.

tests/build/test_build_request.py
```diff
@@ -1348,3 +1348,85 @@ def test_prod_non_custom_base_image(self, tmpdir):
...
+            }
+        )
+        kwargs = get_sample_prod_params()
+        build_request.set_params(**kwargs)
+        build_request.render()
+
+        assert {
+                "name": plugin_name,
+                "args": plugin_args
+        } in build_request.dj.dock_json[plugin_type]
```
twaugh: Worth testing that it stays in the same order rather than getting moved to the end?

tests/build/test_build_request.py
```diff
@@ -1348,3 +1348,85 @@ def test_prod_non_custom_base_image(self, tmpdir):
...
+
+
+
```
twaugh: Trailing space.

inputs/prod_customize.json
maxamillion: It is, but it could easily be shipped in the package and since the lists are empty it would effectively not alter anything by default. But if it is preferred not to ship it, then that's fine too.

osbs/build/build_request.py
maxamillion: +1

osbs/build/build_request.py
maxamillion: I figured if it's KeyError'ing then something is missing or wrong and we don't want to attempt to then reference the thing that caused the Exception in the output. I'm open to suggestions on how to do that, I could potentially just dump `plugin_dict`.

osbs/build/build_request.py
maxamillion: +1

osbs/build/build_request.py
maxamillion: Same concern as above about the Exception.

tests/build/test_build_request.py
maxamillion: I meant that as literally "after running the render() method on build_request, it should be there" I can change the verbage if that's desired.

tests/build/test_build_request.py
maxamillion: Same as above

tests/build/test_build_request.py
maxamillion: I think that makes sense, I'll add it.

tests/build/test_build_request.py
maxamillion: +1

inputs/prod_customize.json
twaugh: Just don't want it overwriting a prod_customize.json file that's been carefully altered.

osbs/build/build_request.py
twaugh: Oh, good point.

Let's leave it as it is.

osbs/build/build_request.py
twaugh: Agreed.

tests/build/test_build_request.py
twaugh: OK, that's fair.

tests/build/test_build_request.py
twaugh: That's fine.

inputs/prod_customize.json
maxamillion: +1

maxamillion: Ultimately we could mark the prod_customize.json file as %config(noreplace) in the rpm, but then `/usr/share/` would not be an appropriate place for it.

docs/configuration_file.md
twaugh: Think this still needs changing?

docs/configuration_file.md
twaugh: Did this get fixed?

docs/configuration_file.md
twaugh: Oh, I see where it was fixed now.

maxamillion:
Is this waiting on me for anything? (I *think* I addressed everything that was asked of me but if there's further review or something I missed then I'm happy to work through those items.)

docs/configuration_file.md
```diff
@@ -110,3 +110,53 @@ Some options are also mandatory.
...
+        }
+    ]
+}
+```
+
+Such that:
+
+* `disable_plugins` will define a list of lists that define the plugin type of the plugin that is to be removed (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`) and the name of the plugin.
+
+* `enable_plugins` is used to add plugins or modify already enabled plugins by overriding args passed to the plugin, these must be defined as key-value pairs as illustrated above. It should be noted that plugins added here will be executed at the end of the list of plugins in that particular `plugin_type` (`prebuild_plugins`, `prepublish_plugins`, `postbuild_plugins`, `exit_plugins`).
```
lcarva: executed at the end, unless the plugin has already been previously added. In such case, the plugin's arguments are fully overridden and plugin order is preserved.

maxamillion: 👍

twaugh:
Think it's just the 'list of lists' vs 'list of dicts' in the docs now.
https://github.com/projectatomic/osbs-client/pull/475/files#diff-137d027cc88931359b8498be63ec2d6cR158

lcarva:
I'm merging this as Tim's only concern has been addressed.

PR #481, opened 2016-11-29
Revert "configure pulp_pull plugin"

vrutkovs:
This reverts commit 6d9bf3c79d103e6fbc61cd37d65de1ca02c8bce8.

Pub doesn't support associating images via v2 ID, so we have to temporary revert it


vrutkovs:
`pulp_pull` needs a change and shouldn't be disabled

PR #482, opened 2016-11-30
add krb5 auth params for exit_koji_promote (auth to koji with krb5)

maxamillion:


osbs/build/build_request.py
```diff
@@ -486,7 +486,13 @@ def render_koji_promote(self, use_auth=None):
 
             if use_auth is not None:
                 self.dj.dock_json_set_arg('exit_plugins', 'koji_promote',
-                                          'use_auth', use_auth)
+                                            'use_auth', use_auth)
+
+            if self.spec.koji_use_kerberos.value:
```
lcarva: Worth enforcing that if koji_use_kerberos is set to True, principal and keytab values are also set? Alternatively, we could remove the `koji_use_kerberos` setting and only assume its behavior if `koji_kerberos_principal` and `koji_kerberos_keytab` are set.

maxamillion: @lcarva I'm good either way, the only reasdon I added `koji_use_kerberos` was there is a `use_kerberos` config flag already for the client and I was aiming for consistency. Which ever you and @twaugh (and Co.) prefer, I'm happy to go that direction.

twaugh: If kerberos is in use for both, would the principal ever be different?

maxamillion: @twaugh Yes, this koji auth is being internally used by the system for the koji content generator metadata import, the krb5 keytab is expected to be in the buildroot (either via an openshift secret or just drop it in the image). This is to configure `exit_koji_promote` while running "inner" in order to enable that for Fedora's use case because as of December 12th, ssl cert based auth will be completely disabled in Fedora's koji. (Do note that this getting merged and released by Dec 12th  is not a blocker, I can patch our package in the internal Infrastructure dnf repos if needed so there's no real urgency here).

twaugh: OK, so I prefer `koji_use_kerberos` and the `koji_*` keytab and principal parameters, and we just need to check that those parameters are set.

maxamillion: @twaugh when you say "need to check" is there something not currently being done in this pull request that you would like to see done in the code or is that just a general comment?

PR #487, opened 2016-12-09
-c is optional, and having no effect

### LINKED ISSUE #476: osbs option to be deprecated
osbs option '-c' for component name might be deprecated, as this value is anyway overwritten with value from Dockerfile (BZComponent), unless we want to allow not to specify Component name in Dockerfile and rely in that case on this option

rcerven:
I want to resolve #476 

-c is optional now, and has no effect whatsoever

instead of it, is used labels 'com.redhat.component' or 'BZComponent' from dockerfile

also made component label in dockerfile mandatory, if no found will raise exception

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -362,10 +362,10 @@ class OSBS(object):
         """
 
         def get_required_label(labels, names):
-        """
-        get value of label list, first found is returned
-        names has to be non empty tuple
-        """
+            """
+            get value of label list, first found is returned
+            names has to be non empty tuple
+            """
             assert names  # always called with a non-empty literal tuple so names[0] is safe
             for label_name in names:
                 if label_name in labels:
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -365,7 +365,7 @@ class OSBS(object):
             for label_name in names:
                 if label_name in labels:
                     return labels[label_name]
-            
+
             raise OsbsValidationException("required label '{name}' missing "
                                           "from Dockerfile"
                                           .format(name=names[-1]))
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -362,20 +362,25 @@ class OSBS(object):
         """
 
         def get_required_label(labels, names):
+        """
+        get value of label list, first found is returned
+        names has to be non empty tuple
+        """
+            assert names  # always called with a non-empty literal tuple so names[0] is safe
             for label_name in names:
                 if label_name in labels:
                     return labels[label_name]
 
             raise OsbsValidationException("required label '{name}' missing "
                                           "from Dockerfile"
-                                          .format(name=names[-1]))
+                                          .format(name=names[0]))
 
         df_parser = utils.get_df_parser(git_uri, git_ref, git_branch=git_branch)
         build_request = self.get_build_request()
         labels = df_parser.labels
 
-        name_label = get_required_label(labels, ('Name', 'name'))
-        component = get_required_label(labels, ('BZComponent', 'com.redhat.component'))
+        name_label = get_required_label(labels, ('name', 'Name'))
+        component = get_required_label(labels, ('com.redhat.component', 'BZComponent'))
 
         build_request.set_params(
             git_uri=git_uri,
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -338,8 +338,9 @@ class OSBS(object):
     @osbsapi
     def create_prod_build(self, git_uri, git_ref,
                           git_branch,  # may be None
-                          user, component,
-                          target,      # may be None
+                          user,
+                          component=None,
+                          target=None,
                           architecture=None, yum_repourls=None,
                           koji_task_id=None,
                           scratch=None,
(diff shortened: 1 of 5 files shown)


twaugh:
FWIW `--component` will have no effect rather than overriding anything. This should be documented though.

vrutkovs:
Ah, right, updated the review comment

osbs/cli/main.py
```diff
@@ -559,7 +559,7 @@ def cli():
                               help="build architecture")
     build_parser.add_argument("-u", "--user", action='store', required=True,
                               help="prefix for docker image repository")
-    build_parser.add_argument("-c", "--component", action='store', required=True,
+    build_parser.add_argument("-c", "--component", action='store', required=False,
                               help="name of component")
```
twaugh: This is where you could put help text like "not used; use com.redhat.component Dockerfile label".

tests/test_api.py
```diff
@@ -72,7 +72,7 @@ def test_get_pod_for_build(self, osbs):
 
     def test_create_build_with_deprecated_params(self, osbs):
         class MockParser(object):
-            labels = {'Name': 'fedora23/something'}
+            labels = {'Name': 'fedora23/something', 'com.redhat.component': 'fedora_component'}
```
twaugh: Can we make this use `TEST_COMPONENT`, and remove 'component' from the kwargs below?

osbs/api.py
```diff
@@ -371,6 +371,16 @@ def create_prod_build(self, git_uri, git_ref,
                                           "from Dockerfile"
                                           .format(name=name_label_name))
 
+        for component_label_name in ['BZComponent', 'com.redhat.component']:
+            if component_label_name in labels:
+                component_label = labels[component_label_name ]
+                break
+        else:
+            raise OsbsValidationException("required label '{}' missing "
```
twaugh: coveralls reports this line doesn't get run during tests.

twaugh:
`create_prod_build()`'s docstring should probably describe its `component` parameter as being able to be `None`, like e.g. `target`.

osbs/build/spec.py
```diff
@@ -165,7 +165,6 @@ def __init__(self):
             self.git_uri,
             self.git_ref,
             self.user,
-            self.component,
```
twaugh: Actually looking further at this test, I see it's whole purpose is to test whether deprecated options cause problems. :)

So I was wrong, we should include `self.component` here. In any case, we can't take it out altogether because it is a positional argument.

osbs/api.py
```diff
@@ -351,25 +352,31 @@ def create_prod_build(self, git_uri, git_ref,
...
         :return: BuildResponse instance
         """
+
+        def get_required_label(labels, names):
+            for label_name in names:
+                if label_name in labels:
+                    return labels[label_name]
+                    break
+            else:
+                raise OsbsValidationException("required label '{name}' missing "
```
twaugh: This line doesn't get coverage in tests.

twaugh:
Unit testing should verify:
* supplying 'component' keyword parameter to create_build() or create_prod_build() does not cause errors
* Dockerfile without a com.redhat.component/BZComponent label raises a validation error
* supplying '--component' CLI parameter has no effect (no 'component' keyword parameter supplied to create_* calls)

osbs/api.py
```diff
@@ -351,25 +352,30 @@ def create_prod_build(self, git_uri, git_ref,
...
-        for name_label_name in ['name', 'Name']:
-            if name_label_name in labels:
-                name_label = labels[name_label_name]
-                break
-        else:
-            raise OsbsValidationException("required label '{name}' missing "
-                                          "from Dockerfile"
-                                          .format(name=name_label_name))
+
+        name_label = get_required_label(labels, ('name', 'Name'))
```
lcarva: I think we want to the order to be old-style, then new-style. i.e., `('Name', 'name')`
This will match the order that is used for the component label as well.
It'll also report the new-style label in the error, which is preferred.

osbs/api.py
```diff
@@ -351,25 +352,30 @@ def create_prod_build(self, git_uri, git_ref,
...
         """
+
+        def get_required_label(labels, names):
+            for label_name in names:
+                if label_name in labels:
+                    return labels[label_name]
+            
+            raise OsbsValidationException("required label '{name}' missing "
+                                          "from Dockerfile"
+                                          .format(name=label_name))
```
lcarva: I'm always wary about using a variable that is defined in a for-loop statement.
If names is empty, the for loop will not set `label_name` which would cause this statement to raise something like: `NameError: name 'label_name' is not defined`

This is a nested method so it's unlikely to hit this scenario, but if this is a concern, checks name `names` is truthy and raise a `ValueError` with a meaningful message.

osbs/api.py
```diff
@@ -351,25 +352,30 @@ def create_prod_build(self, git_uri, git_ref,
...
         """
+
+        def get_required_label(labels, names):
+            for label_name in names:
+                if label_name in labels:
+                    return labels[label_name]
+            
+            raise OsbsValidationException("required label '{name}' missing "
+                                          "from Dockerfile"
+                                          .format(name=names[1]))
```
lcarva: This method is assuming names has at least two items, but it's not really obvious. It's also unnecessary.

Consider using `names[-1]` instead, and adding to the beginning of function:
```python
if not names:
   raise ValueError('List of label names cannot be empty.')
```

osbs/api.py
```diff
@@ -351,25 +352,30 @@ def create_prod_build(self, git_uri, git_ref,
...
+        :param component: str, not used anymore
+        :param target: str, koji target
         :param architecture: str, build architecture
         :param yum_repourls: list, URLs for yum repos
         :param koji_task_id: int, koji task ID requesting build
         :param scratch: bool, this is a scratch build
         :return: BuildResponse instance
         """
+
+        def get_required_label(labels, names):
```
twaugh: Maybe add this?:
```
assert names  # always called with a non-empty literal tuple so names[-1] is safe
```
Then when reviewing this later when we've all forgotten what it's about, we'll know that `names[-1]` is safe. This sort of pattern (using a variable defined in a for loop) is the kind of thing a static analysis tool will wonder about too.

osbs/api.py
```diff
@@ -351,25 +352,30 @@ def create_prod_build(self, git_uri, git_ref,
...
+            for label_name in names:
+                if label_name in labels:
+                    return labels[label_name]
+
```
twaugh: Think this line has whitespace in it.

tests/test_api.py
```diff
@@ -154,9 +154,34 @@ class MockParser(object):
                                    TEST_GIT_BRANCH, TEST_USER,
                                    TEST_COMPONENT, TEST_TARGET, TEST_ARCH)
 
+
+    @pytest.mark.parametrize('label_name', ['BZComponent', 'com.redhat.component', 'Name', 'name'])
+    def test_missing_component_and_name_labels(self, osbs, label_name):
```
twaugh: Does this make the `test_create_prod_build_missing_name_label` unit test above it obsolete?

tests/test_api.py
```diff
@@ -179,9 +203,47 @@ class MockParser(object):
...
+        # TODO: test situation when a buildconfig already exists
+        class MockParser(object):
+            labels = {'Name': 'fedora23/something', 'com.redhat.component': TEST_COMPONENT}
+            baseimage = 'fedora23/python'
+        (flexmock(utils)
+            .should_receive('get_df_parser')
+            .with_args(TEST_GIT_URI, TEST_GIT_REF, git_branch=TEST_GIT_BRANCH)
+            .and_return(MockParser()))
+        response = osbs.create_prod_build(TEST_GIT_URI, TEST_GIT_REF,
+                                     TEST_GIT_BRANCH, TEST_USER)
```
twaugh: Indentation doesn't look right.

tests/test_api.py
rcerven: above test is testing, missing any labels, mine test is testing if 1 is missing

PR #489, opened 2016-12-09
Chain rebuild fixes

lcarva:
Addresses a couple of issues:

1. If `builder_build_json_dir` is not defined, it properly falls back to `build_json_dir` from the `[general]` section. This was causing the `import_image` plugin to be configured with `build_json_dir` param as `None`.

2. Removed check for requiring `git_push_url` and `git_push_username` when using triggers. Also removed all references to those as there's no longer an intention to use them. Verified that if these values are in osbs.conf, they are simply ignored.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -77,7 +77,6 @@ class BuildRequest(object):
         :param distribution_scope: str, distribution scope for this image
                                    (private, authoritative-source-only, restricted, public)
         :param use_auth: bool, use auth from atomic-reactor?
-        :param git_push_url: str, URL for git push
         """
 
         # Here we cater to the koji "scratch" build type, this will disable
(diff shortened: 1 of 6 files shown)


osbs/build/build_request.py
```diff
@@ -724,16 +723,8 @@ def render(self, validate=True):
         self.dj.dock_json_set_arg('prebuild_plugins', "pull_base_image", "parent_registry",
                                   source_registry.docker_uri if source_registry else None)
 
-        # The rebuild trigger requires git_branch and git_push_url
-        # parameters, but those parameters are optional. If either was
-        # not provided, remove the trigger.
+        # Set to true to disable triggers in BuildConfig
```
lcarva: Good catch! Added test coverage.

tests/build/test_build_request.py
```diff
@@ -1137,59 +1134,38 @@ def test_render_prod_request_with_trigger(self, tmpdir, registry_uri,
...
+        kwargs['pdc_url'] = 'https://pdc.example.com'
+        kwargs['smtp_uri'] = 'smtp.example.com'
 
         build_request.set_params(**kwargs)
         build_json = build_request.render()
 
-        assert build_json["metadata"]["labels"]["git-branch"] is not None
+        assert build_request.is_custom_base_image() is True
 
-        # Verify the triggers are now disabled
```
twaugh: This comment would still be valid?

tests/build/test_build_request.py
```diff
@@ -1137,59 +1134,38 @@ def test_render_prod_request_with_trigger(self, tmpdir, registry_uri,
...
         assert "triggers" not in build_json["spec"]
 
-        # Verify the rebuild plugins are all disabled
```
twaugh: This comment would still be valid?

tests/build/test_build_request.py
lcarva: Yes, not sure why it was removed.

Issue #485, opened 2016-12-08
osbs version option


rcerven:
add option to osbs to show its version (--version)

PR #491, opened 2016-12-13
add --version option

rcerven:
osbs issue #485 

adds version option for osbs client

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -10,6 +10,7 @@ import collections
 
 import json
 import logging
+import pkg_resources
 
 from textwrap import dedent
 import codecs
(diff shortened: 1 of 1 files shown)


osbs/cli/main.py
```diff
@@ -457,6 +458,11 @@ def str_on_2_unicode_on_3(s):
 
 
 def cli():
+    try:
+        version = pkg_resources.get_distribution("osbs-client").version
+    except pkg_resources.DistributionNotFound:
```
lcarva: In what scenario would this exception be triggered? When I have it installed from git, `get_distribution` is still able to find it properly.

rcerven: it is when it is not installed

twaugh:
Does it make sense to test it?

rcerven:
well all you have to test is osbs --version :)

twaugh:
I was thinking about unit testing really...

rcerven:
in that case it doesn't make sense to test

PR #492, opened 2016-12-16
Ability to ensure a tag is present in ImageStream

lcarva:


lcarva:
I'll address the py2.6 string formatting issues and `test_ensure_image_stream_tag` (which I can't seem to reproduce locally...)

osbs/api.py
```diff
@@ -589,6 +589,18 @@ def get_image_stream_tag(self, tag_id):
         return self.os.get_image_stream_tag(tag_id)
 
     @osbsapi
+    def put_image_stream_tag(self, tag_id, tag):
```
twaugh: Would `update_image_stream_tag()` be a better name? The caller is only expected to use this to update an existing tag I think (`ensure_*` handles creating new ImageStreamTags).

Could have a docstring explaining how the parameters are used.

Is `tag` intended to be a JSON string or an object? I feel like it ought to be an object, as this core function's implementation is the only place that JSON is mentioned.

osbs/api.py
```diff
@@ -589,6 +589,18 @@ def get_image_stream_tag(self, tag_id):
...
+        return self.os.put_image_stream_tag(tag_id, tag)
+
+    @osbsapi
+    def ensure_image_stream_tag(self, stream, tag_id, scheduled=False):
```
twaugh: Could have a docstring to explain the parameters.

osbs/core.py
```diff
@@ -705,6 +705,54 @@ def get_image_stream_tag(self, tag_id):
...
+        url = self._build_url("imagestreamtags/%s" % tag_id)
+        response = self._put(url, data=tag,
+                             headers={"Content-Type": "application/json"})
+        check_response(response)
+        return response
+
+    def ensure_image_stream_tag(self, stream, tag_id, tag_template, scheduled=False):
+        stream_id = stream['metadata']['name']
+        insecure = (stream['metadata']
+                        .get('annotations', {})
```
twaugh: Indentation could line up nicely instead.

lcarva: I thought this was indented nicely :(

Any of these better?

```python
insecure = (stream['metadata'].get('annotations', {})
               .get('openshift.io/image.insecureRepository')
            == 'true')
```


```python
insecure = (
    stream['metadata'].get('annotations', {})
                      .get('openshift.io/image.insecureRepository')
    == 'true')
```

rcerven: this seems to be more readable 

```
insecure = (stream['metadata'].get('annotations', {})
            .get('openshift.io/image.insecureRepository')
            == 'true')
```

twaugh: Whatever makes pep8's E127 error go away I'll be happy with :)

osbs/core.py
```diff
@@ -705,6 +705,54 @@ def get_image_stream_tag(self, tag_id):
...
         return response
 
+    def put_image_stream_tag(self, tag_id, tag):
+        url = self._build_url("imagestreamtags/%s" % tag_id)
+        response = self._put(url, data=tag,
+                             headers={"Content-Type": "application/json"})
+        check_response(response)
+        return response
+
+    def ensure_image_stream_tag(self, stream, tag_id, tag_template, scheduled=False):
```
twaugh: I think the API needs to make sense all together, so let's just try to aim for consistency. Something doesn't quite feel right about these two. The calling code shouldn't have to know very much about the details of how the objects are structured, so expecting it to format JSON seems too much.

osbs/api.py
```diff
@@ -589,6 +589,34 @@ def get_image_stream_tag(self, tag_id):
...
+        """
+        return self.os.put_image_stream_tag(tag_id, tag)
+
+    @osbsapi
+    def ensure_image_stream_tag(self, stream, tag_id, scheduled=False):
+        """Ensures the tag is monitored in ImageStream
+
+        :param stream: dict, ImageStream object
+        :param tag_id: str, name of tag to check, without name of
+                            ImageStream as prefix
```
lcarva: Agree. Using `tag_id` vs `tag_name`

osbs/core.py
lcarva: Agree. I removed the requirement to convert to Json. Instead, this is now done internally right before calling the actual REST API. This should make usage easier.

osbs/api.py
```diff
@@ -589,6 +589,34 @@ def get_image_stream_tag(self, tag_id):
         return self.os.get_image_stream_tag(tag_id)
 
     @osbsapi
+    def put_image_stream_tag(self, tag_id, tag):
+        """Create or modify ImageStreamTag object
+
+        :param tag_id: str, name of ImageStreamTag prefixed by name of
```
twaugh: Nit: should "prefixed by" be "including"? The name of the ImageStreamTag is 'imagestream:tag'.

osbs/api.py
twaugh: What will use this API method? I see why the ensure method would be used, but I'm not sure I see why this method would be used on its own.

lcarva: It's a building block. We could not expose it in `api.py`, and only leave it as a helper method in `core.py`. But I think it's generic enough that it could be used in the future for a different use case.

osbs/api.py
lcarva: Yes, that is better.

PR #495, opened 2017-01-05
Set osbs-client version in the build json

### LINKED ISSUE #477: Inject osbs-client version into build JSON
The build JSON does not give any indication of which osbs-client version constructed it, other than the clues in its structure and contents.

It should explicitly include its own version in the build JSON it creates.

vrutkovs:
Fixes #477

### DIFF
--- a/inputs/prod_inner.json
+++ b/inputs/prod_inner.json
@@ -1,4 +1,5 @@
 {
+  "client-version": "{{VERSION}}",
   "prebuild_plugins": [
     {
       "args": {
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -27,6 +27,7 @@ from osbs.build.spec import BuildSpec
 from osbs.constants import SECRETS_PATH, DEFAULT_OUTER_TEMPLATE, DEFAULT_INNER_TEMPLATE, DEFAULT_CUSTOMIZE_CONF
 from osbs.exceptions import OsbsException, OsbsValidationException
 from osbs.utils import looks_like_git_hash, git_repo_humanish_part_from_uri
+from osbs import __version__ as client_version
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


tests/build/test_build_request.py
```diff
@@ -1465,3 +1465,7 @@ def test_render_prod_custom_site_plugin_override(self):
 
         assert unmodified_build_request.dj.dock_json[plugin_type][plugin_index]['name'] == plugin_name
         assert build_request.dj.dock_json[plugin_type][plugin_index]['name'] == plugin_name
+
+    def test_has_version(self):
+        br = BuildRequest(INPUTS_PATH)
+        assert 'version' in br.dj.dock_json
```
twaugh: Maybe also check its type and value?

inputs/prod_inner.json
```diff
@@ -1,4 +1,5 @@
 {
+  "version": "{{VERSION}}",
```
twaugh: Is `version` descriptive enough or should it be `client-version`? To me `version` sounds like it might be specified a schema version or something.

tests/build/test_build_request.py
vrutkovs: Right, done

osbs/build/build_request.py
```diff
@@ -27,6 +27,7 @@
 from osbs.constants import SECRETS_PATH, DEFAULT_OUTER_TEMPLATE, DEFAULT_INNER_TEMPLATE, DEFAULT_CUSTOMIZE_CONF
 from osbs.exceptions import OsbsException, OsbsValidationException
 from osbs.utils import looks_like_git_hash, git_repo_humanish_part_from_uri
+from osbs import __version__ as version
```
twaugh: This causes a landscape.io warning for remove_tag_and_push_registries() which takes a parameter of the same name.

Maybe we could import it as client_version?

vrutkovs: Ah, right, done

PR #496, opened 2017-01-06
Enable add_help plugin

vrutkovs:
Related: https://github.com/projectatomic/atomic-reactor/pull/613

### DIFF
--- a/inputs/prod_inner.json
+++ b/inputs/prod_inner.json
@@ -45,6 +45,10 @@
       "name": "change_from_in_dockerfile"
     },
     {
+      "name": "add_help",
+      "required": false
+    },
+    {
       "name": "add_dockerfile"
     },
     {


inputs/prod_inner.json
```diff
@@ -48,6 +48,10 @@
       "name": "add_dockerfile"
     },
     {
+      "name": "add_help",
+      "required": false
```
twaugh: This is listed after `add_dockerfile` and I initially thought it would add in the Dockerfile as it was prior to the help ADD instruction being added, but that plugin instructs the Dockerfile to store itself inside the built image and does not take a copy so that seems fine.

vrutkovs: Right, this seems a little confusing. Not sure if it would be better to move this plugin before `add_dockerfile` to show this explicitly though

twaugh: It would make it easier to understand at first glance, yes.

vrutkovs: Done

twaugh:
Looks great.

PR #497, opened 2017-01-06
Enable import_image plugin for custom base images

lcarva:


osbs/build/build_request.py
```diff
@@ -377,17 +377,34 @@ def adjust_for_registry_api_versions(self):
     def adjust_for_triggers(self):
         """
         Remove trigger-related plugins if no triggers set
+        import_image plugin is only removed for non-custom base images
```
twaugh: I'm wondering if it ever makes sense to turn off this plugin.

base -> imageA -> imageB

If imageA does not get automatically triggered, but imageB does, what will ever trigger imageB?

osbs/build/build_request.py
```diff
@@ -377,17 +377,34 @@ def adjust_for_registry_api_versions(self):
...
-            for when, which in [("prebuild_plugins", "check_and_set_rebuild"),
-                                ("prebuild_plugins", "stop_autorebuild_if_disabled"),
-                                ("prebuild_plugins", "update_parent_image_stream_tag"),
-                                ("postbuild_plugins", "import_image"),
-                                ("exit_plugins", "sendmail")]:
+            remove_plugins.append(("postbuild_plugins", "import_image"))
+            for when, which in remove_plugins:
                 logger.info("removing %s from request because there are no triggers",
                             which)
                 self.dj.remove_plugin(when, which)
```
twaugh: I guess these two loops only differ in the reason they log about. Could we avoid repetition by having the `if`s decide on both the list of plugins to remove, and the reason for doing it?

vrutkovs: +1, this would improve code's readability

osbs/build/build_request.py
lcarva: That's a good point. As is, it'll only get turned off if there are no `triggers` in `spec` (meaning autorebuild is disabled in the system). We could leave it enabled in this case as well, but it provides no benefit.

osbs/build/build_request.py
lcarva: Agree. Changes done.

osbs/build/build_request.py
twaugh: If 'base' will run import_image, and 'imageA' has opted out of triggers, but 'imageB' has opted in to triggers, imageB will never be triggered.

In this situation, wouldn't it make sense for 'imageA' to run import_image *even though* it does not want to be triggered by its parent?

Or does this not matter because import_image only does useful things for base images (which might have additional tags added outside of the build process)? If that's the case, we should probably explicitly only run import_image for base images...

lcarva: Whether or not triggers are used is defined per environment, not per image. 
When autorebuild is enabled in a environment, all images will have triggers in their BuildConfig - even images that have opted out of autorebuild. 
The only exception is custom base images - since no other images trigger it, the triggers are always removed to avoid incorrectly tracking `koji/image-build`.

twaugh: Ah, that's what I was missing
Thanks.

osbs/build/build_request.py
```diff
@@ -377,16 +377,32 @@ def adjust_for_registry_api_versions(self):
...
+            ("exit_plugins", "sendmail"),
+        ]
+
+        should_remove = False
+        if triggers and self.is_custom_base_image():
+            msg = "removing %s from request because custom base image"
+            del self.template['spec']['triggers']
+            should_remove = True
+
         if not triggers:
```
twaugh: Could we make this `elif not triggers:` to make it clear that we don't ever want both blocks to run?

osbs/build/build_request.py
twaugh: So is `import_image` removed for *all* images, or only non-base images? I'm still confused after reading the docstring and the code. :(

osbs/build/build_request.py
lcarva: Yes, it should've been. Thanks for catching it.

osbs/build/build_request.py
lcarva: I cleared up the docstring, let me know if it makes more sense now.

Issue #499, opened 2017-01-10
unified label names


rcerven:
label class for to handle multiple label names

PR #500, opened 2017-01-10
unified labels names

rcerven:
#499, creates label class for use across whole osbs

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -361,26 +361,15 @@ class OSBS(object):
         :return: BuildResponse instance
         """
 
-        def get_required_label(labels, names):
-            """
-            get value of label list, first found is returned
-            names has to be non empty tuple
-            """
-            assert names  # always called with a non-empty literal tuple so names[0] is safe
-            for label_name in names:
-                if label_name in labels:
-                    return labels[label_name]
-
-            raise OsbsValidationException("required label '{name}' missing "
-                                          "from Dockerfile"
-                                          .format(name=names[0]))
-
         df_parser = utils.get_df_parser(git_uri, git_ref, git_branch=git_branch)
         build_request = self.get_build_request()
-        labels = df_parser.labels
+        labels = utils.Labels(df_parser.labels)
 
-        name_label = get_required_label(labels, ('name', 'Name'))
-        component = get_required_label(labels, ('com.redhat.component', 'BZComponent'))
+        try:
+            _, name_value = labels.get_name_and_value(utils.Labels.LABEL_TYPE_NAME)
+            _, component = labels.get_name_and_value(utils.Labels.LABEL_TYPE_COMPONENT)
+        except KeyError:
+            raise OsbsValidationException("required label missing from Dockerfile")
 
         build_request.set_params(
             git_uri=git_uri,
@@ -391,7 +380,7 @@ class OSBS(object):
             build_image=self.build_conf.get_build_image(),
             build_imagestream=self.build_conf.get_build_imagestream(),
             base_image=df_parser.baseimage,
-            name_label=name_label,
+            name_label=name_value,
             registry_uris=self.build_conf.get_registry_uris(),
             registry_secrets=self.build_conf.get_registry_secrets(),
             source_registry_uri=self.build_conf.get_source_registry_uri(),
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -367,3 +367,85 @@ def run_command(*popenargs, **kwargs):
             message="Command %s returned %s\n\n%s" % (cmd, retcode, output)
         )
     return output
+
+
+class Labels(object):
+    """
+    Provide access to a set of labels which have specific semantics
+
+    The set of labels comes from here:
+    https://github.com/projectatomic/ContainerApplicationGenericLabels
+
+    Note that only a subset of label types (those used by OSBS) is supported:
+
+    - LABEL_TYPE_NAME: repository name of the image
+    - LABEL_TYPE_VERSION: version of the image
+    - LABEL_TYPE_RELEASE: release number for this version
+    - LABEL_TYPE_ARCH: architecture for the image
+    - LABEL_TYPE_VENDOR: owner of the image
+    - LABEL_TYPE_SOURCE: authoritative location for publishing
+    - LABEL_TYPE_COMPONENT: Bugzilla (or other tracker) component
+    - LABEL_TYPE_HOST: build host used to create the image
+    """
+    LABEL_TYPE_NAME = object()
+    LABEL_TYPE_VERSION = object()
+    LABEL_TYPE_RELEASE = object()
+    LABEL_TYPE_ARCH = object()
+    LABEL_TYPE_VENDOR = object()
+    LABEL_TYPE_SOURCE = object()
+    LABEL_TYPE_COMPONENT = object()
+    LABEL_TYPE_HOST = object()
+    LABEL_NAMES = {
+        LABEL_TYPE_NAME: ('name', 'Name'),
+        LABEL_TYPE_VERSION: ('version', 'Version'),
+        LABEL_TYPE_RELEASE: ('release', 'Release'),
+        LABEL_TYPE_ARCH: ('architecture', 'Architecture'),
+        LABEL_TYPE_VENDOR: ('vendor', 'Vendor'),
+        LABEL_TYPE_SOURCE: ('authoritative-source-url', 'Authoritative_Registry'),
+        LABEL_TYPE_COMPONENT: ('com.redhat.component', 'BZComponent'),
+        LABEL_TYPE_HOST: ('com.redhat.build-host', 'Build_Host')
+    }
+
+    def __init__(self, df_labels):
+        """
+        Create a new Labels object
+        providing access to actual newest labels as well as old ones
+        """
+        self._df_labels = df_labels
+        self._label_values = {}
+        for label_type, label_names in Labels.LABEL_NAMES.items():
+            for lbl_name in label_names:
+                if lbl_name in df_labels:
+                    self._label_values[label_type] = (lbl_name, df_labels[lbl_name])
+                    break
+
+    def get_name(self, label_type):
+        """
+        returns the most preferred label name
+        if there isn't any correct name in the list
+        it will return newest label name
+        """
+        if label_type in self._label_values:
+            return self._label_values[label_type][0]
+        else:
+            return Labels.LABEL_NAMES[label_type][0]
+
+    @staticmethod
+    def get_new_names_by_old():
+        """Return dictionary, new label name indexed by old label name."""
+        newdict = {}
+
+        for label_type, label_names in Labels.LABEL_NAMES.items():
+            for oldname in label_names[1:]:
+                newdict[oldname] = Labels.LABEL_NAMES[label_type][0]
+        return newdict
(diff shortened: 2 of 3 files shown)


twaugh:
Relates to: https://github.com/projectatomic/atomic-reactor/pull/606

twaugh:
Let's all be clear on what the interface is supposed to look like before making any more code changes.

twaugh:
My original proposal for it: https://github.com/projectatomic/atomic-reactor/pull/606#issuecomment-270402677

tests/test_utils.py
```diff
@@ -239,3 +239,53 @@ def test_get_instance_token_file_name():
...
 
     assert get_instance_token_file_name('spam') == expected
+
+@pytest.mark.parametrize(('labels', 'fnc', 'expect'), [
+        ({},
+         ("get_label_only", Labels.LABEL_TYPE_COMPONENT),
+         "com.redhat.component"),
+        ({},
+         ("get_label_only", "new_label"),
+         "new_label"),
```
twaugh: No, this test is wrong. "new label" is not a valid key.

tests/test_utils.py
```diff
@@ -239,3 +239,53 @@ def test_get_instance_token_file_name():
...
+        ({},
+         ("get_label_only", ""),
+         ""),
```
twaugh: "" is not a valid key. This test should be removed.

tests/test_utils.py
```diff
@@ -239,3 +239,53 @@ def test_get_instance_token_file_name():
...
+          "Version": "version", "Architecture": "architecture",
+          "Release": "release", "BZComponent": "com.redhat.component",
+          "Authoritative_Registry": "authoritative-source-url"}),
+        ({"Name" : "old",
+          "name" : "new"},
+          ("get_label_and_value", Labels.LABEL_TYPE_NAME),
+          ("name", "new")),
+        ({},
+          ("get_label_and_value", "doest_exist"),
+          KeyError),
```
twaugh: "doest_exist" is not a valid key -- calling this method with some invalid key is not defined behaviour and we should not test it.

tests/test_utils.py
```diff
@@ -239,3 +239,53 @@ def test_get_instance_token_file_name():
...
+        ({"some_label" : "some_value"},
+          ("get_label_and_value", "some_label"),
+          ("some_label", "some_value"))
```
twaugh: "some_label" is not a valid key.

The only valid keys are the ones defined in the Labels class.

tests/test_utils.py
```diff
@@ -239,3 +239,53 @@ def test_get_instance_token_file_name():
...
+    ])
```
twaugh: Missing test case:
```
({},
 ("get_label_and_value", Labels.TYPE_LABEL_NAME),
 KeyError),
```

tests/test_utils.py
```diff
@@ -239,3 +239,47 @@ def test_get_instance_token_file_name():
...
 
     assert get_instance_token_file_name('spam') == expected
+
+@pytest.mark.parametrize(('labels', 'fnc', 'expect'), [
+        ({},
+         ("get_label_only", Labels.LABEL_TYPE_COMPONENT),
+         "com.redhat.component"),
+        ({},
+         ("get_label_only", "doesnt_exist"),
+         KeyError),
```
twaugh: Maybe `Exception`? We're outside the specification of how this function ought to behave, so if it were to raise e.g. `TypeError` that would be perfectly valid behaviour and the test should still pass.

tests/test_utils.py
twaugh: If you want to test this, expect `Exception` rather than `KeyError`.

tests/test_utils.py
```diff
@@ -239,3 +239,47 @@ def test_get_instance_token_file_name():
...
+          "Authoritative_Registry": "authoritative-source-url"}),
+        ({"Name" : "old",
+          "name" : "new"},
+          ("get_label_and_value", Labels.LABEL_TYPE_NAME),
+          ("name", "new")),
+        ({},
+          ("get_label_and_value", "doest_exist"),
+          KeyError),
+    ])
+def test_labels(labels, fnc, expect):
```
twaugh: Missing test: get_label_and_value called with a valid label type but for which the Dockerfile provides no label.

twaugh: Should be:
```
({},
 ("get_label_and_value", Labels.LABEL_TYPE_NAME),
 KeyError),
```

tests/test_utils.py
```diff
@@ -239,3 +239,50 @@ def test_get_instance_token_file_name():
...
+          "Version": "version", "Architecture": "architecture",
+          "Release": "release", "BZComponent": "com.redhat.component",
+          "Authoritative_Registry": "authoritative-source-url"}),
+        ({"Name" : "old",
+          "name" : "new"},
+         ("get_label_and_value", Labels.LABEL_TYPE_NAME),
+         ("name", "new")),
+        ({},
+         ("get_label_and_value", Labels.LABEL_TYPE_NAME),
+         Exception),
```
twaugh: We can be more specific here -- we definitely want this to be an instance of `KeyError`.

tests/test_utils.py
```diff
@@ -239,3 +239,50 @@ def test_get_instance_token_file_name():
...
+         KeyError),
+        ({},
+         ("get_label_and_value", "doest_exist"),
+         Exception),
+    ])
+def test_labels(labels, fnc, expect):
+    label = Labels(labels)
+
+    fn, arg = fnc
+    if isinstance(expect, type):
```
twaugh: Should this be `isinstance(expect, Exception)`?

osbs/utils.py
```diff
@@ -358,3 +358,68 @@ def run_command(*popenargs, **kwargs):
             message="Command %s returned %s\n\n%s" % (cmd, retcode, output)
         )
     return output
+
+
+class Labels(object):
```
twaugh: Could include a class docstring here, something like:
```
    """
    Provide access to a set of labels which have specific semantics

    The set of labels comes from here:
    https://github.com/projectatomic/ContainerApplicationGenericLabels

    Note that only a subset of label types (those used by OSBS) is supported:

    - LABEL_TYPE_NAME: repository name of the image 
    - LABEL_TYPE_VERSION: version of the image
    - LABEL_TYPE_RELEASE: release number for this version
    - LABEL_TYPE_ARCH: architecture for the image
    - LABEL_TYPE_VENDOR: owner of the image
    - LABEL_TYPE_SOURCE: authoritative location for publishing
    - LABEL_TYPE_COMPONENT: Bugzilla (or other tracker) component
    - LABEL_TYPE_HOST: build host used to create the image
    """
```

tests/test_utils.py
rcerven: but I am passing class of exception not instance

twaugh: Ah, right. OK, this is fine as it is.

twaugh:
This looks good!

Seeing as osbs-client only has one place where it looks at labels (OSBS.create_prod_build in osbs/api.py), how about modifying that to use this new class?

twaugh:
Looks good.

Now, before we merge, is everyone happy with the method names? One issue I see is that they have different ways of referring to the name of the label -- `label` in `get_label_only()` vs `name` in `get_name_and_value()` and `get_new_names_by_old()`. Do we want to make that consistent?

vrutkovs:
>Do we want to make that consistent?

I think `label` should be present, as `get_label_name_and_value` (maybe `get_label_items`?) and `get_label_names_by_old` sound better to me

osbs/api.py
```diff
@@ -361,26 +361,15 @@ def create_prod_build(self, git_uri, git_ref,
...
-
         df_parser = utils.get_df_parser(git_uri, git_ref, git_branch=git_branch)
         build_request = self.get_build_request()
-        labels = df_parser.labels
+        labels = utils.Labels(df_parser.labels)
 
-        name_label = get_required_label(labels, ('name', 'Name'))
-        component = get_required_label(labels, ('com.redhat.component', 'BZComponent'))
+        try:
+            _, name_label = labels.get_label_and_value(utils.Labels.LABEL_TYPE_NAME)
```
vrutkovs: `name_label` here is actually label's value

rcerven:
maybe get_name_only() instead get_label_only() ?

rcerven:
get_label_items() instead get_label_name_and_value() is nice

but as for get_new_names_by_old()
this seems to be nicely describing what it returns

lcarva:
`get_label_items()` is confusing to me, because it implies it'll behave similarly to `dict.items()` which is not the case since it only returns one item, perhaps `get_label_item()` is more accurate? But I still think `get_label_name_and_value()` better describes the method behavior. To keep it consistent I think either we should have `_label_` in both of these methods or omit them, not a mix. For example:

```python
get_name_and_value()
get_name()
```
or
```python
get_label_name_and_value()
get_label_name()
```

rcerven:
this seem nice:
get_name_and_value()
get_name()

twaugh:
There's something to be said about *not* having `label` in the method name -- after all, it's already in the class name.

vrutkovs:
+1 to `get_name` and `get_name_and_value`

rcerven:
seems like all like : 
get_name_and_value()
get_name()

so what about : 
get_new_names_by_old()

vrutkovs:
@rcerven, how about `get_preferred` or `get_preferred_for_old`?

rcerven:
preferred was before used as "best option" so not always 'new'

vrutkovs:
`_PREFERRED_LABELS` is sorted like that, I guess its safe to assume that 'new' == 'preferred'

osbs/utils.py
```diff
@@ -358,3 +358,85 @@ def run_command(*popenargs, **kwargs):
...
+
+    def __init__(self, df_labels):
+        """
+        Create a new Labels object
+        providing access to actual newest labels as well as old ones
+        """
+        self._df_labels = df_labels
+        self._label_values = {}
+        for label_type in Labels.LABEL_NAMES:
+            for lbl_name in Labels.LABEL_NAMES[label_type]:
```
lcarva: Could be nicer to use `items()` here:

```python
for label_type, label_names in Labels.LABEL_NAME.items():
  for lable_name in label_names:
...
```

osbs/utils.py
```diff
@@ -358,3 +358,85 @@ def run_command(*popenargs, **kwargs):
...
+
+    @staticmethod
+    def get_new_names_by_old():
+        """Return dictionary, new label name indexed by old label name."""
+        newdict = {}
+
+        for label_type, label_names in Labels.LABEL_NAMES.items():
+            for oldname in label_names[1:]:
+                newdict[oldname] = Labels.LABEL_NAMES[label_type][0]
+        return newdict
```
vrutkovs: This method returns a dict, unlike others. 

Why not just return a new name instead?
Where would this method be used?

twaugh:
If we can't agree on how `get_new_names_by_old()` should behave or should be named, how about dropping this method until we actually want to use it?

osbs/utils.py
rcerven: this is suppose to replace 
in pre_add_labels_in_df.py
DEFAULT_ALIASES dictionary

rcerven:
squashed

PR #504, opened 2017-01-17
prod.json: set runPolicy explicitely

vrutkovs:
This should help us to fight https://bugzilla.redhat.com/show_bug.cgi?id=1400132

It seems the builds stuck in New are caused by the following code path:
 * If the BuildConfig has no runPolicy set, Serial policy [will be assigned to it](https://github.com/openshift/origin/blob/master/pkg/build/controller/controller.go#L100)
 * When the build is complete OSE calls [OnComplete](https://github.com/openshift/origin/blob/master/pkg/build/controller/controller.go#L106) handler, which in its turn starts Policy's [handleComplete](https://github.com/openshift/origin/blob/master/pkg/build/controller/policy/policy.go#L130) handler
 * In `handleComplete` OSE tries to find the next build to run by listing all builds and looking for a suitable one.
 * Since this situation is happening for almost each build, OSE spends most of its time listing all builds.

Setting a policy is a stab to stop this sooner - run `handleComplete` after the build is moving to Complete and not in BuildController's `HandleBuild`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,9 @@ python:
 install:
   - "pip install -r requirements.txt"
   - "pip install -r tests/requirements.txt"
-  - "pip install pytest-cov coveralls"
+# coverage 4.3.2 fails with py 2.6:
+# https://bitbucket.org/ned/coveragepy/issues/555/coverage-432-claims-support-for-py26-but
+  - "pip install coverage==4.3.1 pytest-cov coveralls"
   - "if [[ ${TRAVIS_PYTHON_VERSION%.*} -gt 2 ]]; then pip install -r requirements-py3.txt; fi"
 # command to run tests
 script: py.test -vv tests --cov osbs
--- a/inputs/prod.json
+++ b/inputs/prod.json
@@ -8,6 +8,7 @@
     }
   },
   "spec": {
+    "runPolicy": "Serial",
     "source": {
       "type": "Git",
       "git": {


twaugh:
Could you include a URL to the upstream bug report tracking the coverage failure in the commit that freezes it? Either in the message or as a comment in the .travis.yml is fine with me:
https://bitbucket.org/ned/coveragepy/issues/555/coverage-432-claims-support-for-py26-but

That way we can track when we can remove that freeze.

vrutkovs:
Done

twaugh:
How about a brief explanation of why this is a useful thing to do? What you told me in IRC was pretty compelling... :)

vrutkovs:
Updated the PR description with a theory on what happens

mmilata:
Hmm, [this function](https://github.com/openshift/origin/blob/v1.3.2/pkg/build/api/v1/defaults.go#L5) should set the policy to Serial whenever BuildConfig is submitted without one. But maybe it doesn't - does this patch improve the issue?

vrutkovs:
>Hmm, this function should set the policy to Serial whenever BuildConfig is submitted without one. 

Intersting, we saw a lot of `Build %!s(MISSING)/%!s(MISSING) does not have start policy label set, using default (Serial)` warnings in log.

>But maybe it doesn't - does this patch improve the issue?

We're not sure - this would take a lot of time to reproduce it again, as all the builds should be removed.

mmilata:
> Intersting, we saw a lot of Build %!s(MISSING)/%!s(MISSING) does not have start policy label set, using default (Serial) warnings in log.

Looking at 3.3.1.6 source, I wonder how can the label be not set. Can this be from a leftovers from times when we created Builds directly instead of instantiating BuildConfigs?

twaugh:
I don't think so -- we've been using BuildConfigs since before this particular troublesome cluster was put in service IIRC.

mmilata:
Fair enough. Then this is perhaps something to investigate. Can you find example of build without the `openshift.io/build.start-policy` label, or with the label set to empty string? Is it different from other builds somehow?

vrutkovs:
Hmm, interesting. I haven't found the build with this annotation yet

mmilata:
> Hmm, interesting. I haven't found the build with this annotation yet

FYI it's a label, not annotation, different dict under the `metadata` key.

vrutkovs:
Added a sample of those builds - https://gist.github.com/vrutkovs/b7e4b020dd48770f9c6a8500ef1f442b

mmilata:
The build without the `openshift.io/build.start-policy` label is half a year older than the other one. The label was introduced in OSE 3.3, is it possible that you were running older version at the time (May 2016)?

vrutkovs:
Yes, I guess we were.

I'm gonna remove those builds on my dev cluster and see if that will unstuck the scheduler

lcarva:
> I don't think so -- we've been using BuildConfigs since before this particular troublesome cluster was put in service IIRC.

Not sure if relevant, but scratch builds are not created from a BuildConfig.

vrutkovs:
>The code path in your PR description is concerned only with new builds though, isn't it?

I'm not sure, the logs say that even some week-old builds were there

mmilata:
I re-read the analysis and it hits the nail on the head (good work!), however I don't think setting the policy explicitly will help. The only place from which `OnComplete` is called is the `HandleBuild` function in build controller, regardless of whether the label is set or not. Am I missing something?

Looking at the code you might be able to work around this by setting `build.spec.strategy.type` to `Jenkins` for old builds but that might open even larger can of worms:))

mmilata:
That said, setting the policy explicitly won't hurt either so +1 for merging this.

vrutkovs:
Surprisingly this seems to have relieved our dev cluster. I collected a list of affected builds:
```
$ cat go_template_file
{{.metadata.name}} - {{index .metadata.labels "openshift.io/build.start-policy"}}
$ oc get builds -a --no-headers -o name | xargs -n1 oc get -o go-template-file=go_template_file | grep "<no value>" > builds
``` 
And then patched those:
```
$ cat builds | cut -d ' ' -f 1 | xargs -n1 oc patch build -p '{"metadata": {"labels": {"'openshift.io/build.start-policy'": "Serial"}}}'
```

(there is probably a cleaner way, but I'm a caveman)

After restarting the master the builds started to get scheduled again. Weird 'Build ... does not have start policy label set' are gone. I'll send some more builds there to verify that and report back in the bugzilla ticket

twaugh:
Restarting the master would cause the builds to get scheduled anyway, wouldn't it? That's exactly the workaround we've been using.

vrutkovs:
In order to verify this without restarting we can also try just waiting for about 30 mins - a new cycle of build scheduling will begin and builds hopefully should get unstuck

twaugh:
Do we think this is no longer needed?

lcarva:
It didn't help, as builds were still stuck in New state.

vrutkovs:
Closing this PR - it didn't really help on a testing cluster, and a proper Openshift fix is on the way

PR #505, opened 2017-01-17
Create parent's ImageStreamTag

lcarva:
Needs testing

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -15,6 +15,7 @@ import stat
 import sys
 import warnings
 import datetime
+import requests
 from functools import wraps
 
 from osbs.build.build_request import BuildRequest
@@ -22,7 +23,7 @@ from osbs.build.build_response import BuildResponse
 from osbs.build.pod_response import PodResponse
 from osbs.constants import BUILD_RUNNING_STATES
 from osbs.core import Openshift
-from osbs.exceptions import OsbsException, OsbsValidationException
+from osbs.exceptions import OsbsException, OsbsValidationException, OsbsResponseException
 # import utils in this way, so that we can mock standalone functions with flexmock
 from osbs import utils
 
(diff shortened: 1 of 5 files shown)


osbs/api.py
```diff
@@ -287,6 +288,36 @@ def _create_scratch_build(self, build_request):
         build_json['metadata']['name'] = build_config_name
         return BuildResponse(self.os.create_build(build_json).json())
 
+    def _get_image_stream_info_for_build_request(self, build_request):
+        """Return ImageStream, ImageStreamTag, and ImageStreamTag name
+        for base_image of build_request
+
+        If build_reuqest is not auto instantiated, objects are not fetched
```
twaugh: typo: request

osbs/api.py
```diff
@@ -287,6 +288,36 @@ def _create_scratch_build(self, build_request):
...
+        and None, None, None is returned.
+        """
+        image_stream = None
+        existing_ist = None
+        image_stream_tag_name = None
+
+        if build_request.is_auto_instantiated():
```
twaugh: Is `is_auto_instantiated()` meant to be accurate at this point? We seem to be doing this calculation across two places. Should it be named `has_ist_trigger()` instead?

lcarva: I think `has_ist_trigger` is a much more accurate to that method, especially because you cannot tell just by the `BuildRequest` that it will be auto instantiated.

osbs/api.py
```diff
@@ -314,25 +362,30 @@ def _create_build_config_and_build(self, build_request):
...
+                         build_config_name)
             self.os.update_build_config(build_config_name, json.dumps(existing_bc))
 
+        if image_stream:
+            changed_ist = self.ensure_image_stream_tag(image_stream,
+                                                       image_stream_tag_name,
+                                                       scheduled=True)
+            logger.debug('Changed parent ImageStreamTag? %s', changed_ist)
+
+        if not existing_bc:
```
vrutkovs: Move this part up to `else` after `if existing_bc`?

lcarva: It's actually in this order for a reason. created.

If the BC already exists and an update is to be performed, we want to update/create IST after BC has been updated. Creating IST will trigger the build instantiation and we want to make sure it's using the latest and greatest BuildConfig.

If the BC does not exist, then IST needs to be created first. This is to avoid two builds being auto instantiated (one for BC creation and one for IST creation). Creating IST first wouldn't trigger build instantiation because well.. BuildConfig doesn't exist yet! But once BuildConfig is created, then it would be auto instantiated as expected.

vrutkovs: I see, so `ensure_image_stream_tag` has to be run before we create a new BC - or after existing is updated, right?

Lets add a comment about that one, as without knowing this its easy to 'optimize' this

lcarva: Agree.

lcarva: Added doc.

lcarva:
@vrutkovs, @twaugh PTAL

tests/test_api.py
```diff
@@ -806,50 +811,166 @@ def test_create_build_config_create(self):
...
+        # build, because this test assumes an auto_instantiated BuildRequest
+        expect_auto = existing_is
+        with NamedTemporaryFile(mode='wt') as fp:
+            fp.write(dedent("""\
+                [general]
+                build_json_dir = inputs
+                """))
+            fp.flush()
+            config = Configuration(fp.name)
+            osbs = OSBS(config, config)
```
twaugh: Nitpick: OSBS() doesn't need to be inside the `with` block.

tests/test_api.py
```diff
@@ -806,50 +811,166 @@ def test_create_build_config_create(self):
...
+                .once()
+                .and_return('build-id'))
+
+            (flexmock(osbs.os)
+                .should_receive('get_build')
+                .with_args('build-id')
+                .once()
+                .and_return(flexmock(json=lambda: {'spam': 'maps'})))
+
+        if not expect_auto:
```
twaugh: `else:`?

osbs/api.py
```diff
@@ -300,7 +329,16 @@ def _create_build_config_and_build(self, build_request):
         logger.debug('build config to be named "%s"', build_config_name)
         existing_bc = self._get_existing_build_config(build_json)
 
-        if existing_bc is not None:
+        image_stream, existing_ist, image_stream_tag_name = \
```
twaugh: landscape.io says `existing_ist` isn't used.

tests/test_api.py
lcarva: Moved it out.

tests/test_api.py
lcarva: changed

osbs/api.py
lcarva: that's because it isn't :)
Modifying `_get_image_stream_info_for_build_request` to not return this value anymore.

PR #510, opened 2017-01-23
enforce version label

### LINKED ISSUE #509: enforce required 'version' label in dockerfile

rcerven:
fixes #509

enforces all required labels at the same place


rcerven:
according our documentation we require name,release,version in Dockerfile

right now only name and release are checked in osbs-client, and if they are missing it fails, but for missing version it doesn't fail here, but later on in some of atomic-reactor's plugins

so as we require all 3, this check in osbs-client should include also 'version',
so they all fail in the same place

tests will be modified, when this change  will be accepted

vrutkovs:
Right, a comment about that should be added to the sources, as now its unclear why do we try to fetch a version, but don't use it later on

rcerven:
it is used later on, but in atomic-reactor

osbs/api.py
```diff
@@ -413,10 +413,29 @@ def create_prod_build(self, git_uri, git_ref,
         build_request = self.get_build_request()
         labels = utils.Labels(df_parser.labels)
 
+        required_missing = False
         try:
             _, name_value = labels.get_name_and_value(utils.Labels.LABEL_TYPE_NAME)
+        except KeyError:
```
vrutkovs: Code duplication. Let try:
```
required_labels = [utils.Labels.LABEL_TYPE_NAME, utils.Labels.LABEL_TYPE_COMPONENT, utils.Labels.LABEL_TYPE_VERSION]
for label in required_labels:
    try:
        labels.get_name_and_value(label)
    except KeyError:
                         labels.get_name(label))

_, name_value = labels.get_name_and_value(utils.Labels.LABEL_TYPE_NAME)
_, component = labels.get_name_and_value(utils.Labels.LABEL_TYPE_COMPONENT)
```

rcerven: that way I'll call the methods 2 times, so is it worth?

vrutkovs: I think it is - the methods are fast, and the clarity of the code is more important in this case. 

@twaugh, @lcarva thoughts?

lcarva: @vrutkovs I think that would still raise an exception on the first miss and not continue with the checks as we want.

How about something a little more dynamic:

```python
labels = {}
required_missing = False
for label in [utils.Labels.LABEL_TYPE_NAME, 
              utils.Labels.LABEL_TYPE_COMPONENT, 
              utils.Labels.LABEL_TYPE_VERSION]:
    try:
        labels[label] = labels.get_name_and_value(label)
    except KeyError:
        required_missing = True
                     labels.get_name(utils.Labels.LABEL_TYPE_COMPONENT))

if required_missing:
    raise OsbsValidationException("required label(s) missing from Dockerfile")

# ... later ...
build_request.set_params(
    name_label=labels[utils.Labels.LABEL_TYPE_NAME],
    component=labels[utils.Labels.LABEL_TYPE_COMPONENT],
```

PR #511, opened 2017-01-25
build config name contains hash

rcerven:


osbs/utils.py
```diff
@@ -341,6 +344,9 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
     while name.endswith('-'):
```
lcarva: This is no longer needed.

osbs/utils.py
```diff
@@ -341,6 +344,9 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
...
         name = name[:-1]
 
+    hash_str = b64encode(sha256(repo + branch).hexdigest())[:hash_size]
```
lcarva: It's more deterministic to compute the hash before repo and branch are truncated.

osbs/utils.py
```diff
@@ -341,6 +344,9 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
...
+    name += separator + hash_str
```
lcarva: Nitpick: it might be clearer to just do:
```python
name = separator.join([repo, branch, hash_str])
```

osbs/utils.py
```diff
@@ -316,18 +318,19 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
...
     repo = git_repo_humanish_part_from_uri(repo)
     branch = branch or 'unknown'
+    hash_size = 5
 
     repo_chars = []
     branch_chars = []
     groups = ((repo, repo_chars), (branch, branch_chars))
 
-    size = len(separator)
+    size = sep_size = len(separator)
```
lcarva: This should just be updated to reflect the minimum `size`:
```python
size = (2 * len(separator)) + hash_size
```

Doing so will also avoid having to determine the limit on every loop iteration.

osbs/utils.py
```diff
@@ -316,18 +318,19 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
 
     repo = git_repo_humanish_part_from_uri(repo)
     branch = branch or 'unknown'
+    hash_size = 5
```
lcarva: Make it a param?

osbs/utils.py
rcerven: well it doesn't hurt to not have there multiple dashes "in the end" (before hash)

lcarva: it's not needed, I'd rather remove it to avoid confusion later on.

osbs/utils.py
```diff
@@ -316,12 +318,14 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
 
     repo = git_repo_humanish_part_from_uri(repo)
     branch = branch or 'unknown'
+    hash_str = b64encode(sha256(repo + branch).hexdigest())[:hash_size]
```
twaugh: From test failures it looks like you need:
```
full = repo + branch
...(sha256(full.encode('utf-8')))...
```

vrutkovs: Why would we want to have both base64 encoding of the hash here? 

I think one of those should be sufficient, any of those would produce Openshift-safe results.

twaugh: It's about packing those 5 characters with useful information. The entropy of the first five characters of the hexdigest is much lower than that of the first five characters of the base64 encoded hexdigest (it uses upper and lower case, and more than the first 6 characters of the alphabet, etc).

twaugh: Wait, I see what you mean. We should be applying b64encode to the actual value of the hash, not to its ASCII representation.

twaugh: So actually this should be more like:
```
full = repo + branch
hash_str = urlsafe_b64encode(sha256(full.encode('utf-8')).digest())[:hash_size]
if hash_str.endswith('-'):
    # Replace final '-' with '_' to make a valid name
    hash_str = hash_str[:len(hash_str) - 1] + '_'
```

urlsafe_b64encode() avoids '/' and '+'.

twaugh:
We also need to this for names for scratch builds.

tests/build/test_spec.py
```diff
@@ -69,7 +69,7 @@ def test_spec_name(self):
         spec = BuildSpec()
         spec.set_params(**kwargs)
 
-        assert spec.name.value == 'reponame-master'
+        assert spec.name.value == 'reponame-master-Hqrbt'
```
twaugh: This line should be removed.

tests/constants.py
```diff
@@ -14,7 +14,7 @@
 
 TEST_BUILD = "test-build-123"
 TEST_CANCELLED_BUILD = "test-build-cancel-123"
-TEST_BUILD_CONFIG = "path-master"
+TEST_BUILD_CONFIG = "path-master-zR5Oj"
```
twaugh: Any asserts about this name should be removed.

osbs/utils.py
```diff
@@ -316,12 +318,14 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
 
     repo = git_repo_humanish_part_from_uri(repo)
     branch = branch or 'unknown'
+    hash_str = b64encode(sha256((repo + branch).encode('utf-8')).digest()).decode('ascii')[:hash_size]
```
twaugh: Can we split this line into smaller pieces so it's easier to see what's going on?

While doing that, let's make sure to avoiding having the base64-encoded value contain any characters OpenShift won't like: specify `altchars=b'__'`.

osbs/utils.py
```diff
@@ -316,12 +318,16 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
 
     repo = git_repo_humanish_part_from_uri(repo)
     branch = branch or 'unknown'
+    full = (repo + branch).encode('utf-8')
+    shaval = sha256(full).digest()
+    hash_str = b64encode(shaval).decode('ascii')[:hash_size]
```
twaugh: `b64encode(shaval, altchars=b'__')`

tests/build/test_spec.py
```diff
@@ -69,7 +69,7 @@ def test_spec_name(self):
         spec = BuildSpec()
         spec.set_params(**kwargs)
 
-        assert spec.name.value == 'reponame-master'
+        assert spec.name.value.startswith('reponame-master')
```
twaugh: I even think we don't need this assert at all.

tests/test_utils.py
```diff
@@ -94,9 +95,15 @@ def test_get_time_from_rfc3339_valid(rfc3339, seconds, tz):
     ('spam', '-----n', 10, '-', 'spam'),
     ('https://github.com/blah/spam.git', 'bacon', 10, '-', 'spam-bacon'),
 ])
-def test_make_name_from_git(repo, branch, limit, separator, expected):
-    assert make_name_from_git(repo, branch, limit, separator) == expected
+def test_make_name_from_git(repo, branch, limit, separator, expected, hash_size=5):
+    bc_name = make_name_from_git(repo, branch, limit + len(separator) + hash_size, separator, hash_size=hash_size)
 
+    assert re.match(re.escape(expected + separator) + (hash_size * '.'), bc_name)
```
twaugh: If we're doing this level of manipulation, we don't need regular expressions at all. Just trim off 'hash_size + 1' characters from the end before comparison.

tests/constants.py
```diff
@@ -14,7 +14,7 @@
 
 TEST_BUILD = "test-build-123"
 TEST_CANCELLED_BUILD = "test-build-cancel-123"
-TEST_BUILD_CONFIG = "path-master"
+TEST_BUILD_CONFIG = "path-master-....."
```
twaugh: Also needs an equivalent change in tests/fake_api.py to use matching instead of string comparison when looking up URL resources.

tests/build/test_spec.py
rcerven: well it is test if Buildspec is returning correct name

osbs/utils.py
```diff
@@ -316,12 +318,16 @@ def make_name_from_git(repo, branch, limit=53, separator='-'):
 
     repo = git_repo_humanish_part_from_uri(repo)
     branch = branch or 'unknown'
+    full = (repo + branch).encode('utf-8')
+    shaval = sha256(full).digest()
```
twaugh: Nitpick: I'd rather have
```
full = repo + branch
shaval sha256(full.encode('utf-8')).digest()
```
Not a blocker for merging.

PR #516, opened 2017-01-30
build config name contains hash fix

rcerven:


twaugh:
Will you add an assert statement to verify the name is in the format OpenShift will accept?

rcerven:
I'll do it in that another PR, which will actually resolve problem, so I'll make unit tests for that as well

vrutkovs:
Not sure if removing base64 encoding is a good idea, as we might have hash collision. The alternative way is to `.lower()` the result instead.

Lets see the test results for that

twaugh:
I'd rather not mess around with the output of base64. If we need more bits, let's use more characters for the hexdigest.

rcerven:
I have tested it for all our repo/branches and all sha were unique

tests/test_utils.py
```diff
@@ -104,6 +104,20 @@ def test_make_name_from_git_collide():
...
 
+SHA_INPUT_FILE = 'tests/input_for_sha.txt'
+
+def test_make_name_from_git_all_from_file():
+
+    all_sha = set()
+    with open(SHA_INPUT_FILE) as f:
+        lines =  f.read().splitlines()
+
+    for line in lines:
```
twaugh: Could use:
```
with open(SHA_INPUT_FILE) as f:
    for line in f:
```

rcerven: but I need line count also, so I can either count lines in that loop or leave it as it is

twaugh: Ah yes. Let's leave it as-is.

lcarva: BTW, `enumerate()` is a great way to achieve that.

```python
index = -1
for index, line in enumerate(f):
  ...
number_of_lines = index + 1
```

PR #517, opened 2017-02-03
unified version

rcerven:
when we are doing release 'version' of osbs-client has to be changed in 3 places, spec, setup, __init__

this way we can reuse version from __init__


maxamillion:
👍 LGTM

setup.py
```diff
@@ -12,6 +12,7 @@
 import glob
 
 from setuptools import setup, find_packages
+from osbs import __version__ as osbs_client_version
```
lcarva: It's bad form to import `osbs` from `setup.py`, since it actually loads the module prior to installing it.
If that file ever imports something from `requirements.txt`, this will cease to work.

If we do go down this approach, let's add a big fat comment to `osbs/__init__.py` to clarify this.

PR #519, opened 2017-02-09
Avoid invalid BuildConfig names

twaugh:
When creating a BuildConfig name avoid underscores, which are not allowed at all, and periods, which are only allowed when followed by certain characters.

### DIFF
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -289,6 +289,10 @@ def get_time_from_rfc3339(rfc3339):
         return timegm(time_tuple)
 
 
+
+VALID_BUILD_CONFIG_NAME_CHARS = re.compile('[-a-z0-9]')
+
+
 def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
     """
     return name string representing the given git repo and branch
@@ -303,9 +307,10 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
     Assuming '-XXXX' (5 chars) and '-build' (6 chars) as default
     suffixes, name should be limited to 53 chars (64 - 11).
 
-    Since OpenShift rejects names that end in dash, the
-    generated name may be trimmed resulting in a name shorter
-    than the limit.
+    OpenShift is very peculiar in which BuildConfig names it
+    allows. For this reason, only certain characters are allowed.
+    Any disallowed characters will be removed from repo and
+    branch names.
 
     :param repo: str, the git repository to be used
     :param branch: str, the git branch to be used
@@ -320,6 +325,10 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
     full = repo + branch
     shaval = sha256(full.encode('utf-8')).hexdigest()
     hash_str = shaval[:hash_size]
+
+    # Sanitize repo and branch names
+    repo = ''.join(filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(repo)))
+    branch = ''.join(filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(branch)))
 
     repo_chars = []
     branch_chars = []
@@ -338,11 +347,9 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
             continue
         break
 
-    repo = ''.join(repo_chars)
-    branch = ''.join(branch_chars)
-    name = separator.join([repo, branch, hash_str])
-
-    return name
+    repo = ''.join(repo_chars).strip('-')
+    branch = ''.join(branch_chars).strip('-')
+    return separator.join(filter(None, (repo, branch, hash_str)))
 
 
 def get_instance_token_file_name(instance):
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -10,6 +10,7 @@ import os
 import os.path
 import pytest
 import datetime
+import re
 import sys
 from time import tzset
 
@@ -21,6 +22,9 @@ from osbs.utils import (buildconfig_update,
                         get_instance_token_file_name, Labels)
 from osbs.exceptions import OsbsException
 import osbs.kerberos_ccache
+
+
+BC_NAME_REGEX = r'^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'
 
 
 def test_buildconfig_update():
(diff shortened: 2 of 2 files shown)


osbs/utils.py
```diff
@@ -327,20 +328,34 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
...
                 size += 1
                 if size >= limit:
                     break
         else:
             continue
         break
 
+    components = []
     repo = ''.join(repo_chars)
+    if repo:
```
rcerven: why the if? there would be always at least 1 valid character, I don't think that git would allow you name without a-z0-9 char

osbs/utils.py
```diff
@@ -327,20 +328,34 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
...
+        components.append(repo)
+
     branch = ''.join(branch_chars)
-    name = separator.join([repo, branch, hash_str])
+    if branch:
```
rcerven: why the if? there would be always at least 1 valid character, I don't think that git would allow you name without a-z0-9 char

osbs/utils.py
```diff
@@ -327,20 +328,34 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
 
     size = (2 * len(separator)) + hash_size
 
+    disallowed = '._'
```
lcarva: Can we do `allowed` instead to guarantee this is not a problem again?

rcerven: and if they would be having wrong chars in names it would fail? we are suppose to get rid of wrong chars to submit valid name

csomh: Agree. we now that `'-' + string.ascii_lowercase + string.digits` is allowed. Anything that is not in that should be removed or replaced with something.

osbs/utils.py
```diff
@@ -327,20 +328,34 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
...
     for i in range(max(len(repo), len(branch))):
         for group, group_chars in groups:
             if i < len(group):
-                group_chars.append(group[i])
+                char = group[i]
+                if char in disallowed:
```
lcarva: I think it might be easier to understand the code if `repo` and `branch` are sanitized right after hash is computed:

```python
valid_chars = re.compile('[0-9a-z]')
repo = filter(valid_chars.match, repo)
branch = filter(valid_chars.match, branch)
```

Here's a sample:

```python
>>> import re
>>> repo = 'hello.world-bah!'
>>> valid_chars = re.compile('[0-9a-z]')
>>> filter(valid_chars.match, repo)
'helloworldbah'
```

rcerven: +1

osbs/utils.py
lcarva: Given the problems we've had with this function, I think we should perform this check as a safety net. But perhaps we can use filter here as well :)

```python
>>> '-'.join(filter(lambda x: x, [None, '', 'spam', 'bacon']))
'spam-bacon'
```

osbs/utils.py
vrutkovs: +1 to a solution which uses just a list of allowed chars

osbs/utils.py
twaugh: You can create a branch called '___' if you want.

osbs/utils.py
twaugh: +1

osbs/utils.py
twaugh: +1

osbs/utils.py
twaugh: +1

osbs/utils.py
```diff
@@ -289,6 +289,10 @@ def get_time_from_rfc3339(rfc3339):
         return timegm(time_tuple)
 
 
+
+VALID_BUILD_CONFIG_NAME_CHARS = re.compile('[a-z0-9]')
```
csomh: Is it a good idea to remove every `-`? The only thing we should make sure, that filtered branch and repo names are not beginning or ending in `-`.

osbs/utils.py
```diff
@@ -321,6 +326,10 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
     shaval = sha256(full.encode('utf-8')).hexdigest()
     hash_str = shaval[:hash_size]
 
+    # Sanitize repo and branch names
+    repo = ''.join(filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(repo)))
+    branch = ''.join(filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(branch)))
```
csomh: I think `list()` can be omitted, strings are already iterable.

osbs/utils.py
```diff
@@ -340,9 +349,7 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
 
     repo = ''.join(repo_chars)
     branch = ''.join(branch_chars)
-    name = separator.join([repo, branch, hash_str])
-
-    return name
+    return separator.join(filter(None, (repo, branch, hash_str)))
```
csomh: I think `filter()` is superfluous here, all members to be joined should be 'clean' by this point.

osbs/utils.py
lcarva: I'd rather lean on the safe side on this.

osbs/utils.py
lcarva: That's what I thought! But python3 has its own mind...

In python 3, `filter` always returns a filter generator, which can only be converted to `str` via `str.join`. And in python 2, `filter` will either return a `list`, or a `str`, depending on the `typeof` input. To keep things working for both python2 and python3, convert the input to `list` to force `filter` in python2 to return a `list`. This way, both the `list` and `filter generator` can both work with `str.join`.

osbs/utils.py
lcarva: filter is needed to handle case where all of repo, or branch, contain invalid characters.

```python
>>> '-'.join(filter(None, [None, '', 'spam', 'bacon']))
'spam-bacon'
>>> '-'.join([None, '', 'spam', 'bacon'])
TypeError: sequence item 0: expected string, NoneType found
sequence item 0: expected string, NoneType found
>>> '-'.join(['', 'spam', 'bacon'])
'-spam-bacon'
```

osbs/utils.py
twaugh: I'm with @csomh, would be nice to retain readability.

lcarva: We need to strip both `-` and `.` from both ends of `repo` and `branch`. But this doesn't make me feel very confident that the issue is fixed for good....

lcarva: If we still want to allow dashes and dots:
1) remove non allowed characters from repo and branch. Anything other than `[\-\.a-z0-9]' - We still want this because underscores, for instance, are not allowed.
2) strip and rstrip `'.'` and `'-'` from repo and branch

Let's agree on the algorithm before any further code changes.

lcarva: BTW `strip` and `rstrip` is not sufficient. Need something like `re.sub` to ensure proper replacement when the repo or branch name ends in both `'-'` and `'.'`:

```python
>>> repo = '---...--..blah..---..--'
>>> re.sub(r'^(\.|\-)*', '', repo)
'blah..---..--'
>>> re.sub(r'(\.|\-)*$', '', repo)
'---...--..blah'
```

Definitely a corner case, but I'd rather have this fixed in a way that won't bite us again.

twaugh: Ok, maybe you were right to start with and we should just not use them at all.

osbs/utils.py
csomh: This would mean that repo is branch name has no letters or numbers at all, which I don't think would be allowed, or anyone would make something like this. But even if they did, the TypeError would be an appropriate answer :)

osbs/utils.py
csomh: Ok, I was not aware of the python3 behavior.

osbs/utils.py
csomh: Let's not complicate ourselves with `.`, and keep only `-`. So:

```
VALID_BUILD_CONFIG_NAME_CHARS = re.compile('[\-a-z0-9]')

repo = filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(repo))
repo = ''.join(repo).strip('-')

branch = filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(branch))
branch = ''.join(branch).strip('-')
```

csomh: Oh, striping actually should happen after the length of the members was adjusted, something like:
```python
VALID_BUILD_CONFIG_NAME_CHARS = re.compile('[\-a-z0-9]')
MEMBER_LIMIT = (limit - hash_size - 2) // 2

# ... calculate hash

repo = filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(repo))
repo = ''.join(repo)
repo = repo[:min(MEMBER_LIMIT, len(repo))]
repo = repo.strip('-')

branch = filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(branch))
branch = ''.join(branch)
branch = branch[:min(MEMBER_LIMIT, len(branch))]
branch = branch.strip('-')
```

twaugh: The benefit of not just allocating half the remaining length to each of the two components, and instead adding characters from each one by one until the length limit is reached, is that you get a more readable name in the case that one component of the name is over half the remaining length.

csomh: True. I did not consider this.

twaugh: Test-case added as demo ('longernamethan-this').

twaugh:
I'm pretty happy with how it looks at this point, LGTM.

twaugh:
Added another commit to try harder to include '-' for readability.

tests/test_utils.py
```diff
@@ -82,23 +83,34 @@ def test_get_time_from_rfc3339_valid(rfc3339, seconds, tz):
...
+    ('longer-name-than', 'this', 22, '-', 'longer-name-than-this'),
+    ('longer--name--than', 'this', 22, '-', 'longer--name--tha-this'),
 ])
 def test_make_name_from_git(repo, branch, limit, separator, expected, hash_size=5):
     bc_name = make_name_from_git(repo, branch, limit + len(separator) + hash_size, separator, hash_size=hash_size)
 
     assert expected == bc_name[:-(hash_size + len(separator))]
 
+    # Is this a valid name for OpenShift to use?
+    valid = re.compile(r'^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$')
```
lcarva: Let's apply this in `test_make_name_from_git_all_from_file` as well

twaugh:
Good idea, updated.

lcarva:
Verified against list of internal repos and branches, no errors detected.

PR #520, opened 2017-02-14
Create worker build

lcarva:


lcarva:
@twaugh, @vrutkovs, @rcerven PTAL

First take on the `create_worker_build` API method. Please note:
1. Unit tests are still in development and not part of this WIP PR - but existing tests are still passing
2. Added initial, relevant, support for `filesystem_koji_task_id` parameter, but can remove if we agree it's outside the scope of this change.
3. `worker_inner.json` is based off of `prod_inner.json` with some plugins removed and new ones added as `required=False` since they don't yet exist. I can remove reference to non-existent plugins if that's desired.
4. `worker_inner.json` also defines `buildstep_plugins` to use `docker_api`. This is based on the assumption that build step abstraction PR is merged. However, even without that PR, build still succeeds as the `buildstep_plugins` key in json is simply ignored (warning logged).

osbs/api.py
```diff
@@ -405,12 +418,19 @@ def create_prod_build(self, git_uri, git_ref,
         :param architecture: str, build architecture
         :param yum_repourls: list, URLs for yum repos
         :param koji_task_id: int, koji task ID requesting build
+        :param filesystem_koji_task_id: int, koji task for base image filesystem
         :param scratch: bool, this is a scratch build
+        :param platform: str, the platform to build for
```
vrutkovs: This should be the name of the target cluster in `osbs.conf`, right?

inputs/worker_inner.json
```diff
@@ -0,0 +1,158 @@
...
+  "prebuild_plugins": [
+    {
+      "name": "add_filesystem",
+      "args": {
+        "koji_hub": "{{KOJI_HUB}}",
+        "from_task_id": "{{FILESYSTEM_KOJI_TASK_ID}}"
+      }
+    },
+    {
+      "name": "select_platform",
```
twaugh: I don't think we should include 'select_platform'.

inputs/worker_inner.json
```diff
@@ -0,0 +1,158 @@
...
+    {
+      "name": "tag_and_push",
+      "args": {
+        "registries": {
+          "{{REGISTRY_URI}}": { "insecure": true }
+        }
+      }
+    },
+    {
+      "name": "pulp_init",
```
twaugh: I don't think we should include 'pulp_init'.

inputs/worker_inner.json
```diff
@@ -0,0 +1,158 @@
...
+        "load_exported_image": true,
+        "method": "gzip"
+      }
+    },
+    {
+      "name": "pulp_pull",
+      "args": {}
+    },
+    {
+      "name": "koji_upload",
```
twaugh: Should this be koji_promote?

With this pull request, I think we want to be able to create a single worker build and have it complete on its own. When we get to orchestration, later, we can decide whether to start off with supporting only a single worker build, or to support multiple worker builds (and use koji_upload in the worker builds).

osbs/api.py
twaugh: No, it should be the name of the platform, e.g. x86_64.

osbs/conf.py
```diff
@@ -296,6 +296,9 @@ def get_distribution_scope(self):
     def get_architecture(self):
         return self._get_value("arch", self.conf_section, "architecture")
 
+    def get_platform(self):
+        return self._get_value("platform", self.conf_section, "platform")
```
twaugh: I don't think we want to add a 'platform' configuration file keyword.

inputs/worker_inner.json
rcerven: worker will already know which platform from orchestrator right?
so that might be actually plugin for orchestrator?

osbs/api.py
vrutkovs: Its not clear why won't `architecture` be reused then. Platforms would not match architectures?

osbs/conf.py
rcerven: we already have architecture there why not use that?

tests/test_api.py
```diff
@@ -290,15 +371,60 @@ def test_get_build_api(self, osbs):
...
 
-    def test_get_build_request_api(self, osbs):
-        build = osbs.get_build_request()
+    @pytest.mark.parametrize('build_type', (
+        None,
+        'simple',
+        'prod',
+        'prod-without-koji'
+    ))
+    def test_get_build_request_api_build_type(self, osbs, build_type):
```
rcerven: can we add comment that we are just testing if providing deprecated 'build_type' won't break anything?

osbs/conf.py
lcarva: Added so we can test worker build from CLI, perhaps a `--platform` argument is more desirable?

tests/test_api.py
lcarva: Added. Thanks!

osbs/api.py
twaugh: Not necessarily (platform is more than the CPU architecture, it's also the host OS etc).

But more importantly 'architecture' is a label we definitely want to be set in the worker build, not in configuration, i.e. that label's value should ultimate come from querying the machine the build was run on.

We want 'platform' to come from the caller of this method, not from osbs.conf.

osbs/conf.py
twaugh: platform is more than CPU architecture. The names are likely to overlap but the concepts are different.

Yes, this would be better as a CLI argument.

lcarva: Added as CLI arg. Thanks!

inputs/worker_inner.json
lcarva: Switched to koji_promote.

inputs/worker_inner.json
lcarva: Removed.

inputs/worker_inner.json
lcarva: Removed.

lcarva:
@twaugh `bump_release` is not currently part of `worker_inner.json`, which means that it will only work with Dockerfiles that define the `release` label. Should this be the case? Or should `bump_release` be added for now so that it can be removed later when `release` value is handled properly by a different MR?

I think adding `bump_release` for now will get us closer to the goal of having a functional worker build.

lcarva:
Removed resource limits restrictions as that was misunderstanding on my part.

twaugh:
Agreed about bump_release.

osbs/api.py
```diff
@@ -502,6 +519,27 @@ def create_build(self, **kwargs):
...
+
+        Pass through method to create_prod_build with the following
+        modifications:
+            - platform param is required
+            - inner template set to worker_inner.json if not set
+            - outer template set to worker.json if not set
+
+        :return: BuildResponse instance
+        """
+        if not kwargs.get('platform'):
```
rcerven: Shouldn't architecture be mandatory for workers as well?

twaugh: No.

The 'architecture' keyword is a configuration setting to override the implicit value (taken from asking the running system, like 'uname -m') for the 'architecture' label in the image. It is nothing to do with controlling which node should build the image.

IMHO we should deprecate 'architecture' altogether in fact.

lcarva:
Added `bump_release`. `worker_inner.json` differs from `prod_inner.json` by not including any chain rebuild related plugins and by specifying the builstep_plugins.

inputs/worker.json
```diff
@@ -0,0 +1,40 @@
+{
+  "apiVersion": "v1",
+  "kind": "BuildConfig",
+  "metadata": {
+    "name": "{{NAME}}",
+    "labels": {
+      "is_autorebuild": "false"
```
twaugh: No need for this.

inputs/worker_inner.json
```diff
@@ -0,0 +1,155 @@
...
+      "args": {
+        "parent_registry": "{{REGISTRY_URI}}",
+        "parent_registry_insecure": true
+      }
+    },
+    {
+      "args": {
+        "hub": "{{KOJI_HUB}}"
+      },
+      "name": "bump_release"
```
twaugh: We should be consistent in the order of the "args" and "name" keys to make it easier to read.

inputs/worker_inner.json
```diff
@@ -0,0 +1,155 @@
...
+    },
+    {
+      "name": "distribution_scope"
+    }
+  ],
+  "buildstep_plugins": [
+    {
+      "name": "docker_api"
+    }
+  ],
```
twaugh: Think we want to specify "buildstep_plugins" or omit it and pick up the default? Not sure which I prefer.

inputs/worker_inner.json
```diff
@@ -0,0 +1,155 @@
...
+    },
+    {
+      "name": "koji_promote",
+      "args": {
+        "kojihub": "{{KOJI_HUB}}",
+        "url": "{{OPENSHIFT_URI}}",
+        "verify_ssl": false,
+        "blocksize": 10485760,
+        "koji_keytab": false,
+        "koji_principal": false
```
twaugh: Could/should any of these be omitted so they pick up the default?

osbs/constants.py
```diff
@@ -17,7 +17,9 @@
 DEFAULT_CONFIGURATION_FILE = "/etc/osbs.conf"
 DEFAULT_CONFIGURATION_SECTION = "default"
 DEFAULT_OUTER_TEMPLATE = "prod.json"
+WORKER_OUTER_TEMPLATE = "worker.json"
 DEFAULT_INNER_TEMPLATE = "prod_inner.json"
+WORKER_INNER_TEMPLATE = "worker_inner.json"
 DEFAULT_CUSTOMIZE_CONF = "prod_customize.json"
```
twaugh: What about the ability to customize worker_inner.json?

lcarva: Added

inputs/worker_inner.json
lcarva: It's copied from prod_inner.json.

inputs/worker_inner.json
lcarva: Changed all to list `"name"` first.

inputs/worker.json
lcarva: Removed.

inputs/worker_inner.json
lcarva: Which is not a good reason... but it seems that plugin doesn't provide sensible defaults, so I think it's best to leave them in.

```python
    def __init__(self, tasker, workflow, kojihub, url,
                 verify_ssl=True, use_auth=True,
                 koji_ssl_certs=None, koji_proxy_user=None,
                 koji_principal=None, koji_keytab=None,
                 metadata_only=False, blocksize=None,
                 target=None, poll_interval=5)
```

inputs/worker_inner.json
lcarva: Removed `buildstep_plugins` as this feature is not in yet, and it's actually not needed since it should gracefully fallback to `docker_api` plugin once implemented.

inputs/worker.json
```diff
@@ -0,0 +1,38 @@
+{
+  "apiVersion": "v1",
+  "kind": "BuildConfig",
+  "metadata": {
+    "name": "{{NAME}}",
+    "labels": {}
```
twaugh: Maybe remove this line? Not needed if there are no labels.

inputs/worker_inner.json
twaugh: OK.

inputs/worker.json
lcarva: Removed.

osbs/api.py
```diff
@@ -502,6 +525,31 @@ def create_build(self, **kwargs):
...
+            - release param is required
+            - inner template set to worker_inner.json if not set
+            - outer template set to worker.json if not set
+            - customize configuration set to worker_customize.json if not set
+
+        :return: BuildResponse instance
+        """
+        for required in ('platform', 'release'):
+            if not kwargs.get(required):
+                raise ValueError('Worker build requires %s param' % required)
```
rcerven: this will raise the 1st error only, it would be better to report all problems and raise after

lcarva: True, but considering that this is done way before submitting a build, the user feedback is immediate, and it doesn't create unnecessary builds. I think if we add another required arg, then it's beneficial to check for all errors.

PR #523, opened 2017-02-28
client config and tokens secrets

rcerven:


osbs/conf.py
```diff
@@ -426,3 +426,18 @@ def get_oauth2_token(self):
...
+                               "client_config_secret")
+
+    def get_token_secrets(self):
+        value = self._get_value("token_secrets", self.conf_section,
+                                "token_secrets")
+        token_dict = {}
+        if value:
+            for pairs in value.split():
+                pair = pairs.split(':',1)
+                token_dict[pair[0]] = pair[1] if len(pair)>1 else ''
```
rcerven: should there be allowed more than 1 ':' ?

yeah I can raise exception when it is obviously wrong

osbs/build/build_request.py
```diff
@@ -867,7 +889,13 @@ def render(self, validate=True):
                            # add the path to the plugin's
                            # configuration. This is done elsewhere.
                            None):
-                          self.spec.registry_secrets.value})
+                          self.spec.registry_secrets.value,
+
+                          ('buildstep_plugins', 'orchestrate_build', 'osbs_client_config'):
```
rcerven: just path in this case makes more sense, and defaulting to 'osbs.conf'

osbs/conf.py
twaugh: Only one `:` makes sense.

osbs/build/build_request.py
lcarva: Alternatively, if it's just mounted on `/etc/osbs.conf` then there's no configuration needed. The current approach definitely provides more flexibility, but I'm not sure it's needed at all. Unless of course, mounting a secret that contains a file `osbs.conf` on `/etc` is problematic.

lcarva: Also, can we omit `orchestrate_plugin` from this PR? Currently, this PR shouldn't be merged until the plugin is finalized and merged. This dependency seems unnecessary.

rcerven: there isn't dependency, if the plugin won't be configured it will be just skipped

twaugh: @lcarva, let's not mount over /etc. Don't forget, we also use dockpulp.conf when Pulp integration is enabled, as well as /etc/yum.repos.d/ which we'd have to recreate.

twaugh: @lcarva, IMHO it's fine to merge this PR before the plugin exists; as @rcerven said, if the plugin is not in the build request (because it doesn't exist yet) it won't be configured.

The only dependency is exactly what the parameter name is, and that's something we can just agree on before merge.

osbs/build/build_request.py
```diff
@@ -260,31 +260,41 @@ def render_unique_tag_only(self):
...
+        if plugin is not None:
+            has_plugin_conf = self.dj.dock_json_has_plugin_conf(plugin[0],
+                                                                plugin[1])
         if 'secrets' in self.template['spec']['strategy']['customStrategy']:
-            if has_plugin_conf:
-                secret_path = os.path.join(SECRETS_PATH, secret)
-                logger.info("Configuring %s secret at %s", secret, secret_path)
+            if not plugin or has_plugin_conf:
+                if not isinstance(secrets, dict):
+                    secrets = {secrets : ""}
```
twaugh: I'm not sure I like parameters that get automatically changed to a different type.

How about making the method signature `set_secret_for_plugin(self, secret, plugin=None, mount_path=None)`? Then iterating over the secrets in `token_secrets` could be done in render() instead of here.

osbs/conf.py
```diff
@@ -426,3 +427,34 @@ def get_oauth2_token(self):
...
+        will_raise = False
+        if value:
+            for pairs in value.split():
+                if pairs.count(':') >= 1:
+                    logger.error("token_secret can't contain multiple colons : %s", pairs)
+                    will_raise = True
+                    continue
+                pair = pairs.split(':')
+
+                if len(pair) > 1:
```
twaugh: What's this test for? It seems to be what causes 'foo:x' and 'foo:' to be treated like 'foo'.

osbs/conf.py
```diff
@@ -426,3 +427,34 @@ def get_oauth2_token(self):
...
+                    if pair[1] in ("/"):
+                        logger.error("token_secret file path must be valid : %s", pair[1])
+                        will_raise = True
+                        continue
+                    token_dict[pair[0]] = pair[1]
+
+                else:
+                    token_dict[pair[0]] = ''
```
twaugh: Think I slightly prefer None as the default here. it's the mount path to use, or it's None.

osbs/conf.py
```diff
@@ -426,3 +427,34 @@ def get_oauth2_token(self):
...
+
+        if will_raise:
+            raise OsbsException("Wrong token_secrets configuration")
```
twaugh: Elsewhere we use `OsbsValidationException` for config errors I think.

osbs/conf.py
rcerven: following check will chech if it is empty or / and will raise exception

osbs/build/build_request.py
vrutkovs: @twaugh this can be handled by symlink from `/etc/osbs.conf` to a dir where the secret is mounted. I'd prefer to have a minimum configuration if possible

osbs/conf.py
```diff
@@ -426,3 +427,34 @@ def get_oauth2_token(self):
...
+
+    def get_token_secrets(self):
+        value = self._get_value("token_secrets", self.conf_section,
+                                "token_secrets")
+        token_dict = {}
+        will_raise = False
+        if value:
+            for pairs in value.split():
+                if pairs.count(':') >= 1:
+                    logger.error("token_secret can't contain multiple colons : %s", pairs)
```
vrutkovs: Nitpick: extra space before the semicolon

osbs/conf.py
```diff
@@ -426,3 +427,34 @@ def get_oauth2_token(self):
...
+                    will_raise = True
+                    continue
+                pair = pairs.split(':')
+
+                if len(pair) > 1:
+                    if pair[1] in ("/"):
+                        logger.error("token_secret file path must be valid : %s", pair[1])
```
vrutkovs: Nitpick: extra space before the semicolon

osbs/build/build_request.py
twaugh: What would put the symlink there? It would have to be baked into the builder image? In that case there would need to be a "well-known" mount path for the osbs.conf secret (and a separate one for the dockpulp config if we use this method for the Pulp integration).

We could set a default, sure... but isn't that more work than just making it configurable (which it would have to be anyway)?

osbs/conf.py
rcerven: do you mean extra space before colon in that logger message?

osbs/conf.py
rcerven: ok

osbs/conf.py
rcerven: sure

osbs/conf.py
vrutkovs: Yeah, I meant the colon in logger message

osbs/conf.py
```diff
@@ -426,3 +427,34 @@ def get_oauth2_token(self):
...
+                                "token_secrets")
+        token_dict = {}
+        will_raise = False
+        if value:
+            for pairs in value.split():
+                if pairs.count(':') >= 1:
+                    logger.error("token_secret can't contain multiple colons: %s", pairs)
+                    will_raise = True
+                    continue
+                pair = pairs.split(':')
```
twaugh: Nicer if we can give these names, e.g. `secret, mount_path = pairs.split(':')`. Also I don't think we need to check whether ':' is in the mount_path as that can still be a valid path.

osbs/conf.py
twaugh: Ah, I misread this.

tests/build/test_build_request.py
```diff
@@ -1538,3 +1538,53 @@ def test_reactor_config(self, secret):
...
+
+        if secret is not None:
+            assert get_secret_mountpath_by_name(build_json, 'osbsconf') == os.path.join(SECRETS_PATH, secret)
+            assert get_plugin(plugins, plugin_type, plugin_name)
+            print(plugins)
+            assert plugin_value_get(plugins, plugin_type, plugin_name,
+                                    'args', 'osbs_client_config') == os.path.join(SECRETS_PATH, secret)
+        else:
+            with pytest.raises(AssertionError):
+                assert get_secret_mountpath_by_name(build_json, 'osbsconf')
```
twaugh: Could these two lines by replaced by this?:
```
assert not get_secret_mountpath_by_name(build_json, secret)
```

tests/build/test_build_request.py
```diff
@@ -1538,3 +1538,53 @@ def test_reactor_config(self, secret):
...
+        kwargs['client_config_secret'] = secret
+        br.set_params(**kwargs)
+
+        br.dj.dock_json_set_param(plugin_type, [])
+        br.dj.add_plugin(plugin_type, plugin_name, plugin_args)
+        build_json = br.render()
+        plugins = get_plugins_from_build_json(build_json)
+
+        if secret is not None:
+            assert get_secret_mountpath_by_name(build_json, 'osbsconf') == os.path.join(SECRETS_PATH, secret)
```
twaugh: Should we use `secret` instead of `'osbsconf'` here?

tests/test_conf.py
```diff
@@ -246,6 +260,41 @@ def test_param_retrieval(self, config, kwargs, cli_args, expected):
...
+         {'secret': 'path:with:colons'}),
+
+        ({'default': {'token_secrets': 'secret:path secret2:path2'}},
+         {'secret': 'path', 'secret2': 'path2'}),
+
+        ({'default': {'token_secrets': 'secret:path secret2:path2 secret3:path3'}},
+         {'secret': 'path', 'secret2': 'path2', 'secret3': 'path3'}),
+
+        ({'default': {'token_secrets': 'secret:path secret2 secret3:path3'}},
+         {'secret': 'path', 'secret2': None, 'secret3': 'path3'}),
```
twaugh: Could we include a test for other whitespace? e.g.
```
({'default': {'token_secrets': '\n   foo:/bar\n   bar:/baz'}},
 {'foo': '/bar', 'bar': '/baz'}),
```

osbs/conf.py
```diff
@@ -426,3 +427,30 @@ def get_oauth2_token(self):
...
+        value = self._get_value("token_secrets", self.conf_section,
+                                "token_secrets")
+        token_dict = {}
+        will_raise = False
+        if value:
+            for pairs in value.split():
+                pair = pairs.split(':', 1)
+
+                if len(pair) == 2:
+                    if pair[1] in "/":
```
rcerven: when using 'in' it will also parse None
so it covers '/' and 'None' cases

osbs/build/build_request.py
```diff
@@ -260,18 +260,25 @@ def render_unique_tag_only(self):
 
         self.dj.remove_plugin('postbuild_plugins', 'tag_from_config')
 
-    def set_secret_for_plugin(self, plugin, secret):
-        has_plugin_conf = self.dj.dock_json_has_plugin_conf(plugin[0],
-                                                            plugin[1])
+    def set_secret_for_plugin(self, secret, plugin=None, mount_path=None):
```
rcerven: sure

osbs/build/build_request.py
```diff
@@ -867,7 +874,14 @@ def render(self, validate=True):
                            # add the path to the plugin's
                            # configuration. This is done elsewhere.
                            None):
-                          self.spec.registry_secrets.value})
+                          self.spec.registry_secrets.value,
+
+                          ('buildstep_plugins', 'orchestrate_build', 'osbs_client_config'):
+                          self.spec.client_config_secret.value})
+
+        if self.spec.token_secrets.value:
```
rcerven: ahh right

tests/build/test_build_request.py
rcerven: we can

tests/build/test_build_request.py
```diff
@@ -1538,3 +1538,53 @@ def test_reactor_config(self, secret):
...
+            assert get_plugin(plugins, plugin_type, plugin_name)
+            print(plugins)
```
rcerven: right debug ;)

tests/build/test_build_request.py
rcerven: get_secret_mountpath_by_name will raise exception so we need that raise check

tests/test_conf.py
rcerven: added

osbs/build/build_request.py
lcarva: Ok, I think enabling the plugin to take a param for a custom location for osbs.conf is a good idea. It'll also help facilitate unit testing. 

My only concern is to have this being done by this PR. But it's not a major concern. I was hoping to fully configure that plugin in its own PR. Let's leave it if others don't have a problem with it.

osbs/build/build_request.py
```diff
@@ -260,18 +260,33 @@ def render_unique_tag_only(self):
...
-    def set_secret_for_plugin(self, plugin, secret):
-        has_plugin_conf = self.dj.dock_json_has_plugin_conf(plugin[0],
-                                                            plugin[1])
+    def set_secret_for_plugin(self, secret, plugin=None, mount_path=None):
+        """
+        Sets secret for plugin, if no plugin specified
+        it will also set general secret
+
+        :param secret: str, secret name
+        :param pugin: touple, (plugin type, plugin name, argument name)
```
twaugh: typos: pugin -> plugin, touple -> tuple

osbs/conf.py
twaugh: That's pretty non-obvious I think. I'd prefer to make this explicit, e.g. `if pair[1] in ['', '/']:`.

tests/build/test_build_request.py
```diff
@@ -1538,3 +1538,52 @@ def test_reactor_config(self, secret):
...
+        }
+    ])
+    def test_token_secrets(self, secret):
+        br = BuildRequest(INPUTS_PATH)
+        kwargs = get_sample_prod_params()
+        kwargs['token_secrets'] = secret
+        br.set_params(**kwargs)
+        build_json = br.render()
+
+        for (sec, pat) in secret.items():
```
twaugh: Nit-pick: I'd rather name the second one `path`.

osbs/build/build_request.py
vrutkovs: Right, the buildroot image would have to have a symlink to a specific location - that might get very unreliable. Lets use dir path

tests/build/test_build_request.py
```diff
@@ -1538,3 +1538,52 @@ def test_reactor_config(self, secret):
...
+        plugins = get_plugins_from_build_json(build_json)
+
+        if secret is not None:
+            assert get_secret_mountpath_by_name(build_json, secret) == os.path.join(SECRETS_PATH, secret)
+            assert get_plugin(plugins, plugin_type, plugin_name)
+            assert plugin_value_get(plugins, plugin_type, plugin_name,
+                                    'args', 'osbs_client_config') == os.path.join(SECRETS_PATH, secret)
+        else:
+            with pytest.raises(AssertionError):
+                assert get_secret_mountpath_by_name(build_json, secret)
```
twaugh: We don't want the `assert` on this line.

tests/test_conf.py
```diff
@@ -246,6 +260,47 @@ def test_param_retrieval(self, config, kwargs, cli_args, expected):
...
+
+        ({'default': {'token_secrets': '\t\n   secret:path   \t\t  secret2\n\n \tsecret3:path3 \n\t\n'}},
+         {'secret': 'path', 'secret2': None, 'secret3': 'path3'}),
+    ])
+    def test_get_token_secrets(self, config, expected):
+        with self.config_file(config) as config_file:
+            conf = Configuration(conf_file=config_file)
+
+            if expected is not Exception:
```
vrutkovs: We should be expecting `OsbsValidationException` here - this test would make sure no unclear exceptions are thrown

osbs/conf.py
rcerven: ok

osbs/build/build_request.py
rcerven: fixed

vrutkovs: Typos: `plugin` and `tuple`

tests/build/test_build_request.py
rcerven: right we don't need assert at all here

tests/build/test_build_request.py
rcerven: done

tests/test_conf.py
rcerven: ok

osbs/build/build_request.py
rcerven: fixed

PR #524, opened 2017-03-01
Rename tests/build so that py.test spots tests there

twaugh:
It seems like Travis CI defaults to norecursedirs=build somewhere, but only for Python 3.5. This means all the tests in tests/build were getting ignored for the Travis CI Python 3.5 runs.


rcerven:
+1

PR #525, opened 2017-03-03
Coordinate plugins for orchestrator/worker builds

twaugh:
This commit introduces the concept of an 'arrangement version'. For orchestrator builds, the plugins run in the orchestrator build and those run in the worker build must be coordinated.  Changes to this arrangement will be made by defining a new arrangement version and creating new 'inner' JSON templates for the orchestrator and worker builds. The DEFAULT_ARRANGEMENT_VERSION constant will be incremented to match this new version.

The osbs.conf configuration file now has an arrangement_version keyword, which defaults to DEFAULT_ARRANGEMENT_VERSION.

The create_orchestrator_build() method takes an optional keyword parameter 'arrangement_version', which defaults to the value provided from configuration. This is used to select the orchestrator_inner:n.json template to use, as well as being provided to the orchestrate_build plugin as part of its build_kwargs parameter.

The create_worker_build() method has a new required 'arrangement_version' parameter. This is used to select the worker_inner:n.json template to use.

When a requested arrangement version is not present (the inner template file does not exist) for either an orchestrator or worker build, OsbsValidationException is raised.

This is made possible by avoiding nested calls to @osbsapi-wrapped methods. The create_prod_build method is renamed to _do_create_prod_build (which is not @osbsapi-wrapped), and create_*_build methods all use this.


lcarva:
@twaugh, I'm assuming the idea is that `create_orchestrator_build` will also perform the same operation?

If so, do we need a check to ensure they are the same arrangement version, or simply reading from the same place (`self.build_conf.get_arrangement_version()`) is sufficient?

twaugh:
Once we have the create_orchestrator_build() API in place, that method will provide `arrangement_version` as a parameter to the orchestrate_build plugin. When it calls create_worker_build() it will pass that parameter on.

In that way, the worker build will use the same arrangement_version that was used to create the orchestrator build.

Once all this is in place, I actually think it would be good to make create_worker_build() *require* `arrangement_version`.

lcarva:
Can we make it required now? I'd rather have this api method well defined as early as possible.
Perhaps adjust the CLI code accordingly?

twaugh:
Made a required parameter. Now there's no way for it to have a default I'll leave the configuration parameter until another PR.

osbs/api.py
```diff
@@ -536,17 +536,23 @@ def create_worker_build(self, *args, **kwargs):
...
             - outer template set to worker.json if not set
             - customize configuration set to worker_customize.json if not set
 
         :return: BuildResponse instance
+
         """
-        for required in ('platform', 'release'):
-            if not kwargs.get(required):
+        for required in ('platform', 'release', 'arrangement_version'):
+            if kwargs.get(required) is None:
```
lcarva: We still want to make sure that `platform` and `release` are truthy.

twaugh: Updated. No test update because the test was broken (see https://github.com/projectatomic/osbs-client/pull/528).

twaugh: I've reimplemented this differently now. Rather than remember whether each param must be truthy, let's start arrangement_version at 1 so they all must be truthy.

twaugh:
It's not an API version really: the API remains the same regardless of which arrangement is used (it's the implementation that changes).

I think 'version' on its own is not descriptive enough.

twaugh:
I realised it should be passing the arrangement version through to the orchestrate_build plugin using the build_kwargs parameter rather than in a separate parameter. Updated.

This also means there will not be an atomic-reactor pull request required for this.

osbs/api.py
```diff
@@ -557,21 +564,32 @@ def create_worker_build(self, *args, **kwargs):
...
+            - inner template set to worker_inner:n.json if not set
             - outer template set to worker.json if not set
             - customize configuration set to worker_customize.json if not set
 
         :return: BuildResponse instance
         """
-        for required in ('platform', 'release'):
+        for required in ('platform', 'release', 'arrangement_version'):
             if not kwargs.get(required):
                 raise ValueError('Worker build requires %s param' % required)
```
rcerven: maybe we can first loop and report all errors and raise after?

osbs/api.py
```diff
@@ -581,6 +604,8 @@ def create_orchestrator_build(self, *args, **kwargs):
         Pass through method to create_prod_build with the following
         modifications:
             - platforms param is required
+            - arrangement_version param may be used to select which
+              orchestrator_inner:n.json template to use
             - inner template set to orchestrator_inner.json if not set
```
lcarva: inner template set to orchestrator_inner:n.json if not set

osbs/api.py
```diff
@@ -590,11 +615,25 @@ def create_orchestrator_build(self, *args, **kwargs):
         if not kwargs.get('platforms'):
             raise ValueError('Orchestrator build requires platforms param')
 
-        kwargs.setdefault('inner_template', ORCHESTRATOR_INNER_TEMPLATE)
+        try:
+            arrangement_version = kwargs['arrangement_version']
```
lcarva: Simpler?

```python
arrangement_version = kwargs.setdefault('arrangement_version', 
                                        self.build_conf.get_arrangement_version())
```

twaugh:
Updated.

PR #526, opened 2017-03-03
Describe new client_config_secret and token_secrets options

twaugh:
This was missed from https://github.com/projectatomic/osbs-client/pull/523 by mistake.


twaugh:
@rcerven, might be nice to pull this in before releasing the new version of osbs-client.

rcerven:
+1

PR #527, opened 2017-03-03
add create_orchestrator_build api method

lcarva:
Requires https://github.com/projectatomic/atomic-reactor/pull/656


osbs/api.py
```diff
@@ -553,6 +558,29 @@ def create_worker_build(self, *args, **kwargs):
...
+        :return: BuildResponse instance
+        """
+        if not kwargs.get('platforms'):
+            raise ValueError('Orchestrator build requires platforms param')
+
+        kwargs.setdefault('inner_template', ORCHESTRATOR_INNER_TEMPLATE)
+        kwargs.setdefault('outer_template', ORCHESTRATOR_OUTER_TEMPLATE)
+        kwargs.setdefault('customize_conf', ORCHESTRATOR_CUSTOMIZE_CONF)
+
+        return self.create_prod_build(*args, **kwargs)
```
vrutkovs: This would return a single `BuildReponse`. Should koji-containerbuild handle it in some special way?

lcarva: No, it should be unaware of the underlying implementation. Tracking the OpenShift build in the returned build_response is sufficient.

vrutkovs: Awesome, thanks

osbs/api.py
```diff
@@ -553,6 +558,29 @@ def create_worker_build(self, *args, **kwargs):
...
+        Pass through method to create_prod_build with the following
+        modifications:
+            - platforms param is required
+            - inner template set to orchestrator_inner.json if not set
+            - outer template set to orchestrator.json if not set
+            - customize configuration set to orchestrator_customize.json if not set
+
+        :return: BuildResponse instance
+        """
+        if not kwargs.get('platforms'):
```
twaugh: Do we consider `[]` to be a valid `platforms` value?

osbs/build/build_request.py
```diff
@@ -183,6 +183,39 @@ def render_reactor_config(self):
...
+
+        if self.spec.platforms.value is None:
+            logger.debug('removing {0} plugin: no platforms'.format(plugin))
+            self.dj.remove_plugin(phase, plugin)
+            return
+
+        if self.spec.client_config_secret.value is None:
+            logger.debug('removing {0} plugin: no secret'.format(plugin))
+            self.dj.remove_plugin(phase, plugin)
+            return
```
twaugh: Should this be fatal, or should the orchestrate_build plugin not provide a path to the `Configuration` instance it creates?

osbs/cli/main.py
```diff
@@ -600,12 +603,19 @@ def cli():
                               help="perform a scratch build")
     build_parser.add_argument("--yum-proxy", action='store', required=False,
                               help="set yum proxy to repos from koji/add-yum-repo params")
-    build_parser.add_argument("--worker", action="store_true", required=False,
-                              default=False, help="create worker build")
     build_parser.add_argument("--platform", action='store', required=False,
-                              help="platform name to use")
+                              help="platform name to use (worker build)")
+    build_parser.add_argument("--platforms", action='append', metavar='PLATFORM',
+                              help="name of each platform to use (orchestrator build)")
```
twaugh: Is there any validation to prevent `--worker --platforms ...` or `--orchestrator --platform ...`?

osbs/api.py
twaugh: I see you explicitly listed it as a failing value in the tests.

Given that for multi-cluster there would be no Koji build created, and for multi-arch it would try to create an empty build, it's best to catch this error early like this.

osbs/build/build_request.py
```diff
@@ -183,6 +183,39 @@ def render_reactor_config(self):
...
             self.dj.remove_plugin('prebuild_plugins', 'reactor_config')
 
+    def render_orchestrate_build(self):
+        phase = 'buildstep_plugins'
+        plugin = 'orchestrate_build'
+        if not self.dj.dock_json_has_plugin_conf(phase, plugin):
+            return
+
+        if self.spec.platforms.value is None:
+            logger.debug('removing {0} plugin: no platforms'.format(plugin))
```
twaugh: landscape.io prefers `%` and separate arguments.

tests/test_api.py
```diff
@@ -205,6 +207,62 @@ class MockParser(object):
...
+        (None, None, ORCHESTRATOR_CUSTOMIZE_CONF),
+        (None, None, None),
+    ))
+    @pytest.mark.parametrize(('platforms', 'raises_exception'), (
+        (None, True),
+        ([], True),
+        (['spam'], True),
+        (['spam', 'bacon'], True),
+    ))
+    def test_create_worker_build(self, osbs, inner_template, outer_template,
```
twaugh: Typo in method name. :-)

tests/test_api.py
```diff
@@ -205,6 +207,62 @@ class MockParser(object):
...
+        if outer_template is not None:
+            kwargs['outer_template'] = outer_template
+        if customize_conf is not None:
+            kwargs['customize_conf'] = customize_conf
+
+        expected_kwargs = {
+            'platforms': platforms,
+            'inner_template': WORKER_INNER_TEMPLATE,
+            'outer_template': WORKER_OUTER_TEMPLATE,
+            'customize_conf': WORKER_CUSTOMIZE_CONF,
```
twaugh: Are these values right?

tests/test_api.py
```diff
@@ -205,6 +207,62 @@ class MockParser(object):
...
+        }
+
+        (flexmock(osbs)
+            .should_call('create_prod_build')
+            .with_args(*args, **expected_kwargs)
+            .times(0 if raises_exception else 1))
+
+        if raises_exception:
+            with pytest.raises(OsbsException):
+                osbs.create_worker_build(*args, **kwargs)
```
twaugh: Shouldn't this be `create_orchestrator_build`?

tests/test_api.py
```diff
@@ -205,6 +207,62 @@ class MockParser(object):
...
+        else:
+            response = osbs.create_worker_build(*args, **kwargs)
```
twaugh: Shouldn't this be `create_orchestrator_build`?

osbs/build/build_request.py
lcarva: It's a required param in the plugin. But that's a good point. It'd be more flexible if it gracefully defaults to whatever `Configuration` uses. I'll change the plugin to make it optional.

osbs/api.py
lcarva: I don't think it should be. The `orchestrate_build` plugin will essentially be a no op in such case. I don't really see a valid uses case for `[]`.

osbs/cli/main.py
lcarva: Unfortunately, `argparse` doesn't provide something like this. I created some helper functions to do so. Let me know what you think.

I also added [groups](https://docs.python.org/2.7/library/argparse.html#argument-groups) for worker and orchestrator. I found it might be useful when using `--help` .

```
$ osbs build --help
usage: osbs build ...

optional arguments:
  -h, --help            show this help message and exit
...

arguments for --worker:
  Required arguments for creating a worker build

  --platform PLATFORM   platform name to use
  --release RELEASE     release value to use

arguments for --orchestrator:
  Required arguments for creating an orchestrator build

  --platforms PLATFORM  name of each platform to use
```

osbs/build/build_request.py
lcarva: As it should... fixed it.

tests/test_api.py
lcarva: Fixed this and the other references to `worker` in this test.

inputs/orchestrator_inner.json
```diff
@@ -0,0 +1,37 @@
...
+      "args": {
+        "platforms": "{{PLATFORMS}}",
+        "build_kwargs": "{{BUILD_KWARGS}}",
+        "osbs_client_config": "{{OSBS_CLIENT_CONFIG}}"
+      }
+    }
+  ],
+  "postbuild_plugins": [],
+  "prepublish_plugins": [],
+  "exit_plugins": []
```
twaugh: If we pull the parent image we'll also need to run `remove_built_image` on exit to remove it.

lcarva: Done

osbs/cli/main.py
```diff
@@ -44,6 +44,31 @@ def print_json_nicely(decoded_json):
     print(json.dumps(decoded_json, indent=2))
 
 
+def check_provided_args(args, check, check_func, error_msg):
+    """Checks arguments against a check function
+
+    :param args: argparse.Namespace, all provided arguments
+    :param check: list<str>, each argument to check
+    :param check_func: filter function for values
```
twaugh: Can this say 'items' instead of values, to make it clear why `[-1]` is necessary later?

lcarva: Good point, changed.

twaugh:
Ready for merge?

lcarva:
Yes, I have no pending changes.

PR #528, opened 2017-03-04
Aid testing by not wrapping flexmock expectation exceptions

twaugh:
Usually API methods will wrap any exceptions in OsbsException. However, if the exception is e.g. a MethodCallError, perhaps because a function expected never to be called has been called, this may cause a test to pass and mask a bug.

Don't wrap MethodCallError or other flexmock expectation exceptions.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -41,6 +41,10 @@ def osbsapi(func):
             # Re-raise OsbsExceptions
             raise
         except Exception as ex:
+            # Propogate flexmock errors immediately (used in test cases)
+            if getattr(ex, '__module__', None) == 'flexmock':
+                raise
+
             # Convert anything else to OsbsException
 
             # Python 3 has implicit exception chaining and enhanced
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -7,7 +7,7 @@ of the BSD license. See the LICENSE file for details.
 """
 from types import GeneratorType
 
-from flexmock import flexmock
+from flexmock import flexmock, MethodCallError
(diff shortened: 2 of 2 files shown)


osbs/api.py
```diff
@@ -40,6 +49,9 @@ def catch_exceptions(*args, **kwargs):
         except OsbsException:
             # Re-raise OsbsExceptions
             raise
+        except (MethodCallError, MethodSignatureError, CallOrderError):
```
lcarva: Instead of actually importing these, can we leverage their `__module__` attribute, and check it in the `Exception` block below?

```python
>>> MethodSignatureError.__module__
'flexmock'
>>> CallOrderError.__module__
'flexmock'
>>> MethodCallError.__module__
'flexmock'
```

twaugh: Great idea! I tried it and it works fine. Updated.

twaugh:
Added test case.

osbs/api.py
```diff
@@ -41,6 +41,11 @@ def catch_exceptions(*args, **kwargs):
             # Re-raise OsbsExceptions
             raise
         except Exception as ex:
+            # Propogate flexmock errors immediately (used in test cases)
+            if (hasattr(ex, '__module__') and
+                    getattr(ex, '__module__') == 'flexmock'):
```
lcarva: If you pass a third param to `getattr`, it's used as the default value instead of raising an exception. Might make a little bit cleaner.

twaugh: Updated.

PR #529, opened 2017-03-05
pulp_pull: set insecure pulp_pull plugin

vrutkovs:
Related to https://github.com/projectatomic/atomic-reactor/pull/660


vrutkovs:
We don't need this anymore - it doesn't change any behavior and its specified in documentation

PR #532, opened 2017-03-08
Fixes to coverage for the test suite itself

twaugh:
Some v1-only cases were not fully tested.

The no-arg-but-raises-exception cases was not covered. Fix this by calling a function that requires an argument and expecting TypeError.


lcarva:
Not sure why, but this actually decreased the code coverage.

lcarva:
Probably because the coveralls jobs didn't take into account both python 2 travis builds. Maybe re-trigger it?

twaugh:
Seems ok now.

PR #534, opened 2017-03-13
Store annotations for orchestrator build

lcarva:


osbs/cli/main.py
```diff
@@ -190,6 +190,48 @@ def cmd_list_builds(args, osbs):
...
+        try:
+            digests_str = "\n".join(["{registry}/{repository}:{tag} {digest}".format(**dig)
+                                     for dig in digests])
+        except (TypeError, KeyError):
+            digests_str = "(invalid value)"
+    else:
+        digests_str = "(unset)"
+
+    if not digests_str:
+        digests_str = "(empty)"
```
vrutkovs: This won't be reached, lets remove it

inputs/orchestrator_inner:1.json
```diff
@@ -35,6 +35,13 @@
   "prepublish_plugins": [],
   "exit_plugins": [
     {
+      "name": "store_metadata_in_osv3",
+      "args": {
+        "url": "{{OPENSHIFT_URI}}",
+        "verify_ssl": false
+      }
+    },
+    {
```
twaugh: This actually causes a test coverage decrease :-)
https://github.com/projectatomic/osbs-client/blob/3ee8276bfaafafe621d70dd9fb8a6a56498a0599/osbs/build/build_request.py#L273

osbs/cli/main.py
lcarva: It's reached if `digests` is an empty list, which is the case of an unsuccessful build, for example. See unit tests.

osbs/cli/main.py
```diff
@@ -290,6 +324,19 @@ def cmd_get_build(args, osbs):
             "koji_build_id": build.get_koji_build_id() or '(unset)',
             "digests": digests_str,
         }
+
+        worker_builds = json.loads(build.get_annotations().get('worker-builds', '{}'))
+        worker_builds_str = make_worker_builds_str(worker_builds)
```
lcarva: Thanks! Added.

inputs/orchestrator_inner:1.json
lcarva: Added a separate commit for adding test coverage to ensure plugins are optional.

osbs/cli/main.py
vrutkovs: Oh, right, `'\n'.join([])` return empty string.

PR #536, opened 2017-03-15
Optional git_branch param for builds

lcarva:


osbs/api.py
```diff
@@ -622,8 +623,9 @@ def create_orchestrator_build(self, *args, **kwargs):
         kwargs.setdefault('outer_template', ORCHESTRATOR_OUTER_TEMPLATE)
         kwargs.setdefault('customize_conf', ORCHESTRATOR_CUSTOMIZE_CONF)
 
+        kwargs.setdefault('git_branch', None)
```
twaugh: So we're defaulting to `None`? I thought it needed to be defaulted to something like `unknown`, or did I misunderstand the problem?

lcarva: It'll default to `'unknown'` if it's `None`. The problem was that the param `git_branch` wasn't provided at all, causing a python `TypeError` when calling that method.

twaugh: OK, got it.

PR #538, opened 2017-03-21
if build_image is used, it is also passed to worker build

rcerven:

### DIFF
--- a/tests/build_/test_build_request.py
+++ b/tests/build_/test_build_request.py
@@ -1044,8 +1044,15 @@ class TestBuildRequest(object):
         # Only one version defined so far
         DEFAULT_ARRANGEMENT_VERSION,
     ])
+    @pytest.mark.parametrize(('build_image', 'build_imagestream', 'worker_build_image'), (
+        ('fedora:latest', None, 'fedora:latest'),
+        (None, 'buildroot-stream:v1.0', KeyError),
+        (None, None, DEFAULT_BUILD_IMAGE),
+        ('fedora:latest', 'buildroot-stream:v1.0', KeyError)
+    ))
     def test_render_orchestrate_build(self, platforms, secret, disabled,
-                                      arrangement_version):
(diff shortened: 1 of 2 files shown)


lcarva:
Looks great! Let's make sure this is unit tested.

twaugh:
Are we happy that if neither build_image or build_imagestream is specified in the osbs.conf or in kwargs to osbs-client, that the plugin parameter will get set to `DEFAULT_BUILDROOT_IMAGE` (which is "buildroot:latest"? I wonder about this, because this will override any build_image set in the osbs.conf used by the orchestrator build to create the worker build.

twaugh:
So actually the problem I mentioned above comes about only because we have a `DEFAULT_BUILDROOT_IMAGE`. Some time we'll remove that.

The code you have looks great once tests are in place.

tests/build_/test_build_request.py
```diff
@@ -1044,8 +1044,17 @@ def test_render_optional_plugins(self, tmpdir):
         # Only one version defined so far
         DEFAULT_ARRANGEMENT_VERSION,
     ])
+    @pytest.mark.parametrize('build_image', [
```
lcarva: I think this should be more explicit on what the expected behavior is. 
This unit test is re-implementing logic that is being tested.
I think a stronger test would explicitly specify what the worker_build_image value should be depending on the values of `build_image` and `build_imagestream`.
Something along the lines of:

```python
@pytest.mark.parametrize(['build_image', 'build_imagestream', 'worker_build_image'], [
    ('fedora:latest', None, 'fedora:latest'),
    ...
])
```

twaugh:
Any objections to merging? Until https://github.com/projectatomic/atomic-reactor/pull/669 is merged this parameter will be ignored.

lcarva:
+1 for merging

rcerven:
yop it can be safely merged

PR #539, opened 2017-03-23
can_orchestrate option in config

rcerven:

### DIFF
--- a/tests/fake_api.py
+++ b/tests/fake_api.py
@@ -323,6 +323,7 @@ distribution_scope = authoritative-source-only
 koji_root = http://koji.example.com/kojiroot
 koji_hub = http://koji.example.com/kojihub
 use_auth = false
+can_orchestrate = true
 """.format(build_json_dir="inputs"))
         fp.flush()
         dummy_config = Configuration(fp.name)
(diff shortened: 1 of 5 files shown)


osbs/cli/main.py
```diff
@@ -352,7 +352,7 @@ def cmd_build(args, osbs):
         required_args = ['platform', 'release']
         unwanted_args = ['platforms']
         create_func = osbs.create_worker_build
-    elif args.orchestrator:
+    elif args.orchestrator and osbs.can_orchestrate():
```
twaugh: Shouldn't specifying `--orchestrator` cause an error when can_orchestrate is set false?

rcerven: well, it actually will fail, as both builds have different requirements, so even when here is fallback to prod_build, it won't pass required args check

twaugh: If you leave this line unchanged, the failure will be logged using the caught exception, something like:
```
Exception caught: OsbsValidationException("can't create orchestrate build when can_orchestrate isn't enabled")
```

Isn't that better? Otherwise the user will be tinkering around with their parameters trying to get their orchestrator build to work, and when they finally get something submitted it won't be the orchestrator build they asked for.

vrutkovs: Perhaps `get_can_orchestrate` should resolve it? It would override `can_orchestrate` value if `--orchestrate` is set and necessary secrets are specified

rcerven: mmm --orchestrate should be overriding can_orchestrate,

can_orchestrate is option which allows it at all,
--orchestrator is just saying I want to do it

rcerven: yeah I can just go to create_orchestrator_build and it will do exception there,
at least it will be same as using api

twaugh: You could also say the API method should override it then. :-)

I think the config option should be honoured. If we have to change it in order to call the API, we know the config is wrong.

twaugh: Yes, I like this approach, just reporting what the API does.

vrutkovs: >can_orchestrate is option which allows it at all,
--orchestrator is just saying I want to do it

Ah, yes, agree. Currently k-cb assumes that if OSBS can orchestrate it will orchestrate - is this still true? Would any k-cb changes be required after this PR is merged?

tests/fake_api.py
```diff
@@ -323,6 +323,7 @@ def osbs(openshift):
 koji_root = http://koji.example.com/kojiroot
 koji_hub = http://koji.example.com/kojihub
 use_auth = false
+can_orchestrate = true
```
twaugh: I was going to suggest adding this to osbs106 too but it seems that's pretty stale.

PR #541, opened 2017-03-30
Retry on conflict updating labels/annotations

twaugh:
We retry up to 10 times, with a delay of 500ms between attempts.

This is necessary because OpenShift itself may be updating the Build object as well.

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -6,6 +6,7 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 from __future__ import print_function, unicode_literals, absolute_import
+from functools import wraps
 import json
 import os
 import numbers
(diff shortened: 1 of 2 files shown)


osbs/core.py
```diff
@@ -53,6 +54,27 @@ def check_response(response):
...
+            if attempt != 0:
+                time.sleep(sleep_seconds)
+
+            try:
+                return func(*args, **kwargs)
+            except OsbsResponseException as ex:
+                if ex.status_code != httplib.CONFLICT:
+                    raise
+
+                last_exception = ex
```
vrutkovs: I wonder if it make sense to log the contents of the Build object so that our theory that its a 3.4 regression?

twaugh: Added debug logging.

rcerven:
+1

PR #543, opened 2017-03-31
render koji_tag_build plugin

lcarva:

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2015 Red Hat, Inc
+Copyright (c) 2015, 2016, 2017 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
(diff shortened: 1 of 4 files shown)


osbs/build/build_request.py
```diff
@@ -603,6 +604,33 @@ def render_koji_promote(self, use_auth=None):
...
+        self.dj.dock_json_set_arg(phase, plugin, 'kojihub', self.spec.kojihub.value)
+        self.dj.dock_json_set_arg(phase, plugin, 'target', self.spec.koji_target.value)
+
+        if (self.spec.koji_use_kerberos.value and
+                self.spec.koji_kerberos_principal.value and
+                self.spec.koji_kerberos_keytab.value):
+            self.dj.dock_json_set_arg(phase, plugin, 'koji_principal',
+                                      self.spec.koji_kerberos_principal.value)
+            self.dj.dock_json_set_arg(phase, plugin, 'koji_keytab',
+                                      self.spec.koji_kerberos_keytab.value)
```
twaugh: Should this share any code with render_koji_promote?

In the future there will also be koji_upload, which will *also* want Koji/Kerberos config.

lcarva: I think we need to come up with a way to better make kojihub available to plugins. Something like this https://github.com/projectatomic/atomic-reactor/issues/478.

PR #544, opened 2017-04-03
Configure koji_ssl_certs_dir for bump_release and koji plugins

vrutkovs:
Related: https://github.com/projectatomic/atomic-reactor/pull/676


twaugh:
Does this need unit testing?

vrutkovs:
I don't think it really does, atomic-reactor would skip unknown args iirc

PR #546, opened 2017-04-04
Add tito configs

vrutkovs:
This also fixes a typo in a spec file


rel-eng/packages/osbs-client
```diff
@@ -0,0 +1 @@
+0.36-1 ./
```
twaugh: How does this relate to the version we already store elsewhere?

setup.py
```diff
@@ -45,6 +45,6 @@ def _install_requirements():
     },
     install_requires=_install_requirements(),
     data_files=data_files.items(),
-    setup_requires=['pytest-runner'],
+    setup_requires=[],
```
twaugh: Would  be nice if the commit message explained why we want to do this.

rel-eng/packages/osbs-client
vrutkovs: This is a pointer for tito of a current version and package subfolder. This file would be automatically updated on `tito release` - but I think it would work even if this version is outdated

setup.py
vrutkovs: Updated (in #547)

vrutkovs:
Note, that this PR won't bind us to tito, it will allow us to build SRPMs for COPR easier

osbs/version.py
```diff
@@ -0,0 +1,9 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
```
twaugh: 2017

rel-eng/version.template
```diff
@@ -0,0 +1,9 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
```
twaugh: 2017

osbs/version.py
vrutkovs: Done

PR #549, opened 2017-04-06
LDAP auth method support

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -15,6 +15,7 @@ import stat
 import sys
 import warnings
 import datetime
+import getpass
 from functools import wraps
 
 from osbs.build.build_request import BuildRequest
@@ -28,6 +29,13 @@ from osbs.core import Openshift
 from osbs.exceptions import OsbsException, OsbsValidationException, OsbsResponseException
 # import utils in this way, so that we can mock standalone functions with flexmock
 from osbs import utils
+
+try:
+    # py2
+    import httplib
+except ImportError:
+    # py3
+    import http.client as httplib
 
 
 # Decorator for API methods.
@@ -726,10 +734,43 @@ class OSBS(object):
 
     @osbsapi
     def get_token(self):
-        return self.os.get_oauth_token()
+        if self.os.use_kerberos:
+            return self.os.get_oauth_token()
+        else:
+            if self.os.token:
+                return self.os.token
+
+            raise OsbsValidationException("no token stored")
 
     @osbsapi
-    def login(self, token):
+    def login(self, token=None, username=None, password=None):
+        if self.os.use_kerberos:
+             raise OsbsValidationException("can't use login when using kerberos")
+
+        if not token:
+            if username:
+                self.os.username = username
+            else:
+                try:
+                    self.os.username = raw_input("Username: ")
+                except NameError:
+                    self.os.username = input("Username: ")
+
+            if password:
+                self.os.password = password
+            else:
+                self.os.password = getpass.getpass()
+            self.os.use_auth = True
+            token = self.os.get_oauth_token()
+
+        self.os.token = token
+        try:
+            self.os.get_user()
+        except OsbsResponseException as ex:
+            if ex.status_code == httplib.UNAUTHORIZED:
+                raise OsbsValidationException("token is not valid")
+            raise
+
         token_file = utils.get_instance_token_file_name(self.os_conf.conf_section)
         token_file_dir = os.path.dirname(token_file)
 
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -441,7 +441,10 @@ def cmd_get_token(args, osbs):  # pylint: disable=W0613
 
 
 def cmd_login(args, osbs):
-    osbs.login(args.TOKEN)
+    if args.token and (args.username or args.password):
+        print("Can't use --token and (--user or --password).")
+        return
+    osbs.login(args.token, args.username, args.password)
 
 
 def cmd_get_user(args, osbs):
@@ -625,7 +628,9 @@ def cli():
 
     get_login_parser = subparsers.add_parser(str_on_2_unicode_on_3('login'),
                                              help='perform login and store token for later use')
-    get_login_parser.add_argument('TOKEN', help='token to be used for login')
+    get_login_parser.add_argument('--token', help='token to be used for login', action="store")
+    get_login_parser.add_argument('-u', '--username', help='Username, will prompt if not provided', action="store")
+    get_login_parser.add_argument('-p', '--password', help='Password, will prompt if not provided', action="store")
     get_login_parser.set_defaults(func=cmd_login)
 
     get_user_parser = subparsers.add_parser(str_on_2_unicode_on_3('get-user'), help='get info about user')
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -18,6 +18,8 @@ import six
 import stat
 import copy
 import datetime
+import getpass
+import sys
 from tempfile import NamedTemporaryFile
 
 from osbs.api import OSBS, osbsapi
@@ -40,7 +42,14 @@ from tests.constants import (TEST_ARCH, TEST_BUILD, TEST_COMPONENT, TEST_GIT_BRA
                              TEST_GIT_URI, TEST_TARGET, TEST_USER, INPUTS_PATH,
                              TEST_KOJI_TASK_ID, TEST_VERSION)
 from tests.fake_api import openshift, osbs, osbs106, osbs_cant_orchestrate
+from osbs.core import Openshift
 
+try:
+    # py2
+    import httplib
+except ImportError:
+    # py3
+    import http.client as httplib
 
 def request_as_response(request):
     """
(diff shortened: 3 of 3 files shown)


rcerven:
tests will follow

osbs/api.py
```diff
@@ -726,28 +727,54 @@ def import_image(self, name):
...
+                self.os.password = getpass.getpass()
+            self.os.use_auth = True
+            token = self.os.get_oauth_token()
+
+        if token:
+            self.os.token = token
+            if self.os.get_user_login_test().status_code == 401:
+                raise OsbsValidationException("token is not valid")
+
+            token_file = utils.get_instance_token_file_name(self.os_conf.conf_section)
```
vrutkovs: Looks like osbs-client would always try to write token to `token_file` now, right?

rcerven: yes, if the token is valid

vrutkovs: I don't really see the point in storing the token across sessions. Its perfectly valid to request a new one for each session.

Also, this would make `token_file` param in osbs.conf obligatory

rcerven: it isn't storing it all the time, only when you do login and provide valid token (or gets token when you enter user/pass)

rcerven: token_file param isn't mandatory because it will default to instance name

osbs/api.py
```diff
@@ -726,28 +727,54 @@ def import_image(self, name):
...
-        token_file_dir = os.path.dirname(token_file)
+    def login(self, token, username, password):
+        if self.os.use_kerberos:
+             raise OsbsValidationException("can't use login when using kerberos")
+
+        if not token:
+            if username:
+                self.os.username = username
+            else:
+                self.os.username = getpass.getpass("Username: ")
```
lcarva: Let's use `raw_input` for username. So user can see what he/she is typing.

osbs/api.py
```diff
@@ -726,28 +727,54 @@ def import_image(self, name):
...
-    def login(self, token):
-        token_file = utils.get_instance_token_file_name(self.os_conf.conf_section)
-        token_file_dir = os.path.dirname(token_file)
+    def login(self, token, username, password):
+        if self.os.use_kerberos:
+             raise OsbsValidationException("can't use login when using kerberos")
+
+        if not token:
+            if username:
+                self.os.username = username
```
lcarva: could simplify this with: `self.os.username = username or raw_input('Username :')`

osbs/api.py
lcarva: `token_file` in osbs.conf is different than the token file stored at `~/.osbs/<instance_name>.token`

osbs/api.py
```diff
@@ -726,28 +727,54 @@ def import_image(self, name):
...
+            else:
+                self.os.username = getpass.getpass("Username: ")
+            if password:
+                self.os.password = password
+            else:
+                self.os.password = getpass.getpass()
+            self.os.use_auth = True
+            token = self.os.get_oauth_token()
+
+        if token:
```
lcarva: Should we raise an error if there isn't a token?

osbs/api.py
```diff
@@ -726,28 +727,54 @@ def import_image(self, name):
...
+            self.os.token = token
+            if self.os.get_user_login_test().status_code == 401:
```
lcarva: Why not just use `get_user` and wrap in try/except?

osbs/api.py
rcerven: that will happen only if credentials weren't ok, in that case that itself will return error and stop

osbs/api.py
rcerven: because whole api is wrapped already in osbsapi, which catches that

osbs/api.py
rcerven: ok

osbs/api.py
vrutkovs: Nitpick: use `httplib.UNAUTHORIZED` instead of hardcoded HTTP error number

rcerven: ok

osbs/api.py
```diff
@@ -726,28 +734,58 @@ def import_image(self, name):
...
+                self.os.password = password
+            else:
+                self.os.password = getpass.getpass()
+            self.os.use_auth = True
+            token = self.os.get_oauth_token()
+
+        if token:
+            self.os.token = token
+            if self.os.get_user_login_test().status_code == httplib.UNAUTHORIZED:
+                raise OsbsValidationException("token is not valid")
```
twaugh: Is it important for this to raise OsbsValidationException instead of OsbsResponseException for some reason?

If so, why not do this?:
```
try:
    self.os.get_user()
except OsbsResponseException as ex:
    if ex.status_code == httplib.UNAUTHORIZED:
        raise OsbsValidationException(...)
    raise
```

It doesn't need a new core method.

osbs/api.py
twaugh: Where does the error get generated? From get_oauth_token()?

rcerven: yes

lcarva: Why the `if-statement` then? Seems to me we should either raise an error if `token` is falsy, or not perform the check at all because something else is guaranteeing it won't be falsy.

rcerven: yeah it isn't needed that was just another safety check :)

osbs/api.py
rcerven: ok

osbs/api.py
lcarva: Let's remove it then, because if we do encounter a scenario where `get_oauth_token` returns falsy, this method will complete execution with no indication that something went wrong.

rcerven: yeah removed

osbs/api.py
```diff
@@ -726,10 +734,43 @@ def import_image(self, name):
...
+            else:
+                self.os.password = getpass.getpass()
+            self.os.use_auth = True
+            token = self.os.get_oauth_token()
+
+        self.os.token = token
+        try:
+            self.os.get_user()
+        except OsbsResponseException as ex:
+            if ex.status_code == httplib.UNAUTHORIZED:
```
lcarva: `atomic-reactor` uses `requests`, we can use it here instead: `requests.codes.UNAUTHORIZED`

twaugh: We're in osbs-client here, which doesn't currently use `requests` though.

lcarva: They both map to `401` regardless. This is a nitpick, but it'd avoid the try-except in importing httplib.

osbs/api.py
```diff
@@ -726,10 +734,43 @@ def import_image(self, name):
...
+                raise OsbsValidationException("token is not valid")
```
vrutkovs: `logger.error` should be used in order to retain the details of original exception

rcerven: original exception is already logged in core via check_response

osbs/api.py
```diff
@@ -726,10 +734,43 @@ def import_image(self, name):
...
+            return self.os.get_oauth_token()
+        else:
+            if self.os.token:
+                return self.os.token
+
+            raise OsbsValidationException("no token stored")
 
     @osbsapi
-    def login(self, token):
+    def login(self, token, username, password):
```
twaugh: This method signature looks wrong. Currently it requires 3 parameters. I think all of them are optional though.

Perhaps instead it should be "login(self, token=None, username=None, password=None):".

osbs/api.py
twaugh: I think it needs the try/except part for Python2/3 handling though.

osbs/api.py
twaugh: Let's not add a new dependency for this though. :-) I think it looks good as-is.

osbs/cli/main.py
```diff
@@ -625,7 +625,10 @@ def cli():
 
     get_login_parser = subparsers.add_parser(str_on_2_unicode_on_3('login'),
                                              help='perform login and store token for later use')
-    get_login_parser.add_argument('TOKEN', help='token to be used for login')
+    get_login_parser.add_argument('--token', help='token to be used for login '
+                                                  '(takes precedence over username and password)', action="store")
```
twaugh: Should it be an error to provide both a token and a username+password?

...if so, should that be at the API level?

osbs/api.py
rcerven: yes

osbs/cli/main.py
rcerven: it isn't possible to create exclusive groups between 2 and 1 parameters

I wrote in there that it takes precedence but I can also just make it error if both variants are provided

osbs/api.py
rcerven: ok

osbs/cli/main.py
rcerven: ok I'll report error when 2 variants are provided

osbs/cli/main.py
```diff
@@ -441,7 +441,10 @@ def cmd_get_token(args, osbs):  # pylint: disable=W0613
 
 
 def cmd_login(args, osbs):
-    osbs.login(args.TOKEN)
+    if args.token and (args.username or args.password):
+        print("Can't use --token and (--user or --pasword).")
```
twaugh: Typo: pasword should be password

osbs/cli/main.py
```diff
@@ -441,7 +441,10 @@ def cmd_get_token(args, osbs):  # pylint: disable=W0613
...
+        return
```
twaugh: Should this be an error?

osbs/cli/main.py
twaugh: Now that we report an error, the help for `--token` needn't mention username and password.

osbs/cli/main.py
rcerven: not sure, based on 406 which is also catching wrong param combinations
it looks a bit weird when printed via logger.error

osbs/cli/main.py
rcerven: yop

osbs/cli/main.py
rcerven: yop

PR #550, opened 2017-04-11
LDAP auth method, cosmetic changes

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -740,7 +740,7 @@ class OSBS(object):
             if self.os.token:
                 return self.os.token
 
-            raise OsbsValidationException("no token stored")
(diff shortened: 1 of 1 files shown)


csomh:
+1

osbs/api.py
```diff
@@ -740,7 +740,7 @@ def get_token(self):
             if self.os.token:
                 return self.os.token
 
-            raise OsbsValidationException("no token stored")
+            raise OsbsValidationException("no token stored for %s" % self.os_conf.conf_section)
```
vrutkovs: Tests should verify the error message

rcerven: test is catching exception, not message, I don't think it is worth to checking that

Issue #483, opened 2016-12-01
scratch builds should have a short random string added to the end of the build name


vrutkovs:
If several scratch builds are started some might fail if those are submitted to OSE too fast - within one second. 

osbs-client should add a short random string to the build name to ensure this doesn't happen.

lcarva:
the random string should actually be part of the tag as well in order to avoid tag collision if the two builds end up in the same docker repo.

vrutkovs:
The tags would get overwritten, they won't collide

lcarva:
The 4-digit string produced is not random:
```
 scratch-63379-20170531181947                 | running                 |                            
 scratch-63379-20170531182008                 | running                 |                            
 scratch-63379-20170531182009                 | running                 |
```

vrutkovs:
Probably RNG was not seeded, as it returns the same value for any build

PR #551, opened 2017-04-11
Make scratch build names unique

csomh:
As of now two scratch builds could have the same name,
due to the fact that build names were generated using a timestamp.

This changes is adding a random 5 digit number to the build names,
which should ensure uniqueness of the build names.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -14,7 +14,6 @@ import os.path
 import stat
 import sys
 import warnings
-import datetime
 import getpass
 from functools import wraps
 
@@ -305,7 +304,10 @@ class OSBS(object):
             builder_img['kind'] = 'DockerImage'
             builder_img['name'] = ref
 
-        build_config_name = 'scratch-%s' % datetime.datetime.now().strftime('%Y%m%d%H%M%S')
+        output_image_name = build_json['spec']['output']['to']['name']
+        # Reuse random string and timestamp values.
+        build_config_name = 'scratch-%s-%s' % tuple(
+            output_image_name.rsplit('-', 2)[-2:])
         logger.debug('starting scratch build %s', build_config_name)
         build_json['metadata']['name'] = build_config_name
         return BuildResponse(self.os.create_build(build_json).json())
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -288,8 +288,6 @@ class OSBS(object):
         logger.debug(build_request)
         build_json = build_request.render()
         build_json['kind'] = 'Build'
-        if 'spec' not in build_json.keys():
-            build_json['spec'] = {}
         build_json['spec']['serviceAccount'] = 'builder'
         build_json['metadata']['labels']['scratch'] = 'true'
 
--- a/tests/build_/test_build_request.py
+++ b/tests/build_/test_build_request.py
@@ -205,7 +205,7 @@ class TestBuildRequest(object):
         assert build_json["spec"]["source"]["git"]["uri"] == TEST_GIT_URI
         assert build_json["spec"]["source"]["git"]["ref"] == TEST_GIT_REF
 
-        expected_output = "john-foo/component:none-20"
+        expected_output = "john-foo/component:none-"
         if registry_uris:
             expected_output = registry_uris[0] + "/" + expected_output
         assert build_json["spec"]["output"]["to"]["name"].startswith(expected_output)
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -8,14 +8,14 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import print_function, absolute_import, unicode_literals
 
 import logging
-import datetime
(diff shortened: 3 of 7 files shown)


osbs/api.py
```diff
@@ -305,7 +306,9 @@ def _create_scratch_build(self, build_request):
             builder_img['kind'] = 'DockerImage'
             builder_img['name'] = ref
 
-        build_config_name = 'scratch-%s' % datetime.datetime.now().strftime('%Y%m%d%H%M%S')
+        build_config_name = 'scratch-%s-%s' % (
```
vrutkovs: Maybe ```import uuid``` and ```str(uuid.uuid1())[:8]``` would be simpler?

lcarva:
Actually, I'm not sure anymore if this will not also update the unique tag used. Could you verify it?

twaugh:
Need to update this part I think: https://github.com/projectatomic/osbs-client/blob/master/osbs/build/spec.py#L300

osbs/api.py
csomh: `uuid.uuid1()` seems to be doing much more then `random`:

```
Generate a UUID from a host ID, sequence number, and the current time.
If 'node' is not given, getnode() is used to obtain the hardware
address.  If 'clock_seq' is given, it is used as the sequence number;
otherwise a random 14-bit sequence number is chosen.
```
Would there be any other benefits of using it?

Also: switched to using `randrange` instead of `randint`, which had the strange behavior to include the end point.

csomh:
Unit tests added. Also tried to solve #483 (unique tags), but somewhat got lost between BuildRequest and BuildSpec usage and how this design should work.

Anyways, this should ensure tag uniqueness and reuses the same unique string to update build names for scratch builds, by reusing the last 20 chars from the output name.

Let me know what you think about it.

osbs/api.py
```diff
@@ -305,7 +302,8 @@ def _create_scratch_build(self, build_request):
             builder_img['kind'] = 'DockerImage'
             builder_img['name'] = ref
 
-        build_config_name = 'scratch-%s' % datetime.datetime.now().strftime('%Y%m%d%H%M%S')
+        build_config_name = 'scratch-%s' % (
+            build_json['spec']['output']['to']['name'][-20:])
```
vrutkovs: 20 here is a magic number - lets not use those or at least use a constant for it.

Why is it even set to 20? Is this the registry hostname part?

tests/build_/test_build_request.py
```diff
@@ -205,7 +205,7 @@ def test_render_simple_request(self, registry_uris, build_image):
         assert build_json["spec"]["source"]["git"]["uri"] == TEST_GIT_URI
         assert build_json["spec"]["source"]["git"]["ref"] == TEST_GIT_REF
 
-        expected_output = "john-foo/component:none-20"
+        expected_output = "john-foo/component:none-"
```
vrutkovs: Tests should mock `random.randrange` and check that the full image is verified

osbs/api.py
```diff
@@ -289,8 +288,6 @@ def _create_scratch_build(self, build_request):
         logger.debug(build_request)
         build_json = build_request.render()
         build_json['kind'] = 'Build'
-        if 'spec' not in build_json.keys():
```
lcarva: Why was this removed? Seems unrelated.

osbs/api.py
lcarva: `spec.output.to.name` refers to the image that this build should produce. This seems like an attempt to re-use the timestamp and random string used there. I like that, but it's not very clear from the code and it seems somewhat brittle.

How about something like this:
```python
output_image_name = build_json['spec']['output']['to']['name']
# Reuse random string and timestamp values.
build_config_name = 'scratch-%s-%s' % tuple(output_image_name.rsplit('-', 2)[-2:])
```

csomh: Yes, this was the intent. Updated like recommended to make it easier to understand. Thanks!

tests/build_/test_build_request.py
csomh: `random.randrange` is called in `BuildSpec`, so it's tested there. Updated those tests to mock both this and `utcnow()`.

For this test, I think it's enough to check only `BuildRequest` functionality (that output name is prefixed with `registry_uris[0]`, when there is one).

osbs/api.py
csomh: I just stumbled upon this, and yes it's unrelated. See comment in commit msg:

```
Remove check for 'spec' from _create_scratch_build

After rendering the build_request, a check was made
to see if the rendered build_json has 'spec'.

This is redundant, as build_request objects already have
'spec' in them.
```

twaugh:
Does commit 61c83ad4a1f298b210e4e00b62bea6402dc75c66 need squashing into some other commit before this is merged?

csomh:
Squashed and rebased.

PR #553, opened 2017-04-19
URL to info page as docker label

rcerven:

### DIFF
--- a/docs/configuration_file.md
+++ b/docs/configuration_file.md
@@ -137,6 +137,7 @@ Some options are also mandatory.
 
 * `can_orchestrate` (*optional*, `boolean`) — allows using orchestrator build, default is false
 
+* `info_url_format` (*optional*, `string`) — format for `url` Dockerfile label, used as a Python format string; replacement field keywords are label names, and they will be replaced with the value of the named label
 
 ## Build JSON Templates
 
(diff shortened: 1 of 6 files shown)


docs/configuration_file.md
```diff
@@ -137,6 +137,9 @@ Some options are also mandatory.
 
 * `can_orchestrate` (*optional*, `boolean`) — allows using orchestrator build, default is false
 
+* `info_url_format` (*optional*, `string`) — format for `url` dockerfile label, by default :
+`https://access.redhat.com/containers/#/{authoritative-source-url}/{name}/images/{version}-{release}`, you can use any label,
```
vrutkovs: Maybe we should set this to a Docker Hub by default, being an opensource project? `https://hub.docker.com/{name}`

twaugh: I actually think we shouldn't have a default value. If the config file doesn't specify a url format we should leave that label alone.

rcerven: removed default

docs/configuration_file.md
```diff
@@ -137,6 +137,7 @@ Some options are also mandatory.
 
 * `can_orchestrate` (*optional*, `boolean`) — allows using orchestrator build, default is false
 
+* `info_url_format` (*optional*, `string`) — format for `url` dockerfile label, you can use `{label_name}` with any label, and if the label is defined it will be substituted.
```
twaugh: Now this is using .format() let's make the documentation clear on that point: "format for `url` Dockerfile label, used as a Python format string; replacement field keywords are label names, and they will be replaced with the value of the named label"

PR #556, opened 2017-04-26
Set a custom nodeselector for scratch builds if low_priority_node_selector config param is set

vrutkovs:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -294,6 +294,9 @@ class OSBS(object):
         build_json['spec']['serviceAccount'] = 'builder'
         build_json['metadata']['labels']['scratch'] = 'true'
 
+        if build_request.low_priority_node_selector:
+            build_json['spec']['nodeSelector'] = build_request.low_priority_node_selector
+
         builder_img = build_json['spec']['strategy']['customStrategy']['from']
         kind = builder_img['kind']
         if kind == 'ImageStreamTag':
@@ -501,6 +504,7 @@ class OSBS(object):
             arrangement_version=arrangement_version,
             info_url_format=self.build_conf.get_info_url_format(),
             artifacts_allowed_domains=self.build_conf.get_artifacts_allowed_domains(),
+            low_priority_node_selector=self.build_conf.get_low_priority_node_selector()
         )
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
         if build_request.scratch:
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -499,3 +499,14 @@ class Configuration(object):
             return [x.strip() for x in value.split(',')]
         else:
             return []
+
+    def get_low_priority_node_selector(self):
+        nodeselector = {}
+        nodeselector_str = self._get_value("low_priority_node_selector", self.conf_section,
+                                           "low_priority_node_selector")
+        if nodeselector_str:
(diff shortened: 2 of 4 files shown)


rcerven:
how does that test tests anything? you flexmock build_request and then you update json

vrutkovs:
The tests are flexmocking `BuildRequest`'s string, received from configuration and checking that parsed dict is found in a json, used to create a new Openshift `Build` object.

The only thing missing is direct read from configuration, but it looks trivial enough to me (and somehow already being covered by other config tests)

osbs/api.py
```diff
@@ -294,6 +294,13 @@ def _create_scratch_build(self, build_request):
         build_json['spec']['serviceAccount'] = 'builder'
         build_json['metadata']['labels']['scratch'] = 'true'
 
+        if build_request.low_priority_node_selector:
+            nodeselector = {}
+            for item in build_request.low_priority_node_selector.split(','):
```
lcarva: Can't we do the string manipulation in only in conf.py? and assume that at the api level it should already be a list?

vrutkovs: I guess we could, yes, I don't really have a preference

twaugh: My preference would be to do this in conf.py, so that at the API level it is a dict.

vrutkovs:
Updated the PR: nodeselector now being parsed in `conf.py`, tests are reading the configuration explicitely

osbs/conf.py
```diff
@@ -491,3 +491,14 @@ def get_can_orchestrate(self):
...
                                "info_url_format")
+
+    def get_low_priority_node_selector(self):
+        nodeselector = {}
+        nodeselector_str = self._get_value("low_priority_node_selector", self.conf_section,
+                                           "low_priority_node_selector")
+        if nodeselector_str:
+            for item in nodeselector_str.split(','):
+                splitted_list = [x.strip() for x in item.strip().split('=')]
+                nodeselector[splitted_list[0]] = splitted_list[1]
```
twaugh: I don't like the variable name `splitted_list`. How about this?:

```
constraints = nodeselector_str.split(',')
nodeselector = dict([constraint.split('=', 1) for constraint in constraints])
```

vrutkovs: Nice, although it doesn't strip the values. I'll rework it a bit

vrutkovs: Reworked it have a (rather complex) list comprehension statement

osbs/conf.py
```diff
@@ -491,3 +491,13 @@ def get_can_orchestrate(self):
...
         return self._get_value("info_url_format", self.conf_section,
                                "info_url_format")
+
+    def get_low_priority_node_selector(self):
+        nodeselector = {}
+        nodeselector_str = self._get_value("low_priority_node_selector", self.conf_section,
+                                           "low_priority_node_selector")
+        if nodeselector_str:
+            nodeselector = dict([[x.strip() for x in constraint.split('=', 1)]
+                                 for constraint in nodeselector_str.split(',')])
```
twaugh: My preference would be to keep to only one `for` in each comprehension, but otherwise this looks fine.

vrutkovs: Rewrote this to use two dict comprehension - first to split keys and values - and the second one to strip extra spaces

PR #558, opened 2017-04-28
add equal label aliases

rcerven:


osbs/utils.py
```diff
@@ -411,7 +419,13 @@ class Labels(object):
...
         LABEL_TYPE_SOURCE: ('authoritative-source-url', 'Authoritative_Registry'),
         LABEL_TYPE_COMPONENT: ('com.redhat.component', 'BZComponent'),
-        LABEL_TYPE_HOST: ('com.redhat.build-host', 'Build_Host')
+        LABEL_TYPE_HOST: ('com.redhat.build-host', 'Build_Host'),
+        LABEL_TYPE_RUN: ('run', 'RUN'),
+        LABEL_TYPE_INSTALL: ('install', 'INSTALL'),
+        LABEL_TYPE_UNINSTALL: ('uninstall', 'UNINSTALL')
+    }
+    LABEL_NAMES_EQUAL = {
+        LABEL_TYPE_DESCRIPTION: ('description', 'io.k8s.description')
```
vrutkovs: Why do we consider those equal? https://github.com/projectatomic/ContainerApplicationGenericLabels/blob/master/vendor/redhat/labels.md lists both, but doesn't state that any of those can be used

osbs/utils.py
```diff
@@ -411,7 +419,13 @@ class Labels(object):
         LABEL_TYPE_VENDOR: ('vendor', 'Vendor'),
         LABEL_TYPE_SOURCE: ('authoritative-source-url', 'Authoritative_Registry'),
         LABEL_TYPE_COMPONENT: ('com.redhat.component', 'BZComponent'),
-        LABEL_TYPE_HOST: ('com.redhat.build-host', 'Build_Host')
+        LABEL_TYPE_HOST: ('com.redhat.build-host', 'Build_Host'),
+        LABEL_TYPE_RUN: ('run', 'RUN'),
+        LABEL_TYPE_INSTALL: ('install', 'INSTALL'),
+        LABEL_TYPE_UNINSTALL: ('uninstall', 'UNINSTALL')
+    }
+    LABEL_NAMES_EQUAL = {
```
lcarva: Where's this used?

rcerven: it is used in corresponding PR in atomic-reactor

osbs/utils.py
twaugh: Agreed, I think the particular sets of equal-preference label aliases to apply should be configurable in osbs.conf, including 'none'.

rcerven: so in osbs.conf something like 

equal_labels = ('desc', 'io.k8s.desc'),('new1', 'new2')

?

rcerven: I guess this seems nicer

equal_labels= label1:label2, label3:label4'

rcerven: ';' group separator, and ',' item separator

seems even better

docs/configuration_file.md
```diff
@@ -143,6 +143,8 @@ Some options are also mandatory.
 
 * `low_priority_node_selector` (*optional*, `string`) — a nodeselector to be applied to builds, which are considered to have less priority than normal ones
 
+* `equal_labels` (*optional*, `string`) — list of equal labels groups, used for adding all equal labels into dockerfile if any is present, label delimiter ':', group delimiter ',' (e.g. `name1:name2:name3, release1:release2, version1:version2`)
```
twaugh: "equal labels" -> "equal-preference labels", and please add explanation: "if any of each set is missing, aliases will be added to complete the set".

osbs/conf.py
```diff
@@ -510,3 +511,18 @@ def get_low_priority_node_selector(self):
             nodeselector = dict([k.strip(), v.strip()] for (k, v) in raw_nodeselector.items())
 
         return nodeselector
+
+    def get_equal_labels(self):
+        equal_labels = []
+        equal_labels_str = self._get_value("equal_labels", self.conf_section,
+                                           "equal_labels")
+        if equal_labels_str:
+            if re.match('^[^:, ]+(:[^:, ]+)+\s*(,\s*[^:, ]+(:[^:, ]+\s*)+)*$', equal_labels_str):
```
twaugh: I feel like we should either use a regular expression to validate and parse the value, or else use string manipulation functions to validate and parse, rather than re to validate and funcs to parse.

If we *do* use a regular expression, I'd like a comment explaining what the intention behind it is. :-)

docs/configuration_file.md
rcerven: do you want to rename 'equal_labels' to the 'equal-preference_labels' or?

do you want me to just add your explanation or replace whole mine?

osbs/conf.py
rcerven: ok I'll add comment, it is just nice to validate it via regs, but for parsing is better just string func

docs/configuration_file.md
twaugh: It can still be called `equal_labels`, I don't mind that. We definitely need to document what the syntax is, which is what you have -- in addition, I would like the explanation I wrote to be included. Something like:
```
* `equal_labels` (*optional*, `string`) -- list of equal-preference label groups; if any of each set is missing, aliases will be added to complete the set; label delimiter ':', group delimiter ',' (e.g. `name1:name2:name3, release1:release2, version1:version2`)
```

rcerven: ok got it

osbs/conf.py
twaugh: FWIW, I think the regular expression you have here is not accurate. We want to express this I think:

```
equal_labels ::= <set> <set>*
set ::= <label> ":" <label ":">+
label ::= <valid_ch>+
valid_ch ::= (valid label character)
```

osbs/conf.py
```diff
@@ -510,3 +511,18 @@ def get_low_priority_node_selector(self):
...
+                label_groups = equal_labels_str.split(',')
+
+                for label_group in label_groups:
+                    equal_labels.append([label.strip() for label in label_group.split(':')])
```
twaugh: Since we're using `\s`, shouldn't we have that also in the exclude-class [^...] part?

tests/test_conf.py
```diff
@@ -355,3 +355,41 @@ def test_get_allowed_artifacts_domain(self, config, expected):
         with self.config_file(config) as config_file:
             conf = Configuration(conf_file=config_file)
             assert conf.get_artifacts_allowed_domains() == expected
+
+    @pytest.mark.parametrize(('config', 'expected'), [
+        ({'default': {'equal_labels': 'label1:label2, label3:label4:label5, label6:label7'}},
+         [['label1', 'label2'], ['label3', 'label4', 'label5'], ['label6', 'label7']]),
+
+        ({'default': {'equal_labels': '      label1:label2   , label3:label4:label5, label6:label7    '}},
```
twaugh: Can we split this up a bit?

tests/test_conf.py
```diff
@@ -355,3 +355,41 @@ def test_get_allowed_artifacts_domain(self, config, expected):
...
+         [['label1', 'label2'], ['label3', 'label4', 'label5'], ['label6', 'label7']]),
+
+        ({'default': {'equal_labels': '      label1:label2:   , label3:label4:label5, label6:label7    '}},
```
twaugh: This looks pretty easy to split up.

tests/test_conf.py
```diff
@@ -355,3 +355,41 @@ def test_get_allowed_artifacts_domain(self, config, expected):
...
+         OsbsValidationException),
+
+        ({'default': {'equal_labels': 'label1:label2,   , label3:label4:label5, label6:label7    '}},
```
twaugh: Same.

osbs/conf.py
```diff
@@ -510,3 +511,22 @@ def get_low_priority_node_selector(self):
             nodeselector = dict([k.strip(), v.strip()] for (k, v) in raw_nodeselector.items())
 
         return nodeselector
+
+    def get_equal_labels(self):
+        equal_labels = []
+        equal_labels_str = self._get_value("equal_labels", self.conf_section,
+                                           "equal_labels")
+        if equal_labels_str:
+            # must be in correct format (e.g. `name1:name2:name3, release1:release2, version1:version2`)
```
twaugh: Maybe put the example on its own line?

osbs/conf.py
```diff
@@ -510,3 +511,23 @@ def get_low_priority_node_selector(self):
...
+        equal_labels = []
+        equal_labels_str = self._get_value("equal_labels", self.conf_section,
+                                           "equal_labels")
+        if equal_labels_str:
+            # must be in correct format
+            # e.g. `name1:name2:name3, release1:release2, version1:version2`
+            # ',' separator for groups
+            # ':' separator for equal preference labels
+            # there has to be at least 2 labels in equal labels in group
+            if re.match('^[^:,\s]+(:[^:,\s]+)+\s*(,\s*[^:,\s]+(:[^:,\s]+\s*)+)*$', equal_labels_str):
```
twaugh: Wrap the parameters to avoid the line length limit?

rcerven: done

osbs/conf.py
```diff
@@ -510,3 +511,24 @@ def get_low_priority_node_selector(self):
             nodeselector = dict([k.strip(), v.strip()] for (k, v) in raw_nodeselector.items())
 
         return nodeselector
+
+    def get_equal_labels(self):
+        equal_labels = []
+        equal_labels_str = self._get_value("equal_labels", self.conf_section,
+                                           "equal_labels")
+        if equal_labels_str:
+            # must be in correct format
```
vrutkovs: I know its too late, but why we didn't chose json for this? dict of lists for equal preference labels would do the trick, I guess

PR #562, opened 2017-05-03
disable specifying build_imagestream and build_image

### LINKED ISSUE #508: throw exception when in used config is specified both build_imagestream and build_image
that way you will be sure which of them it is using

mlangsdorf:
Fixes #508

### DIFF
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -242,6 +242,10 @@ class BuildSpec(object):
             raise OsbsValidationException("yum_repourls must be a list")
         self.yum_repourls.value = yum_repourls or []
         self.use_auth.value = use_auth
+
+        if build_imagestream and build_image:
+            raise OsbsValidationException(
+                'Please only define build_image -OR- build_imagestream, not both')
         self.build_image.value = build_image or DEFAULT_BUILD_IMAGE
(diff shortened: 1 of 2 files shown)


jarodwilson:
Oops, I meant to comment here, not in the issue... Anyway. For improved user-friendliness, might be worth spitting out what we're getting as the defined build_image and build_imagestream values, i.e.:

`+        if build_image and build_imagestream:`
`+            raise OsbsValidationException(`
`+                    'build_image=%s, build_imagestream=%s. Please only '`
`+                    'define build_image -OR- build_imagestream, not both'`
`+                    % (build_image, build_imagestream))`

Okay, I dunno how to paste code snippets in here correctly. Maybe doing so is frowned upon?... :)

jarodwilson:
Apparently, the "insert code" thing above the input box doesn't behave quite as well as triple-back-tick enclosing what it is you wanted to paste, so here's a better view:

```+        if build_image and build_imagestream:
+                    'build_image=%s, build_imagestream=%s. Please only '
+                    'define build_image -OR- build_imagestream, not both'
+                    % (build_image, build_imagestream))```

vrutkovs:
I think we're fine with current message - this is a configuration error and can be easily solved, so there's no need to make the error message verbose

PR #563, opened 2017-05-04
Make sure BuildRequest has low_priority_node_selector set

vrutkovs:
Related: #556

Since the whole class is flexmocked in tests this has slipped: the class instance 
should store the received low_priority_node_selector

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -85,6 +85,7 @@ class BuildRequest(object):
         :param distribution_scope: str, distribution scope for this image
                                    (private, authoritative-source-only, restricted, public)
         :param use_auth: bool, use auth from atomic-reactor?
+        :param low_priority_node_selector: dict, a nodeselector for builds with lower priority
         """
 
         # Here we cater to the koji "scratch" build type, this will disable
@@ -95,6 +96,7 @@ class BuildRequest(object):
             pass
 
         self.base_image = kwargs.get('base_image')
+        self.low_priority_node_selector = kwargs.get('low_priority_node_selector')
 
         logger.debug("setting params '%s' for %s", kwargs, self.spec)
         self.spec.set_params(**kwargs)
@@ -420,7 +422,6 @@ class BuildRequest(object):
         """
         Enable/disable plugins depending on supported registry API versions
         """
-
         versions = self.spec.registry_api_versions.value
 
         try:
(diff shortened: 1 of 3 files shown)


osbs/build/build_request.py
```diff
@@ -95,6 +95,7 @@ def set_params(self, **kwargs):
             pass
 
         self.base_image = kwargs.get('base_image')
+        self.low_priority_node_selector = kwargs.get('low_priority_node_selector')
```
twaugh: How about updating the docstring?

vrutkovs: Updated

tests/test_api.py
```diff
@@ -1506,6 +1489,69 @@ def verify_build_json(passed_build_json):
...
+                    [general]
+                    build_json_dir = {inputs}
+                    [default]
+                    low_priority_node_selector = {nodeselector}
+                    """.format(inputs=INPUTS_PATH, nodeselector=nodeselector_str)))
+                fp.flush()
+                config = Configuration(fp.name)
+        else:
+            config = Configuration()
+        osbs = OSBS(config, config)
```
vrutkovs: Flake8 doesn't know its being used in `flexmock(osbs.os)` later on, weird

vrutkovs:
Updated the PR - added a new test which checks that `_create_scratch_build` renders a correct Build object with `nodeSelector` set if config has an option for it

tests/test_api.py
```diff
@@ -1506,6 +1489,68 @@ def verify_build_json(passed_build_json):
...
```
lcarva: Oh that's a good catch! @vrutkovs can you address this?

tests/test_api.py
lcarva: Think it's complaining that the `osbs` in line 45 is not used:
```python
45: from tests.fake_api import openshift, osbs, osbs106, osbs_cant_orchestrate
```

vrutkovs: Added `noqa` to this line - these are test fixtures and actually are used.

I've fixed other pep8 errors in this file

vrutkovs:
All flake8 errors in this file have been fixed

lcarva:
LGTM, but is coverall stuck?

vrutkovs:
Coveralls has been stuck for a day for master commits also - https://coveralls.io/stories

Issue #564, opened 2017-05-04
test_render_orchestrate_build unit test fails on EL7 host


jarodwilson:
I've been working on duplicating a devel environment I'd set up on a Fedora guest on a Red Hat Enterprise Linux 7 host, and encountered a problem with unit tests failing on EL7 that pass on Fedora:

```$ py.test -x -vv tests/build_/test_build_request.py::TestBuildRequest::test_render_orchestrate_build
======================================= test session starts ========================================
platform linux2 -- Python 2.7.5, pytest-3.0.7, py-1.4.33, pluggy-0.4.0
rootdir: /src/osbs/osbs-client, inifile:
plugins: capturelog-0.7, cov-2.4.0
collected 209 items

tests/build_/test_build_request.py F

============================================= FAILURES =============================================
 TestBuildRequest.test_render_orchestrate_build[None-1.0.6-additional_kwargs0-fedora:latest-None-fedora:latest-True-1-platforms0-client_config_secret-False]

...
            assert kwargs.get('sources_command') == worker_config.get_sources_command()
            assert kwargs.get('vendor') == worker_config.get_vendor()
>           assert (parse_version(worker_openshift_req_version) ==
                    worker_config.get_openshift_required_version())
E           assert ('00000001', ...06', '*final') == ('*(', '*u', "... ", '*u', ...)
E             At index 0 diff: '00000001' != u'*('
E             Right contains more items, first extra item: "*', "
E             Use -v to get the full diff

tests/build_/test_build_request.py:1242: AssertionError
```
Upon talking with lucarval, this looks to be due to differences in parse_version behavior between the two versions of python:

```# Python 2.7.13 (fedora)
>>> from pkg_resources import parse_version
>>> parse_version('1.0.1')
<Version('1.0.1')>
>>> str(parse_version('1.0.1'))
'1.0.1'


# Python 2.7.5 (rhel 7)
>>> from pkg_resources import parse_version
>>> parse_version('1.0.1')    
('00000001', '00000000', '00000001', '*final')
>>> str(parse_version('1.0.1'))
"('00000001', '00000000', '00000001', '*final')"```

I'm thinking it would be a good idea if we made sure all the unit tests passed on one of the main targets for running the actual code. :)

lcarva:
Problem appears to be that rhel7 still uses version `0.9.8` of setuptools. Fedora has `25.1.1`.
[...]

PR #565, opened 2017-05-04
support old setuptools on el7

jarodwilson:
The ancient setuptools in el7 behaves differently from more recent
setuptools, which results in unit test failures when running on an el7
host.

A quick illustration of the differences:

  # Python 2.7.13 (fedora)
  >>> from pkg_resources import parse_version
  >>> parse_version('1.0.1')
  <Version('1.0.1')>
  >>> str(parse_version('1.0.1'))
  '1.0.1'

  # Python 2.7.5 (rhel 7)
  >>> from pkg_resources import parse_version
  >>> parse_version('1.0.1')
  ('00000001', '00000000', '00000001', '*final')
  >>> str(parse_version('1.0.1'))
  "('00000001', '00000000', '00000001', '*final')"

With this change, all 12 test failures I was previously encountering are
now passing. The key to this change is standardizing the old setuptools
output to look just like current setuptools output, using a new function
sanitize_version() in osbs/util.py, rather than calling parse_version()
directly.

A special nod to
http://www.programcreek.com/python/example/7930/pkg_resources.parse_version
for Example 4, which provides the basis for this work-around.

Resolves: #564

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -26,7 +26,7 @@ from osbs.build.manipulate import DockJsonManipulator
 from osbs.build.spec import BuildSpec
 from osbs.constants import SECRETS_PATH, DEFAULT_OUTER_TEMPLATE, DEFAULT_INNER_TEMPLATE, DEFAULT_CUSTOMIZE_CONF
 from osbs.exceptions import OsbsException, OsbsValidationException
-from osbs.utils import looks_like_git_hash, git_repo_humanish_part_from_uri
+from osbs.utils import looks_like_git_hash, git_repo_humanish_part_from_uri, sanitize_version
 from osbs import __version__ as client_version
 
 
@@ -217,7 +217,7 @@ class BuildRequest(object):
             'info_url_format': self.spec.info_url_format.value,
             'koji_hub': self.spec.kojihub.value,
             'koji_root': self.spec.kojiroot.value,
-            'openshift_required_version': self._openshift_required_version,
+            'openshift_required_version': sanitize_version(self._openshift_required_version),
             'pulp_registry_name': self.spec.pulp_registry.value,
             'registry_api_versions': ','.join(self.spec.registry_api_versions.value or []) or None,
             'smtp_additional_addresses': ','.join(self.spec.smtp_additional_addresses.value or []) or None,
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -549,6 +549,9 @@ class BuildRequest(object):
             if self.spec.yum_repourls.value:
                 self.dj.dock_json_set_arg(phase, plugin, 'repos',
                                           self.spec.yum_repourls.value)
+            if self.spec.platforms.value:
+                self.dj.dock_json_set_arg(phase, plugin, 'architectures',
+                                          self.spec.platforms.value)
 
     def render_add_labels_in_dockerfile(self):
         phase = 'prebuild_plugins'
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -385,6 +385,31 @@ def run_command(*popenargs, **kwargs):
     return output
 
 
+def sanitize_version(version):
+    """
+    Take parse_version() output and standardize output from older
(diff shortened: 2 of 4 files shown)


osbs/build/build_request.py
```diff
@@ -26,7 +25,7 @@
 from osbs.build.spec import BuildSpec
 from osbs.constants import SECRETS_PATH, DEFAULT_OUTER_TEMPLATE, DEFAULT_INNER_TEMPLATE, DEFAULT_CUSTOMIZE_CONF
 from osbs.exceptions import OsbsException, OsbsValidationException
-from osbs.utils import looks_like_git_hash, git_repo_humanish_part_from_uri
+from osbs.utils import looks_like_git_hash, git_repo_humanish_part_from_uri, sanitize_version
```
jarodwilson: I'm told up to 99 chars is allowed in this project, and looks_like_git_hash was already there, not something I'm adding. It could be removed, but that's really not part of this change.

osbs/build/build_request.py
```diff
@@ -57,7 +56,7 @@ def __init__(self, build_json_store, inner_template=None,
         self._customize_conf = None  # site customize conf for _inner_template
         self._dj = None
         self._resource_limits = None
-        self._openshift_required_version = parse_version('1.0.6')
+        self._openshift_required_version = sanitize_version('1.0.6')
```
lcarva: I think we still want the "raw" return value of `parse_version` here. By always sanitizing this value, we lose the main benefit of `parse_version` which is the ability to compare versions. The code right now doesn't actually make use of this value, but it did in the past and will likely use it again in the future.

The only place that we want to sanitize this value is in the `render_orchestrate_build` method because the value is encoded as json. In this case, we need to get back to the raw string value. But anywhere else in the code it should be using whatever `parse_version` returns.

tests/test_utils.py
```diff
@@ -332,3 +333,26 @@ def test_labels(labels, fnc, expect):
...
+    ('bacon', '6.6.6', False, False),
+])
+def test_sanitize_version(v1, v2, match, valid):
+    if valid:
+        if match:
+            assert sanitize_version(parse_version(v1)) == v2
+        else:
+            assert sanitize_version(parse_version(v1)) != v2
+    else:
+        try:
```
lcarva: Why not?
```python
with pytest.raises(TheException):
  sanitize_version(parse_version(v1))
```

tests/test_utils.py
```diff
@@ -332,3 +333,26 @@ def test_labels(labels, fnc, expect):
...
+    ('5.3', '5.3.0', True, True),
+    ('2', '2.0.0', True, True),
+    ('7.3.1', '4.5.9', False, True),
+    ('9', '3.2.1', False, True),
+    ('6.8', '6.8.1', False, True),
+    ('bacon', '6.6.6', False, False),
+])
+def test_sanitize_version(v1, v2, match, valid):
+    if valid:
+        if match:
```
lcarva: Not sure if there's a lot of value in the `match` attribute. This is a simple string comparison.

osbs/utils.py
```diff
@@ -23,6 +23,7 @@
 from datetime import datetime
 from io import BytesIO
 from hashlib import sha256
+from pkg_resources import parse_version
```
lcarva: Let's fix this.

tests/build_/test_build_request.py
```diff
@@ -24,7 +24,6 @@
 
 from flexmock import flexmock
 import pytest
-from pkg_resources import parse_version
```
lcarva: Don't we need this import?

jarodwilson: No, it's actually currently imported twice, that's just removing the second redundant import.

osbs/utils.py
jarodwilson: Argh. Stray relic from earlier renditions, thought I'd removed it.

twaugh:
Submitted https://github.com/projectatomic/osbs-client/pull/584 to fix the stickler diagnostic.

PR #568, opened 2017-05-16
arrangement-version argument for worker build

rcerven:

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -349,7 +349,7 @@ def cmd_build(args, osbs):
     required_args = []
     unwanted_args = []
     if args.worker:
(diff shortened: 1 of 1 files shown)


osbs/cli/main.py
```diff
@@ -696,6 +697,8 @@ def cli():
                               help='platform name to use')
     worker_group.add_argument('--release', action='store', required=False,
                               help='release value to use')
+    worker_group.add_argument('--arrangement-version', action='store', required=False,
+                              help='default version of inner template to use')
```
twaugh: I'd remove 'default' here. There's no default, which is why it's a required argument.

rcerven: ahh ok, I'll remove it

PR #570, opened 2017-05-22
arrangement-version argument for worker build fix

rcerven:

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -30,6 +30,7 @@ from osbs.build.pod_response import PodResponse
 from osbs.build.spec import BuildSpec
 from osbs.exceptions import OsbsValidationException, OsbsException, OsbsResponseException
 from osbs.http import HttpResponse
+from osbs.cli.main import cmd_build
 from osbs.constants import (DEFAULT_OUTER_TEMPLATE, WORKER_OUTER_TEMPLATE,
                             DEFAULT_INNER_TEMPLATE, WORKER_INNER_TEMPLATE,
                             DEFAULT_CUSTOMIZE_CONF, WORKER_CUSTOMIZE_CONF,
@@ -60,6 +61,13 @@ def request_as_response(request):
 
     request.json = request.render()
     return request
+
+
+class CustomTestException(Exception):
+    """
+    Custom Exception used to prematurely end function call
+    """
+    pass
 
 
 class MockDfParser(object):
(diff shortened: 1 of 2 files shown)


twaugh:
Do we also need to do this for `scratch`? That parameter can also be specified in osbs.conf.

rcerven:
well not it passes only not empty value, if noting provided it will read config

twaugh:
OK.

So for testing, one thing is you could write a test for `cmd_build` to check that it at least gets as far as calling the create-build function with the correct arguments. For example, you could mock create_worker_build() and spy on its arguments, as well as making it raise some nonsense exception like StopIteration. Then you could call `cmd_build`, excepting it to raise StopIteration (or whatever).. if you ge that far, the spying worked.

tests/test_api.py
```diff
@@ -1710,3 +1711,78 @@ def test_reactor_config_secret(self):
...
+            'tag': None,
+            'target': None,
+            'architecture':None,
+            'yum_repourls': None,
+        }
+        if worker:
+            expected_kwargs['arrangement_version'] = arrangement_version
+
+        print(expected_kwargs)
+        if worker and not raises_exception:
```
vrutkovs: Better idea would be storing `flexmock` call result and adding `.and_raises` if `raises_exception`

tests/test_api.py
```diff
@@ -1710,3 +1711,78 @@ def test_reactor_config_secret(self):
...
+            (flexmock(osbs)
+                .should_receive("create_worker_build")
+                .times(1)
```
vrutkovs: nitpick: `once()`

tests/test_api.py
```diff
@@ -1710,3 +1711,78 @@ def test_reactor_config_secret(self):
...
+                .with_args(**expected_kwargs)
+                .and_raise(StopIteration))
+
+        if not worker and not orchestrator and not raises_exception:
+            (flexmock(osbs)
+                .should_receive("create_prod_build")
+                .times(1)
+                .with_args(**expected_kwargs)
+                .and_raise(StopIteration))
```
vrutkovs: Why is `StopIteration` being thrown here?

rcerven: I just want to check that it got correct arguments and I don't want it to continue, like twaugh suggested

tests/test_api.py
rcerven: it will always raise for our case, but if it will raise IT won't raise that function but unwanted function

tests/test_api.py
vrutkovs: Oh, okay, that makes sense, although we should use a custom exception instead of `StopIteration` - so that if any function using that wouldn't break the test

tests/test_api.py
```diff
@@ -1710,3 +1715,78 @@ def test_reactor_config_secret(self):
...
+            'user': None,
+            'tag': None,
+            'target': None,
+            'architecture': None,
+            'yum_repourls': None,
+        }
+        if worker:
+            expected_kwargs['arrangement_version'] = arrangement_version
+
+        print(expected_kwargs)
```
twaugh: Don't need this line.

tests/test_api.py
```diff
@@ -62,6 +63,10 @@ def request_as_response(request):
     return request
 
 
+class CustomTestException(Exception):
```
twaugh: Maybe a docstring explaining what this is for and how it is used?

tests/test_api.py
twaugh: Maybe a comment at the top of this block explaining what the `and_raise()` calls are for?

tests/test_api.py
rcerven: added

tests/test_api.py
rcerven: removed

PR #571, opened 2017-05-25
iter_lines: set chunk_size = 1 so it would not wait for buffer to fill in

vrutkovs:

### DIFF
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -6,6 +6,7 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 import logging
+import sys
 
 from flexmock import flexmock
 import pytest
@@ -26,6 +27,11 @@ def s():
 
 def has_connection():
     # In case we run tests in an environment without internet connection.
+
+    if sys.version_info < (2, 7):
+        # py 2.6 doesn't have SNI support, required for httpbin, as it has SSLv3 certificates
+        return False
+
     try:
         HttpStream("https://httpbin.org/get", "get")
         return True
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -147,7 +147,7 @@ class HttpStream(object):
         return self.req.iter_content(None)
 
     def iter_lines(self):
-        return self.req.iter_lines(decode_unicode=True)
+        return self.req.iter_lines(chunk_size=1, decode_unicode=True)
 
     def close(self):
         if not self.closed:
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -147,7 +147,12 @@ class HttpStream(object):
         return self.req.iter_content(None)
 
     def iter_lines(self):
-        return self.req.iter_lines(chunk_size=1, decode_unicode=True)
+        kwargs = {
+            'decode_unicode': True
+        }
+        if requests.__version__.startswith('2.6.'):
+            kwargs['chunk_size'] = 1
+        return self.req.iter_lines(**kwargs)
 
     def close(self):
         if not self.closed:


twaugh:
Maybe include in the commit message that this is needed for the version of python-requests we're using? Then years later we'll know whether it's safe to remove.

vrutkovs:
Good idea. Updated the commit message.

`chunk_size` setting won't be applied if a newer `python-requests` version is used. `tests\http.py` is skipped on py2.6 as it doesn't support SNI, required for httpbin.org

tests/test_http.py
```diff
@@ -26,6 +27,11 @@ def s():
 
 def has_connection():
     # In case we run tests in an environment without internet connection.
+
+    if sys.version_info < (2, 7):
+        # py 2.6 doesn't have SNI support, required for httpbin, as it has SSLv3 certificates
+        return False
```
twaugh: This was only causing a warning, not an error. Is it required?

vrutkovs: Its causing SSLError exception in https://travis-ci.org/projectatomic/osbs-client/jobs/235936576:
```
tests/test_http.py:36: in <module>

    @pytest.mark.skipif(not has_connection(),

tests/test_http.py:30: in has_connection

    HttpStream("https://httpbin.org/get", "get")

osbs/http.py:138: in __init__

    self.req = requests.request(method, url, **args)

../../../virtualenv/python2.6.9/lib/python2.6/site-packages/requests/api.py:58: in request

    return session.request(method=method, url=url, **kwargs)

../../../virtualenv/python2.6.9/lib/python2.6/site-packages/requests/sessions.py:518: in request

    resp = self.send(prep, **send_kwargs)

../../../virtualenv/python2.6.9/lib/python2.6/site-packages/requests/sessions.py:639: in send

    r = adapter.send(request, **kwargs)

../../../virtualenv/python2.6.9/lib/python2.6/site-packages/requests/adapters.py:512: in send

    raise SSLError(e, request=request)

E   SSLError: [Errno 1] _ssl.c:493: error:14077438:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert internal error
```

twaugh: Do we know the SNI warnings are related this exception? They look different.

vrutkovs: Its related to http://docs.python-requests.org/en/master/community/faq/#what-are-hostname-doesn-t-match-errors:

>  If you're certain the server's SSL setup is correct (for example, because you can visit the site with your browser) and you're using Python 2.6 or 2.7, a possible explanation is that you need Server-Name-Indication.

>Server-Name-Indication, or SNI, is an official extension to SSL where the client tells the server what hostname it is contacting. This is important when servers are using Virtual Hosting. When such servers are hosting more than one SSL site they need to be able to return the appropriate certificate based on the hostname the client is connecting to.

>Python3 and Python 2.7.9+ include native support for SNI in their SSL modules

py2.6 has no native SNI support and although there seems to be a way to make it work in py2.6 I'm not sure how to do that :(

twaugh: What I'm saying is: SNI may cause diagnostic output, but does it actually cause tests to fail?

So far I'm not convinced that it does. We weren't seeing "hostname doesn't match" errors -- we were seeing an "internal error".

Could we at least try leaving the tests on for Python 2.6, but with the requests.__version__ check in place?

twaugh: OK, it is definitely needed then. Thanks.

vrutkovs: Reverted that commit, tests still fail: https://travis-ci.org/projectatomic/osbs-client/builds/235968189?utm_source=github_status&utm_medium=notification

>We weren't seeing "hostname doesn't match" errors -- we were seeing an "internal error".

I believe this is hidden in this `SSLError`

PR #572, opened 2017-05-30
core: trap ChunkedEncodingError along with of httplib.IncompleteRead, as python-requests throws two exceptions for different cases

vrutkovs:

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -24,6 +24,7 @@ from osbs.exceptions import (OsbsResponseException, OsbsException,
                              OsbsWatchBuildNotFound, OsbsAuthException)
 from osbs.http import decoded_json
 from osbs.utils import graceful_chain_get
+from requests.exceptions import ChunkedEncodingError
 
 try:
     # py2
@@ -436,7 +437,7 @@ class Openshift(object):
                 for line in decoded_json(response.iter_lines()):
                     last_activity = time.time()
                     yield line
-            except httplib.IncompleteRead:
+            except (ChunkedEncodingError, httplib.IncompleteRead):
                 pass
 
             idle = time.time() - last_activity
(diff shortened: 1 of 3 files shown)


tests/test_core.py
```diff
@@ -91,7 +92,7 @@ def test_stream_logs(self, openshift):
         (flexmock(openshift)
             .should_receive('_get')
              # First: timeout in response after 100s
-            .and_raise(httplib.IncompleteRead(''))
+            .and_raise(ChunkedEncodingError(''))
```
twaugh: I actually think we should have two test cases:
* some data is sent, then httplib.IncompleteRead is raised
* no data is sent, then ChunkedEncodingError is raised

I believe this is what's happening. When data has been sent but the remote endpoint closes the connection, that's when we get httplib.IncompleteRead. However if no data has been sent but the remote endpoint closes the connection, that's technically an error because the headers said to expect data -- so we get ChunkedEncodingError.

osbs/core.py
```diff
@@ -436,7 +437,7 @@ def stream_logs(self, build_id):
                 for line in decoded_json(response.iter_lines()):
                     last_activity = time.time()
                     yield line
-            except httplib.IncompleteRead:
+            except ChunkedEncodingError:
```
twaugh: I think we should catch both of these.

vrutkovs: Fixed, we now have two tests for this

tests/test_core.py
```diff
@@ -90,9 +91,31 @@ def test_stream_logs(self, openshift):
...
+
+        (flexmock(time)
+            .should_receive('time')
+            .and_return(0)
+            .and_return(100))
+
+        logs = openshift.stream_logs(TEST_BUILD)
+        assert len([log for log in logs]) == 1
+
+    def test_stream_logs_no_data(self, openshift):  # noqa
```
twaugh: Names of the tests are the wrong way around?

vrutkovs: Ah, correct, I was thinking 'no data after a timeout', but it would be too long. Fixed by swapping test names

rcerven:
change title to reflect the latest change

PR #574, opened 2017-05-30
core: trap ConnectionError and handle it like ChunkedEncodingError

twaugh:
This is to handle BadStatusLine-caused ConnectionErrors.

### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -45,7 +45,9 @@ def decoded_json(iterable):
                 yield line
             else:
                 yield line.decode(requests.utils.guess_json_utf(line))
-    except (requests.exceptions.ChunkedEncodingError, httplib.IncompleteRead):
+    except (requests.exceptions.ChunkedEncodingError,
+            requests.exceptions.ConnectionError,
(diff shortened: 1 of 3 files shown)


osbs/http.py
```diff
@@ -165,7 +165,9 @@ def iter_lines(self):
             kwargs['chunk_size'] = 1
         try:
             return self.req.iter_lines(**kwargs)
-        except (requests.exceptions.ChunkedEncodingError, httplib.IncompleteRead):
+        except (requests.exceptions.ChunkedEncodingError,
```
lcarva: Do we also need to do this in `decoded_json` ?

vrutkovs: It probably won't be extra, yes, although I'm not sure if `ConnectionError` is possible during `iter_lines` - it would most likely be wrapped in `ChunkedEncodingError`

twaugh: I don't think so -- decoded_json() is always called on an iterable, and this is a response.iter_lines(). The error handling is in there now.

So in theory decoded_json() doesn't need to handle any of these.

PR #575, opened 2017-05-30
Handle ConnectionError via OsbsException

lcarva:
The unit tests in `test_core.py` don't leverage the class HttpSession in http.py.
Instead they use the `Connection` class from fake_api.py.
Added additional tests in test_http.py to properly simulate behavior.

The except block in HttpStream.iter_lines didn't seem very useful.
When self.req.iter_lines is called, it returns a generator.
This wouldn't raise CheckedEncodingError, ConnectionError, nor IncompleteRead.
Those error cases are only valid when actually iterating through the generator, not during its creation.
The method decoded_json seemed to handle this properly. 
However, given that decoded_json is always used with iter_lines, I moved the correct error handling out of decoded_json and into iter_lines.

The errors we were seeing is because by the time the exception propagates to Openshift class in core.py, it's been wrapped in either OsbsException or OsbsNetworkException. We should catch those exceptions instead and inspect its cause for `ConnectionError`.

I built a simple hello world image to perform basic sanity checks, but further integration tests should be performed.

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -21,7 +21,8 @@ from osbs.constants import WATCH_MODIFIED, WATCH_DELETED, WATCH_ERROR
 from osbs.constants import (SERVICEACCOUNT_SECRET, SERVICEACCOUNT_TOKEN,
                             SERVICEACCOUNT_CACRT)
 from osbs.exceptions import (OsbsResponseException, OsbsException,
-                             OsbsWatchBuildNotFound, OsbsAuthException)
+                             OsbsWatchBuildNotFound, OsbsAuthException,
+                             OsbsNetworkException)
 from osbs.http import decoded_json
 from osbs.utils import graceful_chain_get
 from requests.exceptions import ChunkedEncodingError, ConnectionError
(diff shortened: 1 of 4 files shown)


osbs/core.py
```diff
@@ -437,10 +438,18 @@ def stream_logs(self, build_id):
...
-                pass
+            # NOTE1: If self._get causes ChunkedEncodingError, ConnectionError,
+            # or IncompleteRead to be raised, they'll be wrapped in
+            # OsbsNetworkException or OsbsException
+            # NOTE2: If decode_json or iter_lines causes ChunkedEncodingError, ConnectionError,
+            # or IncompleteRead to be raised, it'll simply be silenced.
+            # NOTE3: An error may occur in iter_lines during initial contact
+            # with server - prior to streaming data. In this case, exception will be
+            # wrapped in OsbsException or OsbsNetworkException, inspect cause
+            # to detect ConnectionError.
```
lcarva: I left these comments here in order to clarify why this change was done. Not sure if they should stay permanently.

twaugh: As you point out in note 2,  `iter_lines` silences ConnectionError, so the only thing that can raise OsbsException is `check_response`. But it can and will.

lcarva: I hadn't taken that into account. But I think the logic still holds true.
I can remove the notes if that's going to cause confusion later on.

twaugh:
Works in my local testing:
- follow a build that is forever pending (on a system that would cause chunk-encoding-error)
- follow a build that emits no logs for 20m during the build (on a system that would cause bad-status-line)

vrutkovs:
Integration tests passed

PR #577, opened 2017-06-01
random seed before every random

### LINKED ISSUE #483: scratch builds should have a short random string added to the end of the build name

rcerven:

fixes #483

### DIFF
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -305,6 +305,11 @@ class BuildSpec(object):
             logger.debug("setting 'imagestream_insecure_registry' to %r", insecure)
 
         timestamp = utcnow().strftime('%Y%m%d%H%M%S')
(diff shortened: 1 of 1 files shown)


vrutkovs:
Looks good, testing a similar commit meanwhile.

We're gonna need to add a comment about brew builder behaviour

vrutkovs:
A better comment:
```
RNG is seeded once its imported, so in cli calls scratch builds would get unique name. On brew builders we import osbs once - thus RNG is seeded once and `randrange` returns the same values throughout the life of the builder.

Before each `randrange` call we should be calling `.seed` to prevent this
```

PR #579, opened 2017-06-05
get node_selector.platform values from conf

### LINKED ISSUE #578: osbs.conf handling for node_selector.* keyword
When parsing cluster instance configs, keywords beginning "node_selector." should be interpreted as node_selector.platform and the value is a string or "none" for no value.

mlangsdorf:
Fixes #578

if platform exists at the time of build_request.set_params, read the
conf file for a variable node_selector.platform and return its value.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -505,7 +505,8 @@ class OSBS(object):
             info_url_format=self.build_conf.get_info_url_format(),
             artifacts_allowed_domains=self.build_conf.get_artifacts_allowed_domains(),
             low_priority_node_selector=self.build_conf.get_low_priority_node_selector(),
-            equal_labels=self.build_conf.get_equal_labels()
+            equal_labels=self.build_conf.get_equal_labels(),
+            platform_node_selector=self.build_conf.get_platform_node_selector(platform)
         )
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
         if build_request.scratch:
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -532,3 +532,19 @@ class Configuration(object):
                 raise OsbsValidationException("Wrong equal_labels configuration")
 
         return equal_labels
+
+    def get_platform_node_selector(self, platform):
+        """
+        search the configuration for entries of the form node_selector.platform
+        :return None or str
+
+        :param platform: str, platform to search for, can be null
+        """
+        nodeselector = None
+        if platform:
+            nodeselector_str = self._get_value("node_selector." + platform, self.conf_section,
+                                               "node_selector." + platform)
+            if nodeselector_str:
+                nodeselector = nodeselector_str
+
+        return nodeselector
--- a/tests/test_conf.py
+++ b/tests/test_conf.py
@@ -399,3 +399,34 @@ class TestConfiguration(object):
                     conf.get_equal_labels()
             else:
                 assert conf.get_equal_labels() == expected
+
+    @pytest.mark.parametrize(('platform', 'config', 'kwargs', 'expected'), [
+        ('',
+         {},
+         {},
+         None),
+        ('',
+         {'default': {'node_selector.dinner': 'steak.com'}},
+         {},
+         None),
+        ('',
+         {'default': {}},
+         {'node_selector.dinner': 'steak.com'},
+         None),
+        ('breakfast',
+         {'default': {'node_selector.breakfast': 'eggs.com', 'node_selector.dinner': 'steak.com'}},
+         {},
+         'eggs.com'),
+        ('breakfast',
+         {'default': {}},
+         {'node_selector.breakfast': 'eggs.com', 'node_selector.dinner': 'steak.com'},
+         'eggs.com'),
+        ('breakfast',
+         {'default': {'node_selector.breakfast': 'bacon.com'}},
+         {'node_selector.breakfast': 'eggs.com', 'node_selector.dinner': 'steak.com'},
+         'eggs.com'),
+    ])
+    def test_get_node_selector_platform(self, platform, kwargs, config, expected):
+        with self.config_file(config) as config_file:
+            conf = Configuration(conf_file=config_file, **kwargs)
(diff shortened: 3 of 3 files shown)


osbs/conf.py
```diff
@@ -532,3 +532,26 @@ def get_equal_labels(self):
...
+            for key, val in self.args:
+                if key.startswith('node_selector.'):
+                    assign_platform_to_key(key, val)
+        try:
+            conf_items = self.scp.items(self.conf_section)
+            for key, val in conf_items:
+                if key.startswith('node_selector.'):
+                    assign_platform_to_key(key, val)
+        except configparser.Error:
+            return nodeselector
```
twaugh: I think we want kwargs/CLI args to override configuration values, so is this the wrong way around or should assign_platform_to_key() be using .setdefault() instead of assigning?

osbs/api.py
```diff
@@ -505,7 +505,8 @@ def _do_create_prod_build(self, git_uri, git_ref,
             info_url_format=self.build_conf.get_info_url_format(),
             artifacts_allowed_domains=self.build_conf.get_artifacts_allowed_domains(),
             low_priority_node_selector=self.build_conf.get_low_priority_node_selector(),
-            equal_labels=self.build_conf.get_equal_labels()
+            equal_labels=self.build_conf.get_equal_labels(),
+            node_selector_platforms=self.build_conf.get_node_selector_platform()
```
twaugh: We might want to do this inside `create_worker_build()`, and pass through the node selector for the specified platform.

tests/test_conf.py
```diff
@@ -401,3 +401,19 @@ def test_get_equal_labels(self, config, expected):
...
+         {'breakfast': 'egg.com', 'lunch': 'spam.com'}),
+        ({'default': {'node_selector.breakfast': 'egg.com',
+                      'node_selector.lunch': 'spam.com',
+                      'node_selector.dinner': 'spam.com'}},
+         {'breakfast': 'egg.com', 'lunch': 'spam.com', 'dinner': 'spam.com'}),
+    ])
+    def test_get_node_selector_platform(self, config, expected):
+        with self.config_file(config) as config_file:
+            conf = Configuration(conf_file=config_file)
+            assert conf.get_node_selector_platform() == expected
```
twaugh: Would be nice to have a test showing that kwargs > CLI args > conf values.

osbs/conf.py
mlangsdorf: wrong way around; fixed.

tests/test_conf.py
mlangsdorf: test added. dropped support for cli_args in the process

osbs/api.py
mlangsdorf: Has to be passed in to build_request in _do_create_prod_build() because that's were build_request params are set.

I addressed this by altering get_node_selector_platform to take a platforms argument, and only returning node_selectors.platform whose platform matches one of the ones in platforms.

osbs/api.py
```diff
@@ -505,7 +505,8 @@ def _do_create_prod_build(self, git_uri, git_ref,
             info_url_format=self.build_conf.get_info_url_format(),
             artifacts_allowed_domains=self.build_conf.get_artifacts_allowed_domains(),
             low_priority_node_selector=self.build_conf.get_low_priority_node_selector(),
-            equal_labels=self.build_conf.get_equal_labels()
+            equal_labels=self.build_conf.get_equal_labels(),
+            node_selector_platform=self.build_conf.get_node_selector_platform(platforms)
```
twaugh: This should be `(platform)`, not `(platforms)`. Also can we call the parameter `platform_node_selector`, for comparison with `low_priority_node_selector` above it.

osbs/conf.py
```diff
@@ -532,3 +532,27 @@ def get_equal_labels(self):
                 raise OsbsValidationException("Wrong equal_labels configuration")
 
         return equal_labels
+
+    def get_node_selector_platform(self, platforms):
```
twaugh: Can we have a docstring explaining why this takes a parameter i.e. it returns the node selector for the given platform. Also, it should take a `platform` (singular).

mlangsdorf: Fixed. Code substantially revised with the easier API.

osbs/api.py

osbs/conf.py
```diff
@@ -532,3 +532,18 @@ def get_equal_labels(self):
                 raise OsbsValidationException("Wrong equal_labels configuration")
 
         return equal_labels
+
+    def get_platform_node_selector(self, platform):
+        """
+        search the configuration for entries of the form node_selector.platform
+
+        :param platform: str, platform to search for, can be null
```
twaugh: How about `can be None`?

Do we want to document the return value too?

osbs/conf.py
```diff
@@ -532,3 +532,18 @@ def get_equal_labels(self):
...
+        """
+        nodeselector = ""
```
twaugh: It might be better to use `None` here.

osbs/api.py
```diff
@@ -505,7 +505,8 @@ def _do_create_prod_build(self, git_uri, git_ref,
             info_url_format=self.build_conf.get_info_url_format(),
             artifacts_allowed_domains=self.build_conf.get_artifacts_allowed_domains(),
             low_priority_node_selector=self.build_conf.get_low_priority_node_selector(),
-            equal_labels=self.build_conf.get_equal_labels()
+            equal_labels=self.build_conf.get_equal_labels(),
+            platform_node_selector=self.build_conf.get_platform_node_selector(platform)
```
twaugh: Just a note: is the intent of this PR to also use this value adjust the BuildConfig, or are you planning a separate PR for that?

mlangsdorf: separate PR for that.

osbs/conf.py
mlangsdorf: Done.

osbs/conf.py
mlangsdorf: Done.

PR #580, opened 2017-06-06
test_conf.py: don't mock argparse.Namespace with namedtuples as the latter don't support dots in names

vrutkovs:
This also removed unused imports of os and namedtuple

Related to #579


twaugh:
Only improvement I'd suggest is to shorten the commit message summary line and put the explanation in the rest of the commit message.
```
Use argparse.Namespace for args testing

Don't mock argparse.Namespace with namedtuples as the latter don't support dots 
in names.
```

vrutkovs:
Good idea, updated the commit message

PR #581, opened 2017-06-07
build_request: Set 'release' label when provided

dzickusrh:
[Looking for early feedback]

Currently if the spec provides a 'release' param, it is not passed along
to the orchestrator as a label.  Detect that and provide it.  Also remove
the bump_release plugin if the 'release' param is passed in.

Updated tests to verify change.

Signed-off-by: Don Zickus <dzickus@redhat.com>

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -562,6 +562,7 @@ class BuildRequest(object):
             'vendor': self.spec.vendor,
             'authoritative-source-url': self.spec.authoritative_registry,
             'distribution-scope': self.spec.distribution_scope,
+            'release': self.spec.release,
         }
 
         for label, spec in label_spec.items():
@@ -614,6 +615,12 @@ class BuildRequest(object):
         phase = 'prebuild_plugins'
         plugin = 'bump_release'
         if not self.dj.dock_json_has_plugin_conf(phase, plugin):
+            return
+
+        if self.spec.release.value:
+            logger.info('removing %s from request as release already specified',
+                        plugin)
+            self.dj.remove_plugin(phase, plugin)
             return
 
         hub = self.spec.kojihub.value
--- a/tests/build_/test_build_request.py
+++ b/tests/build_/test_build_request.py
@@ -345,6 +345,7 @@ class TestBuildRequest(object):
         assert labels['authoritative-source-url'] == authoritative_registry
         assert labels['vendor'] == vendor
         assert labels['distribution-scope'] == distribution_scope
+        assert 'release' not in labels
 
         rendered_build_image = build_json["spec"]["strategy"]["customStrategy"]["from"]["name"]
         if not build_imagestream:
@@ -456,6 +457,7 @@ class TestBuildRequest(object):
         assert labels['authoritative-source-url'] is not None
         assert labels['vendor'] is not None
         assert labels['distribution-scope'] is not None
+        assert 'release' not in labels
 
     def test_render_prod_without_koji_request(self):
         build_request = BuildRequest(INPUTS_PATH)
@@ -534,6 +536,7 @@ class TestBuildRequest(object):
         assert labels['authoritative-source-url'] is not None
         assert labels['vendor'] is not None
         assert labels['distribution-scope'] is not None
+        assert 'release' not in labels
 
     def test_render_prod_with_secret_request(self):
         build_request = BuildRequest(INPUTS_PATH)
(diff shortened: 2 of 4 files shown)


dzickusrh:
@lcarva 
"I think we want to only set this key if self.spec.release is truthy."

What key?  :-(  The 'release' key for labels?

osbs/build/build_request.py
```diff
@@ -562,6 +562,7 @@ def render_add_labels_in_dockerfile(self):
             'vendor': self.spec.vendor,
             'authoritative-source-url': self.spec.authoritative_registry,
             'distribution-scope': self.spec.distribution_scope,
+            'release': self.spec.release,
```
twaugh: We shouldn't set `'release'` here unless `self.spec.release` is truthy. In other words, we don't want to override the release label with a value unless we have a value to override it with.

lcarva: @dzickusrh, this key.

I had commented on the commit itself before you created the PR. I believe that's why my comment was no longer attached a specific code line. Anyways, let's track the discussion in Tim's comment now.

dzickusrh: @twaugh so what you said makes sense, you don't want to push 'release':None to the orchestrator.  However, isn't that already filtered by the code as it translates 'labels' into 'implicti_labels' as shown below?

        implicit_labels = {}
        label_spec = {
            'vendor': self.spec.vendor,
            'authoritative-source-url': self.spec.authoritative_registry,
            'distribution-scope': self.spec.distribution_scope,
            'release': self.spec.release,

>>     I blindly add 'release' above, which is the red flag you have
        }

        for label, spec in label_spec.items():
            if spec.value is not None:
                implicit_labels[label] = spec.value

>>> but the above for-loop checks for None and prevents it from being added to implicit_labels if it is None, correct?

        self.dj.dock_json_merge_arg(phase, plugin, 'labels', implicit_labels)

>>> the resulting filtered 'implicit_labels' is passed here. 
>>> So I think everything is fine, no?

lcarva: That makes sense to me. I completely missed that follow up for loop.

tests/build_/test_build_request.py
```diff
@@ -998,19 +1003,37 @@ def test_render_bump_release(self, hub, disabled):
...
         plugins = get_plugins_from_build_json(build_json)
 
-        if disabled:
-            with pytest.raises(NoSuchPluginException):
-                get_plugin(plugins, "prebuild_plugins", "bump_release")
+        labels = plugin_value_get(plugins, "prebuild_plugins", "add_labels_in_dockerfile",
+                                  "args", "labels")
 
-        else:
+        if not disabled and not release :
```
lcarva: Can you fix this?

dzickusrh:
@lcarva I had a concern about this snippet in my test case update

-        if disabled:
-            with pytest.raises(NoSuchPluginException):
-                get_plugin(plugins, "prebuild_plugins", "bump_release")
+        labels = plugin_value_get(plugins, "prebuild_plugins", "add_labels_in_dockerfile",
+                                  "args", "labels")
 
-        else:
+        if not disabled and not release:
             assert plugin_value_get(plugins, "prebuild_plugins", "bump_release",
                                     "args", "hub") == hub
+            assert 'release' not in labels
+
+        elif not disabled and release:
+            with pytest.raises(NoSuchPluginException):
+                get_plugin(plugins, "prebuild_plugins", "bump_release")
+            assert 'release' in labels
+
+        elif disabled and not release:
+            with pytest.raises(NoSuchPluginException):
+                get_plugin(plugins, "prebuild_plugins", "bump_release")
+            assert 'release' not in labels
+
+        elif disabled and release:
+            with pytest.raises(NoSuchPluginException):
+                get_plugin(plugins, "prebuild_plugins", "bump_release")
+            assert 'release' in labels

>>> The last 'elif' concerns me.  I read it as 'if the hub is disabled but the release is there. then expect the release'.  I _thought_ if the 'hub/secrets/whatever' is disabled then the 'release' code would not be traversed because the bump_version is not added.  Otherwise if you _can_ add a release in the spec and have it auto-added, then my asserts in the early part of my patch should have failed.

What is the expectation of 'release'.  If it is provided _always_ use it? Or only on production builds?

lcarva:
> What is the expectation of 'release'. If it is provided always use it? Or only on production builds?

For an orchestrator build there are two places the release value can come from. 1) user has specified it in the Dockerfile via the 'release' label; or 2) user has omitted 'release' label in Dockerfile causing the `bump_release` plugin to run and determine what this value should be.

By the time a worker build is to be created by the orchestrator build, the release value has already been determined. Whether the value came from Dockerfile label or from bump_release plugin is irrelevant.

For a worker build, the expected behavior with this changes is that the release value is always given when build is created.

But I think your question is regarding what happens if someone creates an orchestrator build with a release value. I think it makes sense to use the value and propagate it down to the worker builds. We can probably achieve that by simply enabling `add_labels_in_dockerfile` in `orchestrator_inner:1.json`. On the other hand, we can also restrict this behavior.

@twaugh, what do you think?

twaugh:
Yes, maybe running `add_labels_in_dockerfile` is the answer to that.

PR #582, opened 2017-06-08
http: remove decoded_json after python-requests merge

vrutkovs:

### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -43,7 +43,7 @@ def decoded_json(iterable):
         if isinstance(line, str):
             yield line
         else:
-            yield line.decode(requests.utils.guess_json_utf(line))
+            yield line.encode(requests.utils.guess_json_utf(line))
 
 
 class HttpSession(object):
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -33,19 +33,6 @@ except ImportError:
 logger = logging.getLogger(__name__)
 
 
-def decoded_json(iterable):
-    """Return str objects for each JSON string in iterable
-
-    See https://tools.ietf.org/html/rfc4627#section-3
-    """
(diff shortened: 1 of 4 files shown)


osbs/http.py
```diff
@@ -43,7 +43,7 @@ def decoded_json(iterable):
         if isinstance(line, str):
             yield line
         else:
-            yield line.decode(requests.utils.guess_json_utf(line))
+            yield line.encode(requests.utils.guess_json_utf(line))
```
lcarva: Should we rename this method? `decoded_json` seems quite misleading.

twaugh: Hmm, I see what you mean.

What it aims to do is examine the first 4 bytes of `line`, which is presumed to be the raw bytes of the reply, and work out the encoding that has been used. Then, it should decode it to a unicode object (just 'str' in Python 3).

Is `line` not what we think?

vrutkovs: I think the question is to leave method name as `decoded_json` or rename it to `encoded_json`.

I don't have a preference, I'd leave it as is, meaning 'decode from whatever there is - a str in case of py3 or encoded to unicode in py2'

lcarva: I'm wondering if we need the encode at all...
`requests.utils.guess_json_utf` will return the current encoding for the line.
Why encode it again?

twaugh: If python-requests is already returning a decoded response, we don't need decoded_json() at all. Can we confirm that's the case?

vrutkovs: Right, we might not need it at all. Pushed a commit to remove it

vrutkovs:
Removed `decoded_json` completely, added a test to verify that Unicode in the build logs doesn't crash `stream_logs`.

Not sure why coveralls reports coverage decrease though

PR #583, opened 2017-06-08
allow source-registry-uri to be set from CLI

lcarva:


csomh:
Would be nice to have some explanation (the answer to why?) in the commit message. Which are the use cases you think this change is important?

twaugh:
Agreed. Will approve once commit message explains why (although I think I get it).

lcarva:
Updated commit message:
```
In some cases, it's beneficial to use a different source
registry for pulling parent image than the one defined
in osbs.conf. For instance, when building in a test
environment, but needing to use parent image from
production environment.
```

PR #585, opened 2017-06-09
[WIP] core.py: encode utf-8 when streaming logs

vrutkovs:

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -494,7 +494,12 @@ class Openshift(object):
         buildlogs_url = self._build_url("builds/%s/log/" % build_id)
         response = self._get(buildlogs_url, headers={'Connection': 'close'})
         check_response(response)
-        return response.content
+
+        content = response.content
+        if isinstance(content, bytes):
+            return content
+        else:
+            return content.encode('utf-8')
 
     def list_builds(self, build_config_id=None, koji_task_id=None,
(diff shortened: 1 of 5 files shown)


osbs/core.py
```diff
@@ -494,7 +494,12 @@ def logs(self, build_id, follow=False, build_json=None, wait_if_missing=False):
         buildlogs_url = self._build_url("builds/%s/log/" % build_id)
         response = self._get(buildlogs_url, headers={'Connection': 'close'})
         check_response(response)
-        return response.content
+
+        content = response.content
+        if isinstance(content, bytes):
+            return content
+        else:
+            return content.encode('utf-8')
```
twaugh: Why do we need to encode it this early? We shouldn't encode it until we need to serialize it.

lcarva:
I think reverting https://github.com/projectatomic/osbs-client/commit/8182bfdef54f0968c950c27503afe6227b5bfa33 should do the trick.

twaugh:
Yes. iter_lines() returns a byte sequence:
http://docs.python-requests.org/en/master/user/advanced/?highlight=iter_lines#streaming-requests

so we need to discover its encoding (it's JSON so we can do that automatically) and decode it.

vrutkovs:
It doesn't work on Py3 and there are alternative solutions. Closing

PR #586, opened 2017-06-09
add node_selector.platform to buildConfig

mlangsdorf:
if platform has a value and node_selector.platform exists, add it to the
buildConfig in nodeSelector. Make sure that subsequent attempts to adjust
nodeSelector via low_priority_node_selector preserve any existing values.


osbs/api.py
```diff
@@ -292,7 +292,9 @@ def _create_scratch_build(self, build_request):
         build_json['metadata']['labels']['scratch'] = 'true'
 
         if build_request.low_priority_node_selector:
-            build_json['spec']['nodeSelector'] = build_request.low_priority_node_selector
+            temp = build_json['spec']['nodeSelector'].copy()
+            temp.update(build_request.low_priority_node_selector)
+            build_json['spec']['nodeSelector'] = temp
```
twaugh: Can probably `.update` this in-place can't we?

tests/test_api.py
```diff
@@ -979,6 +979,44 @@ def test_build_image(self):
...
+                [default]
+                openshift_url = /
+                sources_command = /bin/true
+                vendor = Example, Inc
+                registry_uri = registry.example.com
+                build_host = localhost
+                authoritative_registry = localhost
+                distribution_scope = private
+                platform = breakfast
+                node_selector.breakfast = bacon.com
```
twaugh: A node selector is a set of key=value pairs, like `node_selector.breakfast = spam=eggs,ham=green`, and expect to have a node selector with key 'spam' having value 'eggs' and key 'ham' having value 'green'.

osbs/api.py
mlangsdorf: I think so.

tests/test_api.py
mlangsdorf: okay, changed the tests to represent that.

osbs/build/build_request.py
```diff
@@ -86,6 +88,7 @@ def set_params(self, **kwargs):
                                    (private, authoritative-source-only, restricted, public)
         :param use_auth: bool, use auth from atomic-reactor?
         :param low_priority_node_selector: dict, a nodeselector for builds with lower priority
+        :param platform_node_selector: dict, a nodeselector for builds with lower priority
```
twaugh: Should be "for a specific platform".

Do we set `self.platform_node_selector` to the value that is passed in?

tests/test_api.py
```diff
@@ -979,6 +979,44 @@ def test_build_image(self):
...
+            .and_return(MockDfParser()))
+
+        flexmock(OSBS, _create_build_config_and_build=request_as_response)
+
+        req = osbs_obj.create_prod_build(TEST_GIT_URI, TEST_GIT_REF,
+                                         TEST_GIT_BRANCH, TEST_USER,
+                                         TEST_COMPONENT, TEST_TARGET,
+                                         TEST_ARCH)
+        img = req.json['spec']['strategy']['customStrategy']['from']['name']
+        assert img == build_image
```
twaugh: Also `assert` about spec.nodeSelector?

tests/test_api.py
```diff
@@ -1542,13 +1580,31 @@ def verify_build_json(passed_build_json):
         build_response = osbs_obj._create_scratch_build(build_request)
         assert build_response.json == {'spam': 'maps'}
 
-    @pytest.mark.parametrize(('nodeselector_str, nodeselector_dict'), [
-        (None, {}),
-        ('foo=  bar', {'foo': 'bar'}),
-        ('foo=bar ,  baz=foo', {'foo': 'bar', 'baz': 'foo'}),
+    @pytest.mark.parametrize(('platform, platform_node, nodeselector_str, nodeselector_dict'), [
```
twaugh: Better names for `platform_node` and `nodeselector_str` would now be `platform_nodeselector` and `lowpriority_nodeselector`.

tests/test_api.py
```diff
@@ -1542,13 +1580,31 @@ def verify_build_json(passed_build_json):
...
+    def test_scratch_build_nodeselector(self, platform, platform_node, nodeselector_str,
+                                        nodeselector_dict):
+        if platform and nodeselector_str:
+            with NamedTemporaryFile(mode='wt') as fp:
+                fp.write(dedent("""\
+                    [general]
+                    build_json_dir = {inputs}
+                    [default]
+                    platform = platform_str
+                    node_selector.platform_str = platform_node_str
```
twaugh: Don't forget to use `{...}` to get these value to be substituted.

tests/test_api.py
```diff
@@ -1575,7 +1643,7 @@ def test_scratch_build_nodeselector(self, nodeselector_str, nodeselector_dict):
         updated_build_json['metadata']['labels']['scratch'] = 'true'
         updated_build_json['spec']['serviceAccount'] = 'builder'
         if nodeselector_dict:
-            updated_build_json['spec']['nodeSelector'] = nodeselector_dict
+            updated_build_json['spec']['nodeSelector'].update(nodeselector_dict)
```
twaugh: We also need to make sure we're checking that the platform node selector is actually used.

osbs/build/build_request.py
```diff
@@ -61,6 +61,8 @@ def __init__(self, build_json_store, inner_template=None,
         # For the koji "scratch" build type
         self.scratch = None
         self.base_image = None
+        # necessary to keep stuff from blowing up
```
twaugh: Is this comment helpful? ;-)

tests/test_api.py
mlangsdorf: Thanks! I knew these tests were dubious. okay, asserts added, stuff changed to make the asserts pass.

tests/test_api.py
mlangsdorf: thanks, fixed.

tests/test_api.py
mlangsdorf: fixed

osbs/build/build_request.py
mlangsdorf: fixed

osbs/build/build_request.py
mlangsdorf: clarified that it's a forward reference.

tests/test_api.py
mlangsdorf: names changed to fit in the line width

osbs/conf.py
```diff
@@ -545,6 +545,10 @@ def get_platform_node_selector(self, platform):
             nodeselector_str = self._get_value("node_selector." + platform, self.conf_section,
                                                "node_selector." + platform)
             if nodeselector_str:
-                nodeselector = nodeselector_str
+                constraints = nodeselector_str.split(',')
+                raw_nodeselector = dict([constraint.split('=', 1) for constraint in constraints])
+                nodeselector = dict([k.strip(), v.strip()] for (k, v) in raw_nodeselector.items())
+#                nodeselector = nodeselector_str
```
mlangsdorf: Fixed

tests/test_api.py
```diff
@@ -1542,20 +1592,32 @@ def verify_build_json(passed_build_json):
         build_response = osbs_obj._create_scratch_build(build_request)
         assert build_response.json == {'spam': 'maps'}
 
-    @pytest.mark.parametrize(('nodeselector_str, nodeselector_dict'), [
-        (None, {}),
-        ('foo=  bar', {'foo': 'bar'}),
-        ('foo=bar ,  baz=foo', {'foo': 'bar', 'baz': 'foo'}),
+    @pytest.mark.parametrize(('platform, platform_node, lowpriority_node, lowpriority_node_dict'), [
```
mlangsdorf: fixed

tests/test_api.py
```diff
@@ -1586,6 +1651,8 @@ def verify_build_json(passed_build_json):
             del updated_build_json['metadata']['name']
 
             assert passed_build_json == updated_build_json
+            if platform or lowpriority_node:
+                assert passed_build_json['spec']['nodeSelector']
```
mlangsdorf: fixed.

osbs/build/build_request.py
```diff
@@ -95,6 +98,9 @@ def set_params(self, **kwargs):
 
         self.base_image = kwargs.get('base_image')
         self.low_priority_node_selector = kwargs.get('low_priority_node_selector')
+        if kwargs.get('platform'):
+            self.platform = kwargs.get('platform')
```
vrutkovs: Nitpick: it doesn't look like this var being used anywhere, do we need it?

mlangsdorf: you need to set platform or else get_platform_nodeselector can't search against it.

vrutkovs: Ah, got it

twaugh: Agreed. This should be:
```
self.platform_node_selector = kwargs.get('platform_node_selector')
```
(i.e. remove the `kwargs.get('platform')` check)

The `platform` keyword is only needed to work out the node selector for that platform.

twaugh: You could squash this commit in if you like: https://github.com/twaugh/osbs-client/commit/d2f33aea6c0d73bdfb8adf645101643b40c09106

mlangsdorf:
Rebased to resolve the conflict.

osbs/build/build_request.py
```diff
@@ -96,6 +99,7 @@ def set_params(self, **kwargs):
 
         self.base_image = kwargs.get('base_image')
         self.low_priority_node_selector = kwargs.get('low_priority_node_selector')
+        self.platform_node_selector = kwargs.get('platform_node_selector')
```
rcerven: if you do 

self.platform_node_selector = kwargs.get('platform_node_selector', {})

you don't need the if,else on 1101

mlangsdorf: Cool, I did not know that.
Fixed.

osbs/api.py
```diff
@@ -475,7 +475,6 @@ def _do_create_prod_build(self, git_uri, git_ref,
             koji_kerberos_keytab=self.build_conf.get_koji_kerberos_keytab(),
             koji_kerberos_principal=self.build_conf.get_koji_kerberos_principal(),
             architecture=architecture,
-            platform=platform,
```
lcarva: Don't we need this?

twaugh: No, this is in the set_params() call where we already set the platform node selector. We don't need the platform for anything else.

lcarva: It's a required param for the `create_worker_build` api method which is called by atomic-reactor's `orchestrate_build` plugin: https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/build_orchestrate_build.py#L203

rcerven: also used in reactor config

twaugh: `create_worker_build()` requires it, yes, and passes it to `_do_create_prod_build()`. It requires it because the build request needs to know what node selector to use. This is handled by passing the node selector for that platform directly to the build request: https://github.com/projectatomic/osbs-client/pull/586/files/4430ab1bafc13b01cd5852b6b93b045fbb3a656b#diff-034b3f8b7a38c1ab83d3dc6cce5f2683R508

We don't also need to pass the platform in because it's not used for anything else.

lcarva: I see the confusion here. From the diff it looks like `platform` is being removed from the parameter list to `_do_create_prod_build`. Clearly, this is not the case. This looks fine.

PR #587, opened 2017-06-12
[WIP] adjust log line decoding as needed

lcarva:


lcarva:
Still early in the process, but this seems to do the trick.

In some cases, the response encoding value does not agree with the enconding requests "thinks" is being used. Still not exactly sure why...

This fix does appear to log unicode characters properly in koji.

twaugh:
See my approach in https://github.com/projectatomic/osbs-client/pull/588.

lcarva:
Abandoning this approach.

PR #588, opened 2017-06-12
Return Unicode text from log-fetching methods

twaugh:
Return Unicode text from log-fetching methods

When streaming or fetching logs, return Unicode text by decoding the response according to the HTTP headers.

Use response.text to automatically decode the HTTP response when not streaming it.

When streaming it, response.iter_lines() returns the encoded bytes. Use the response headers to decide how to decode it; if these are not present, use guess_json_utf() if JSON is expected, and guess 'utf-8' otherwise.
    
We always decode responses before passing them back to the client because only the response headers say how to interpret the bytes.


twaugh:
(more work required but this shows the direction I think we should take)

twaugh:
Note how we were *already* decoding the response when not streaming: https://github.com/projectatomic/osbs-client/blob/master/osbs/http.py#L59

osbs/core.py
```diff
@@ -44,13 +45,19 @@
 
 def check_response(response):
     if response.status_code not in (httplib.OK, httplib.CREATED):
-        if hasattr(response, 'content'):
-            content = response.content
+        if hasattr(response, 'text'):
+            text = response.text
         else:
-            content = ''.join(response.iter_lines())
+            encoding = response.encoding or 'utf-8'
```
twaugh: We could use `chardet.detect` when response.encoding is `None` -- or we could just assume UTF-8. Anyone care?

osbs/cli/capture.py
```diff
@@ -20,18 +20,26 @@ class IterLinesSaver(object):
...
         self.fn = fn
         self.line = 0
 
     def iter_lines(self):
+        encoding = self.encoding
         for line in self.fn():
             path = "{f}-{n:0>3}.json".format(f=self.path, n=self.line)
             logger.debug("capturing to %s", path)
+            try:
+                text = line.decode(encoding, errors='replace')
```
lcarva: str in python 3 has no decode method.

twaugh: Correct. `line` comes from `iter_lines()`, which returns encoded bytes ('bytes' in Python 3, 'str' in Python 2) -- note we're not passing `decode_unicode=True` here.

twaugh: Oops, we *were* passing `decode_unicode=True`. Fixed. :)

osbs/core.py
vrutkovs: I think we're okay with using utf-8 everywhere

osbs/cli/capture.py
vrutkovs: I'm not sure if we should `replace` by default - all our parts should be able to handle Unicode in logs. Maybe we could try decoding it as is and falling back to `errors='replace'` in case exception occurs?

twaugh:
Abandoning this approach.

Logs from containers can be in any encoding, and Docker has no way of discovering which. This means OpenShift also has no way of discovering, which is why it does not use HTTP response headers to declare the encoding when fetching logs.

Worse, logs are not just text in some encoding -- they may also contain ANSI escape sequences. Or, really, anything.

For this reason the logs-fetching API methods should treat logs as a byte stream, returning this to the caller.

PR #589, opened 2017-06-13
no longer attempt to decoded streamed build logs

lcarva:
OpenShift does not respond with any encoding value.
This causes requests module to guess it as ISO-8859-1.
Likely, the enconding is actually UTF-8, but we can't
guarantee it. Therefore, we take the approach of simply
passing through the encoded data with no effort to
attempt decoding it.

### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -143,7 +143,13 @@ class HttpStream(object):
 
     def iter_lines(self):
         kwargs = {
-            'decode_unicode': True
+            # OpenShift does not respond with any encoding value.
+            # This causes requests module to guess it as ISO-8859-1.
+            # Likely, the encoding is actually UTF-8, but we can't
+            # guarantee it. Therefore, we take the approach of simply
+            # passing through the encoded data with no effort to
+            # attempt decoding it.
+            'decode_unicode': False
         }
         if requests.__version__.startswith('2.6.'):
             kwargs['chunk_size'] = 1
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -47,14 +47,14 @@ class Response(object):
 
 
 class TestCheckResponse(object):
-    @pytest.mark.parametrize('content', [None, 'OK'])
+    @pytest.mark.parametrize('content', [None, b'OK'])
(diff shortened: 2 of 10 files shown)


osbs/http.py
```diff
@@ -143,7 +143,13 @@ def iter_chunks(self):
 
     def iter_lines(self):
         kwargs = {
-            'decode_unicode': True
+            # OpenShift does not respond with any encoding value.
+            # This causes requests module to guess it as ISO-8859-1.
+            # Likely, the enconding is actually UTF-8, but we can't
```
twaugh: encoding

lcarva: Fixed.

lcarva:
> Is there any way of testing this?

Added a test to ensure that when streaming logs, `decode_unicode` is always set to False.

tests/test_core.py
```diff
@@ -122,6 +124,29 @@ def test_stream_logs_utf8(self, openshift):  # noqa
...
+            .and_yield(logs)
+            .with_args(decode_unicode=False))
+
+        (flexmock(requests)
+            .should_receive('request')
+            .and_return(fake_response))
+
+        with caplog.atLevel(logging.ERROR):
+            for result in server.stream_logs('anything'):
+                pass
```
twaugh: Can we assert anything about the type of `result`?

lcarva: Changed it to assert that it's instance of `six.binary_type`.

twaugh:
Hmm, the StreamingResponse used by unit tests also needs changing.

twaugh:
This PR is for the minimal change of returning a bytes object generator when calling get_build_logs(follow=True).

For the corresponding change for get_build_logs(follow=False) see https://github.com/projectatomic/osbs-client/pull/590.

twaugh:
@vrutkovs So the concern with only merging this PR is that koji_promote will upload incorrectly-encoded logs because it uses the get_build_logs(follow=False), is that right?

vrutkovs:
@twaugh yes, other than that its fine

lcarva:
I can't approve @twaugh's changes because I'm the author of this PR, but +1 from me.

PR #591, opened 2017-06-14
Add documentation about mock_jsons

twaugh:

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1,0 +1,48 @@
+# Testing
+
+When writing tests for new functionality there are some things to be
+aware of.
+
+Because osbs-client is used to convey queries to and responses from a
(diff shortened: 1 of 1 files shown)


docs/testing.md
```diff
@@ -0,0 +1,32 @@
...
+as well.
+
+For all these fixtures, HTTP requests made during the unit test are
+intercepted and handled by the Connection class, which maps requests
+and methods to files within the tests/mock_jsons directory.
+
+When new functionality is going to make a request not already handled
+by this class, you will need to:
+
+* capture the response returned by the server
```
vrutkovs: We should recommend a way of doing that - IIRC a curl request would be sufficient

twaugh: Good idea, updated.

PR #592, opened 2017-06-14
branch name is always required

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -410,7 +410,7 @@ class OSBS(object):
         return build
 
     def _do_create_prod_build(self, git_uri, git_ref,
-                              git_branch,  # may be None
+                              git_branch,
                               user,
                               component=None,
                               target=None,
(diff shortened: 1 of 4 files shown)


vrutkovs:
`create_prod_build` mentions `:param git_branch: str, branch name (may be None)`

vrutkovs:
Also a lot of `kwargs.setdefault('git_branch', None)`

lcarva:
> Also a lot of kwargs.setdefault('git_branch', None)

In addition, should probably adjust here too:
```python
osbs/build/build_request.py-        if self.spec.git_branch.value:
osbs/build/build_request.py:            self.set_label('git-branch', self.spec.git_branch.value)
osbs/build/build_request.py-        else:
osbs/build/build_request.py:            self.set_label('git-branch', 'unknown')
```

lcarva:
This looks good, but let's hold off on merging until we can verify that our users have a heads up about this change.

twaugh:
Will Fedora tooling need changes or does it already provide the git branch?

lcarva:
> Will Fedora tooling need changes or does it already provide the git branch?

The fix was in pyrpkg which is used by fedpkg.

PR #593, opened 2017-06-14
inputs: Add 'add_labels_in_dockerfile' to orchestrator_inner:1.json

dzickusrh:
Allow ability to set 'release' number outside of a Dockerfile or a bump_verion()
plugin.  This sets the add_labels_in_dockerfile plugin to be executed by the
orchestrator to set a 'release' label in the Dockerfile passed to the worker.

Untested.  Relying on integration tests to verify this.

Signed-off-by: Don Zickus <dzickus@redhat.com>

### DIFF
--- a/inputs/orchestrator_inner:1.json
+++ b/inputs/orchestrator_inner:1.json
@@ -9,6 +9,12 @@
       }
     },
     {
+      "name": "add_labels_in_dockerfile",
+      "args": {
+        "labels": "{{IMPLICIT_LABELS}}"
+      }
+    },
+    {
       "name": "bump_release",
       "args": {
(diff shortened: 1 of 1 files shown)


lcarva:
This appears to be related to https://github.com/projectatomic/osbs-client/pull/581 ?
Should we combine them?

dzickusrh:
@lcarva twaugh told me no, so I made them separate.

twaugh:
Sorry, think I got confused about which things we were talking about. Yes, we probably want to combine them.

twaugh:
Is this still required? I think it was part of https://github.com/projectatomic/osbs-client/pull/581.

PR #594, opened 2017-06-14
support configMaps

mlangsdorf:
Add support for adding, getting, and deleting configMaps.

### DIFF
--- a/inputs/config_map.json
+++ b/inputs/config_map.json
@@ -1,0 +1,7 @@
+{
+  "apiVersion": "v1",
+  "kind": "ConfigMap",
+  "metadata": {
+    "name": "{{NAME}}"
+  }
+}
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -20,6 +20,7 @@ from functools import wraps
 from osbs.build.build_request import BuildRequest
 from osbs.build.build_response import BuildResponse
 from osbs.build.pod_response import PodResponse
+from osbs.build.config_map_response import ConfigMapResponse
 from osbs.constants import (BUILD_RUNNING_STATES, WORKER_OUTER_TEMPLATE,
                             WORKER_INNER_TEMPLATE, WORKER_CUSTOMIZE_CONF,
                             ORCHESTRATOR_OUTER_TEMPLATE, ORCHESTRATOR_INNER_TEMPLATE,
@@ -965,3 +966,50 @@ class OSBS(object):
     @osbsapi
     def can_orchestrate(self):
         return self.build_conf.get_can_orchestrate()
+
+    @osbsapi
+    def create_config_map(self, name, data):
+        """
+        Create an ConfigMap object on the server
+
+        Raises exception on error
+
+        :param name: str, name of configMap
+        :param data: dict, dictionary of data to be stored
+        :returns: ConfigMapResponse containing the ConfigMap with name and data
+        """
+        config_data_file = os.path.join(self.os_conf.get_build_json_store(), 'config_map.json')
+        config_data = json.load(open(config_data_file))
+        config_data['metadata']['name'] = name
+        config_data['data'] = data
+
+        response = self.os.create_config_map(config_data)
+        config_map_response = ConfigMapResponse(response.json())
+        return config_map_response
+
+    @osbsapi
+    def get_config_map(self, name):
+        """
+        Get a ConfigMap object from the server
+
+        Raises exception on error
+
+        :param name: str, name of configMap to get from the server
+        :returns: ConfigMapResponse containing the ConfigMap with the requested name
+        """
+        response = self.os.get_config_map(name)
+        config_map_response = ConfigMapResponse(response.json())
+        return config_map_response
+
+    @osbsapi
+    def delete_config_map(self, name):
+        """
+        Delete a ConfigMap object from the server
+
+        Raises exception on error
+
+        :param name: str, name of configMap to delete from the server
+        :returns: True on success
+        """
+        response = self.os.delete_config_map(name)
+        return response
--- a/osbs/build/config_map_response.py
+++ b/osbs/build/config_map_response.py
@@ -1,0 +1,43 @@
+"""
+Copyright (c) 2017 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from __future__ import print_function, absolute_import, unicode_literals
+
+import logging
+
+from osbs.utils import graceful_chain_get
+
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigMapResponse(object):
+    """
+    Wrapper for JSON describing a ConfigMap
+    """
+
+    def __init__(self, config_map):
+        """
+        :param config_map: dict, data to be stored in the ConfigMap
+        """
+        self._json = config_map
+
+    @property
+    def json(self):
+        return self._json
+
+    def get_data(self):
+        """
+        Find the data stored in the config_map
+
+        :return: dict, the json data that was passed into the ConfigMap on creation
+        """
+        data = graceful_chain_get(self.json, "data")
+        if data is None:
+            return {}
+
+        return data
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -875,6 +875,23 @@ class Openshift(object):
         check_response(response)
         return response
 
+    def create_config_map(self, config_data):
+        url = self._build_k8s_url("configmaps/")
+        response = self._post(url, data=json.dumps(config_data))
+        check_response(response)
+        return response
+
+    def get_config_map(self, config_name):
+        url = self._build_k8s_url("configmaps/%s" % config_name)
+        response = self._get(url)
+        check_response(response)
+        return response
+
+    def delete_config_map(self, config_name):
+        url = self._build_k8s_url("configmaps/%s" % config_name)
+        response = self._delete(url, data='{}')
+        check_response(response)
+
 
 if __name__ == '__main__':
     o = Openshift(openshift_api_url="https://localhost:8443/oapi/v1/",
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -28,6 +28,7 @@ from osbs.conf import Configuration
 from osbs.build.build_request import BuildRequest
 from osbs.build.build_response import BuildResponse
 from osbs.build.pod_response import PodResponse
+from osbs.build.config_map_response import ConfigMapResponse
 from osbs.build.spec import BuildSpec
 from osbs.exceptions import OsbsValidationException, OsbsException, OsbsResponseException
 from osbs.http import HttpResponse
@@ -1908,3 +1909,31 @@ class TestOSBS(object):
                                            inner_template=inner_template,
                                            outer_template=outer_template,
                                            customize_conf=customize_conf)
+
+    # osbs is a fixture here
+    @pytest.mark.parametrize(('name', 'data'), [  # noqa
+        ('special-config',
+         {
+             "special.how": "very",
+             "special.type": "charm"
+         }),
+    ])
+    def test_config_map(self, osbs, name, data):
+        obj = {
+            'apiVersion': 'v1',
+            'kind': 'ConfigMap',
+            'metadata': {'name': name},
+            'data': data,
+        }
+        (flexmock(osbs.os)
+            .should_call('create_config_map')
+            .with_args(obj))
+
+        config_map = osbs.create_config_map(name, data)
+        assert isinstance(config_map, ConfigMapResponse)
+        assert config_map.get_data() == data
+        config_map = osbs.get_config_map(name)
+        assert isinstance(config_map, ConfigMapResponse)
+        assert config_map.get_data() == data
+        config_map = osbs.delete_config_map(name)
+        assert config_map is None
--- a/tests/fake_api.py
+++ b/tests/fake_api.py
@@ -237,6 +237,23 @@ class Connection(object):
                     "file": None,  # not really empty but not relevant
                 },
             },
+
+            (API_PREFIX + "namespaces/default/configmaps/",
+             API_PREFIX + "namespaces/default/configmaps/special-config"): {
+                "post": {
+                    # Contains a configMap
+                    "file": "create_config_map.json",
+                },
+                "get": {
+                    # Contains a configMap
+                    "file": "create_config_map.json",
+                },
+                "delete": {
+                    # doesn't return anything
+                    "file": None,
+                },
+            },
+
         }
 
     @staticmethod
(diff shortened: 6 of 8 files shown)


osbs/api.py
```diff
@@ -966,3 +966,21 @@ def get_resource_quota(self, quota_name):
     @osbsapi
     def can_orchestrate(self):
         return self.build_conf.get_can_orchestrate()
+
+    @osbsapi
+    def create_config_map(self, config_name=None, config_data=None):
+        if not config_name or not config_data:
+            raise OsbsValidationException("Can't create a ConfigMap with no name or data")
```
twaugh: Just declare the method as `def create_config_map(self, config_name, config_data):`, since neither can be omitted.

osbs/core.py
```diff
@@ -866,6 +866,24 @@ def restore_resource(self, resource_type, resource):
         check_response(response)
         return response
 
+    def create_config_map(self, config_name, config_data):
+        url = self._build_k8s_url("configmaps/")
+        response = self._post(url, data=json.dumps(config_data))
```
twaugh: I think we need more than this. `config_data` will be something like `{"metadata.json": "{...}"}`, i.e. the `data` part of the v1.ConfigMap: https://docs.openshift.org/latest/rest_api/kubernetes_v1.html#create-a-configmap

See `create_image_stream` for the equivalent for creating a v1.ImageStream: it reads a template, image_stream.json, and adjusts fields appropriately. I think we probably want to do the same sort of thing as that.

osbs/core.py
```diff
@@ -866,6 +866,24 @@ def restore_resource(self, resource_type, resource):
...
+        check_response(response)
+        return response
+
+    def get_config_map(self, config_name):
+        url = self._build_k8s_url("configmaps/")
+        response = self._get(url, data=json.dumps(config_name))
+        check_response(response)
+        return response
```
twaugh: I think we want to return something the caller can use without having to know the details of what a v1.ConfigMap looks like. See osbs.build.pod_response.PodResponse for the sort of thing I'm thinking of. I think we want an osbs.build.configmap_response.ConfigMapResponse with a get_*() method that retrieves the `data` value.
http://osbs.readthedocs.io/en/latest/multiarch.html#configmapresponse

osbs/api.py
```diff
@@ -966,3 +967,31 @@ def get_resource_quota(self, quota_name):
...
+
+        Raises exception on error
+
+        :param name: str, name of configMap
+        :param data: dict, dictionary of data to be stored
+        """
+        config_data_file = os.path.join(self.os_conf.get_build_json_store(), 'config_map.json')
+        config_data = json.load(open(config_data_file))
+        config_data['metadata']['name'] = name
+        config_data['metadata']['json'] = data
```
twaugh: Don't we want `config_data['data'] = data` here?
https://docs.openshift.org/latest/rest_api/kubernetes_v1.html#v1-configmap

osbs/build/config_map_response.py
```diff
@@ -0,0 +1,43 @@
+"""
+Copyright (c) 2015 Red Hat, Inc
```
twaugh: Please add 2017 here.

osbs/build/config_map_response.py
```diff
@@ -0,0 +1,43 @@
...
+
+from osbs.utils import graceful_chain_get
+
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigMapResponse(object):
+    """
+    Wrapper for JSON describing build pod
```
twaugh: docstring needs updating

osbs/build/config_map_response.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+
+    def __init__(self, config_map):
+        """
+        :param request: http.Request
```
twaugh: There's no `request` param, but `config_map` isn't documented.

osbs/build/config_map_response.py
```diff
@@ -0,0 +1,43 @@
...
+
+    @property
+    def json(self):
+        return self._json
+
+    def get_data(self):
+        """
+        Find the image IDs the containers use.
+
+        :return: dict, image tag to docker ID
```
twaugh: docstring looks wrong

tests/test_api.py
```diff
@@ -1908,3 +1909,24 @@ def test_do_create_prod_build_branch_required(self, osbs, branch_name):
...
+    # osbs is a fixture here
+    @pytest.mark.parametrize(('name', 'data', 'raise_exception'), [  # noqa
+        ('special-config',
+         {
+             "special.how": "very",
+             "special.type": "charm"
+         },
+         False),
+    ])
+    def test_config_map(self, osbs, name, data, raise_exception):
```
twaugh: Is `raise_exception` needed?

tests/test_api.py
```diff
@@ -1908,3 +1909,24 @@ def test_do_create_prod_build_branch_required(self, osbs, branch_name):
...
+        if raise_exception:
+            with pytest.raises(OsbsValidationException):
+                osbs.create_config_map(name, data)
+        else:
+            assert osbs.create_config_map(name, data)
+            config_map = osbs.get_config_map(name)
+            assert isinstance(config_map, ConfigMapResponse)
+            assert config_map.get_data() == data
+            config_map = osbs.delete_config_map(name)
+            assert not config_map
```
twaugh: We could even `assert config_map is None`.

inputs/config_map.json
```diff
@@ -0,0 +1,7 @@
+{
+  "apiVersion": "v1",
+  "kind": "configMap",
```
twaugh: I think it's `ConfigMap` (capital C and M), at least according to this example: https://docs.openshift.org/latest/dev_guide/configmaps.html#overview

Also is it worth adding a `"data": {},` here? (not strictly necessary, just might make it clearer what goes where)

osbs/build/config_map_response.py
mlangsdorf: Fixed.

osbs/api.py

osbs/build/config_map_response.py
mlangsdorf: Fixed

osbs/build/config_map_response.py
mlangsdorf: Fixed

osbs/build/config_map_response.py
mlangsdorf: Fixed

inputs/config_map.json
mlangsdorf: Fixed.

tests/test_api.py
mlangsdorf: Not any longer.
Fixed.

tests/test_api.py
mlangsdorf: Fixed

osbs/api.py
```diff
@@ -966,3 +967,31 @@ def get_resource_quota(self, quota_name):
...
+
+        Raises exception on error
+
+        :param name: str, name of configMap
+        :param data: dict, dictionary of data to be stored
+        """
+        config_data_file = os.path.join(self.os_conf.get_build_json_store(), 'config_map.json')
+        config_data = json.load(open(config_data_file))
+        config_data['metadata']['name'] = name
+        config_data['metadata']['data'] = data
```
twaugh: I think it should be `config_data['data'] = data` shouldn't it?

osbs/build/config_map_response.py
```diff
@@ -0,0 +1,43 @@
...
+        """
+        self._json = config_map
+
+    @property
+    def json(self):
+        return self._json
+
+    def get_data(self):
+        """
+        Find the image IDs the containers use.
```
twaugh: Shouldn't this say something about keys and values?

tests/test_api.py
```diff
@@ -1908,3 +1909,19 @@ def test_do_create_prod_build_branch_required(self, osbs, branch_name):
...
+    # osbs is a fixture here
+    @pytest.mark.parametrize(('name', 'data'), [  # noqa
+        ('special-config',
+         {
+             "special.how": "very",
+             "special.type": "charm"
+         }),
+    ])
+    def test_config_map(self, osbs, name, data):
+        assert osbs.create_config_map(name, data)
```
twaugh: Would it be good to use flexmock to verify that the JSON is prepared correctly from the template?

osbs/core.py
```diff
@@ -872,6 +872,23 @@ def restore_resource(self, resource_type, resource):
...
         return response
 
+    def create_config_map(self, config_data):
+        url = self._build_k8s_url("configmaps/")
+        response = self._post(url, data=json.dumps(config_data))
+        check_response(response)
+        return response
+
+    def get_config_map(self, config_name):
+        url = self._build_k8s_url("configmaps/")
```
twaugh: Need to include the name in the url, otherwise we'll get a v1.ConfigMapList of all config maps: https://docs.openshift.org/latest/rest_api/kubernetes_v1.html#list-or-watch-objects-of-kind-configmap

osbs/core.py
```diff
@@ -872,6 +872,23 @@ def restore_resource(self, resource_type, resource):
...
+        response = self._get(url, data=json.dumps(config_name))
+        check_response(response)
+        return response
+
+    def delete_config_map(self, config_name):
+        url = self._build_k8s_url("configmaps/")
+        response = self._delete(url, data=json.dumps(config_name))
```
twaugh: The url is incorrect, and the JSON data needs to be formatted correctly (include v1.DeleteOptions): https://docs.openshift.org/latest/rest_api/kubernetes_v1.html#delete-a-configmap

tests/test_api.py
twaugh: All we're asserting about the return value from create_config_map() at the moment is that it's truthy. @lcarva, do we mind whether this is an `HttpResponse` (as currently) or a `ConfigMapResponse`?

mlangsdorf: Is that assert supposed to be for the obj or are you indicating where I should add the flexmock call in my existing code?

osbs/core.py
mlangsdorf: fixed.

osbs/build/config_map_response.py
mlangsdorf: well, it should say something different, at any rate.

osbs/api.py
mlangsdorf: I wasn't sure and I don't think it matters because graceful_chain_get will do the right thing regardless, but sure, put in 'data' on the top level.

fixed.

osbs/core.py
mlangsdorf: DeleteOptions has 5 parameters, all optional. Should the caller to delete_config_map pass them in or should I set them in delete_config_map? in which case I'm just passing 'body = {}' as far as I can tell.

tests/test_api.py
twaugh: I put the assert in for context, to show that the `obj = {...}` and `flexmock` parts should go above it. It's saying:

- when osbs.os.create_config_map() is called, check that its parameter compares as equal to `obj` before letting execution continue ('should_call()' is a flexmock 'spy', to verify parameters as well as actually calling the method)
- call osbs.create_config_map(name, data)
- if flexmock allowed the call through we know the parameter was correct
- finally, assert that the return value from osbs.create_config_map(name,data) is truthy (as before)

osbs/api.py
twaugh: It does matter though: OpenShift is on the other side of that PUT request, and it requires things to be in the correct places. It would be e.g. free to ignore keys it doesn't know.

The `graceful_chain_get` part is when we're parsing objects provided to us by OpenShift.

osbs/core.py
twaugh: Reading through the rest of the document and comparing, I think 'body' actually means the body of the request (the 'data=' part of the _delete() call), so I suppose it's required in the sense that requests have content.

Yes, `data='{}'` seems fine.

mlangsdorf: Fixed.

tests/test_api.py
mlangsdorf: It is supposed to return a ConfigMap so it should be a ConfigMapResponse. 

Fixed.

osbs/api.py
```diff
@@ -966,3 +967,33 @@ def get_resource_quota(self, quota_name):
...
+        config_data = json.load(open(config_data_file))
+        config_data['metadata']['name'] = name
+        config_data['data'] = data
+
+        response = self.os.create_config_map(config_data)
+        config_map_response = ConfigMapResponse(response.json())
+        return config_map_response
+
+    @osbsapi
+    def get_config_map(self, name):
```
twaugh: Could add a docstring here.

osbs/api.py
```diff
@@ -966,3 +967,33 @@ def get_resource_quota(self, quota_name):
...
+        response = self.os.get_config_map(name)
+        config_map_response = ConfigMapResponse(response.json())
+        return config_map_response
+
+    @osbsapi
+    def delete_config_map(self, name):
```
twaugh: Could add a docstring here.

twaugh:
Also: this worked fine for me against an OpenShift server.

osbs/api.py

osbs/api.py

osbs/api.py
```diff
@@ -966,3 +967,50 @@ def get_resource_quota(self, quota_name):
...
+        """
+        Delete a ConfigMap object from the server
+
+        Raises exception on error
+
+        :param name: str, name of configMap to dlete from the server
```
twaugh: typo: delete

mlangsdorf: doh!

mlangsdorf:
local python tests on a running cluster show that:
create_config_map(name, data) returns a ConfigMapResponse that spits out data if you do get_data() on it
get_config_map(name) returns a ConfigMapResponse that spits out data if you do get_data() on it
delete_config_map(name) returns True
get_config_map(name) again raises an exception
delete_config_map(name) again raises an exception

Seems valid.

osbs/api.py
```diff
@@ -965,3 +966,50 @@ def get_resource_quota(self, quota_name):
...
+        Get a ConfigMap object from the server
+
+        Raises exception on error
+
+        :param name: str, name of configMap to get from the server
+        :returns: ConfigMapResponse containing the ConfigMap with the requested name
+        """
+        response = self.os.get_config_map(name)
+        config_map_response = ConfigMapResponse(response.json())
+        return config_map_response
```
lcarva: Any reason not to simply return the data portion of the json? `return response.json().get('data')`

twaugh: Rather than a ConfigMapResponse?

The idea is to provide an interface that could grow with any changes that occur, and to have some consistency in the API: ask for a thing, get a class instance that provides access methods.

lcarva: Ah ok, was just wondering if there was a particular reason other than good practice as we are inconsistent with using this pattern.

twaugh: I know there are some API methods that don't do it this way (e.g. `get_resource_quota()`, `get_user()`, the ImageStream methods etc). I wish they all did it this way though, and there are several that do (e.g. `get_build()`, `get_pod_for_build()`).

PR #597, opened 2017-06-15
Allow empty triggers in prod json

bfontecc007:
Signed-off-by: Bret Fontecchio <bfontecc@redhat.com>

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -980,7 +980,7 @@ class BuildRequest(object):
 
         self.render_tag_and_push_registries()
 
-        if 'triggers' in self.template['spec']:
+        if self.has_ist_trigger():
             imagechange = self.template['spec']['triggers'][0]['imageChange']
             imagechange['from']['name'] = self.spec.trigger_imagestreamtag.value
 
--- a/tests/build_/test_build_request.py
+++ b/tests/build_/test_build_request.py
@@ -1267,13 +1267,33 @@ class TestBuildRequest(object):
         with pytest.raises(OsbsValidationException):
             build_request.render()
 
+    def test_render_prod_with_empty_triggers(self, tmpdir):
+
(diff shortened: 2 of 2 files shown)


osbs/build/build_request.py
```diff
@@ -980,7 +980,8 @@ def render(self, validate=True):
 
         self.render_tag_and_push_registries()
 
-        if 'triggers' in self.template['spec']:
+        triggers = self.template['spec'].get('triggers', [])
+        if triggers:
```
rcerven: you can as well use 'self.has_ist_trigger()'

tests/build_/test_build_request.py
```diff
@@ -1267,14 +1267,34 @@ def test_render_prod_with_pulp_no_auth(self):
...
     @staticmethod
-    def create_image_change_trigger_json(outdir):
+    def create_image_change_trigger_json(outdir, custom_triggers=None):
         """
         Create JSON templates with an image change trigger added.
 
         :param outdir: str, path to store modified templates
         """
 
+        triggers = custom_triggers if custom_triggers is not None else [
```
vrutkovs: Nitpick: its a bit tough to read this test. Instead of setting it in the test function we should be passing it using `parametrize`:
```
    @pytest.mark.parametrize(('triggers'), (
        None,
        <here goes that long dict>
    ))
    def create_image_change_trigger_json(outdir, triggers):
```

This way the test would run twice with two different params, which could later on be extended.

bfontecc007: I made a separate empty triggers test instead of parametrizing test_render_prod_request_with_trigger because that test only passes if a whole bunch of plugins load which are kicked off by the trigger. I thought it was less readable when the majority of the test is wrapped in an if/else on the parameters.

twaugh: Makes sense to me.

vrutkovs: Hmm, agree, lets leave it like it is then

PR #598, opened 2017-06-15
don't log ERROR when checking for istreams and istags

rcerven:

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -43,14 +43,14 @@ from .http import HttpSession
 logger = logging.getLogger(__name__)
 
 
-def check_response(response):
+def check_response(response, log_level=logging.ERROR):
     if response.status_code not in (httplib.OK, httplib.CREATED):
         if hasattr(response, 'content'):
             content = response.content
         else:
             content = b''.join(response.iter_lines())
 
-        logger.error("[%d] %s", response.status_code, content)
+        logger.log(log_level, "[%d] %s", response.status_code, content)
         raise OsbsResponseException(message=content, status_code=response.status_code)
 
 
@@ -740,7 +740,7 @@ class Openshift(object):
     def get_image_stream_tag(self, tag_id):
         url = self._build_url("imagestreamtags/%s" % tag_id)
         response = self._get(url)
-        check_response(response)
+        check_response(response, log_level=logging.DEBUG)
         return response
 
     def put_image_stream_tag(self, tag_id, tag):
@@ -793,7 +793,7 @@ class Openshift(object):
     def get_image_stream(self, stream_id):
         url = self._build_url("imagestreams/%s" % stream_id)
         response = self._get(url)
-        check_response(response)
+        check_response(response, log_level=logging.DEBUG)
         return response
 
     def create_image_stream(self, stream_json):
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -59,29 +59,49 @@ class TestCheckResponse(object):
         response = Response(status_code, content=content)
         check_response(response)
 
-    def test_check_response_bad_stream(self, caplog):
+    @pytest.mark.parametrize('log_errors', (True, False))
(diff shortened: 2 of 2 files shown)


osbs/core.py
```diff
@@ -43,14 +43,17 @@
 logger = logging.getLogger(__name__)
 
 
-def check_response(response):
+def check_response(response, error_msg=True):
```
twaugh: Maybe `log_errors`? i.e. making this an "action"-type name rather than "noun".

osbs/core.py
```diff
@@ -43,14 +43,17 @@
...
 
-def check_response(response):
+def check_response(response, log_errors=True):
     if response.status_code not in (httplib.OK, httplib.CREATED):
         if hasattr(response, 'content'):
             content = response.content
         else:
             content = b''.join(response.iter_lines())
 
-        logger.error("[%d] %s", response.status_code, content)
```
lcarva: How about using `logger.log`[0]?
https://docs.python.org/2/library/logging.html#logging.Logger.log

In that case `check_response` could take the log level to use, defaulting to `logging.ERROR`.

osbs/core.py
```diff
@@ -740,14 +743,14 @@ def set_annotations_on_build(self, build_id, annotations):
...
-        check_response(response)
+        check_response(response, log_errors=False)
         return response
 
     def put_image_stream_tag(self, tag_id, tag):
         url = self._build_url("imagestreamtags/%s" % tag_id)
         response = self._put(url, data=json.dumps(tag),
                              headers={"Content-Type": "application/json"})
-        check_response(response)
+        check_response(response, log_errors=False)
```
lcarva: We still want to log errors here right?

osbs/core.py
rcerven: well we want to either error or debug, I don't think we need more levels to use,
and the level int for error is 40 and debug 10, isn't very readable

osbs/core.py
rcerven: right wrong function :)

osbs/core.py
lcarva: > the level int for error is 40 and debug 10, isn't very readable

You wouldn't write `check_response(response, log_level=10)`, you'd write `check_response(response, log_level=logging.DEBUG)` which to me is very readable.

I'm fine as is, just thought it might be a little bit nicer to make it more flexible.

rcerven: ahhh I see

rcerven: ok changed

osbs/core.py
```diff
@@ -43,14 +43,15 @@
...
+def check_response(response, log_level=logging.ERROR):
     if response.status_code not in (httplib.OK, httplib.CREATED):
         if hasattr(response, 'content'):
             content = response.content
         else:
             content = b''.join(response.iter_lines())
 
-        logger.error("[%d] %s", response.status_code, content)
+        log_message = "[%d] %s" % (response.status_code, content)
+        logger.log(log_level, log_message)
```
lcarva: logging module should be the one performing string formatting. Please change it to:
`logger.log(log_level, "[%d] %s", response.status_code, content)`

tests/test_core.py
```diff
@@ -59,29 +59,49 @@ def test_check_response_ok(self, status_code, content):
...
-    def test_check_response_bad_stream(self, caplog):
+    @pytest.mark.parametrize('log_errors', (True, False))
+    def test_check_response_bad_stream(self, caplog, log_errors):
         iterable = [b'iter', b'lines']
         status_code = httplib.CONFLICT
         response = Response(status_code, iterable=iterable)
+        if log_errors:
+            log_type = logging.ERROR  # 40
+        else:
+            log_type = logging.DEBUG  # 10
```
twaugh: Not sure if these comments add anything. We don't care what the values of these are do we?

rcerven: sure I can remove it

PR #600, opened 2017-06-16
Handle Null case for triggers in prod JSON

bfontecc007:
Signed-off-by: Bret Fontecchio <bfontecc@redhat.com>

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -166,6 +166,8 @@ class BuildRequest(object):
     def has_ist_trigger(self):
         """Return True if this BuildConfig has ImageStreamTag trigger."""
         triggers = self.template['spec'].get('triggers', [])
+        if not triggers:
+            return False
         for trigger in triggers:
             if trigger['type'] == 'ImageChange' and \
                     trigger['imageChange']['from']['kind'] == 'ImageStreamTag':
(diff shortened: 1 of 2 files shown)


tests/build_/test_build_request.py
```diff
@@ -29,6 +29,7 @@
                              TEST_COMPONENT, TEST_GIT_BRANCH, TEST_GIT_REF,
                              TEST_GIT_URI, TEST_GIT_URI_HUMAN_NAME)
 
+USE_DEFAULT_TRIGGERS = object()
```
twaugh: Extra blank line needed here for PEP8.
```
tests/build_/test_build_request.py:34:1: E302 expected 2 blank lines, found 1
```

lcarva: I wonder why stickler didn't complain about this.

twaugh:
coveralls is reporting a decrease in coverage because now https://github.com/projectatomic/osbs-client/pull/600/files#diff-6451bca7357ec20f16d74cb3cc6d5553R175 never gets control.

Maybe we need a test case for 'there are triggers, but none are imageChange'? e.g. one of these: https://docs.openshift.org/latest/dev_guide/builds/triggering_builds.html#generic-webhooks

osbs/build/build_request.py
```diff
@@ -166,6 +166,8 @@ def dj(self):
     def has_ist_trigger(self):
         """Return True if this BuildConfig has ImageStreamTag trigger."""
         triggers = self.template['spec'].get('triggers', [])
+        if not triggers:
```
lcarva: To address the code coverage decrease, we could remove this if-block and change the for loop to:
```python
for trigger in triggers or []:
```

lcarva: Or maybe better yet to set it to an empty list when triggers variable is defined:
```python
triggers = self.template['spec'].get('triggers') or []
```

In either case, this is fine as is to me.

bfontecc007:
Looks like I got coverage back up

PR #601, opened 2017-06-19
pass equal_labels configuration to worker

rcerven:

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -215,6 +215,13 @@ class BuildRequest(object):
         self.dj.dock_json_set_arg(phase, plugin, 'platforms', self.spec.platforms.value)
         self.dj.dock_json_set_arg(phase, plugin, 'build_kwargs', build_kwargs)
 
+        equal_labels_string = None
+        equal_labels_sets = []
+        if self.spec.equal_labels.value:
+            for equal_set in self.spec.equal_labels.value:
+                equal_labels_sets.append(':'.join(equal_set))
+            equal_labels_string = ','.join(equal_labels_sets)
+
         # Parameters to be used as Configuration overrides for each worker
         config_kwargs = {
             'authoritative_registry': self.spec.authoritative_registry.value,
@@ -235,6 +242,7 @@ class BuildRequest(object):
             'source_registry_uri': self.spec.source_registry_uri.value,
             'sources_command': self.spec.sources_command.value,
             'vendor': self.spec.vendor.value,
+            'equal_labels': equal_labels_string,
         }
 
         # Remove empty values, and always convert to string for better interaction
--- a/tests/build_/test_build_request.py
+++ b/tests/build_/test_build_request.py
@@ -1163,6 +1163,7 @@ class TestBuildRequest(object):
             'source_registry_uri': 'source_registry_uri',
             'sources_command': 'sources_command',
             'vendor': 'vendor',
+            'equal_labels': [['label1', 'label2'], ['label3', 'label4']],
         },
         {}
     ))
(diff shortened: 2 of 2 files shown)


lcarva:
Needs corresponding changes to `test_render_orchestrate_build` in test_build_request.py.

osbs/build/build_request.py
```diff
@@ -235,6 +235,7 @@ def render_orchestrate_build(self):
             'source_registry_uri': self.spec.source_registry_uri.value,
             'sources_command': self.spec.sources_command.value,
             'vendor': self.spec.vendor.value,
+            'equal_labels': self.spec.equal_labels.value,
```
lcarva: This needs to be serialized back into a str. See how `smtp_error_addresses` is done for instance.
It might be a little more complex for `equal_labels` though.

tests/build_/test_build_request.py
```diff
@@ -1163,6 +1163,7 @@ def test_render_optional_plugins(self, tmpdir):
             'source_registry_uri': 'source_registry_uri',
             'sources_command': 'sources_command',
             'vendor': 'vendor',
+            'equal_labels': 'equal_labels',
```
twaugh: Maybe make this a bit more complicated?

Also needs an assert further down to check it was unpacked and repacked correctly, like the other vars.

rcerven: sure, I thought it is dynamic :)

osbs/build/build_request.py
```diff
@@ -235,6 +241,7 @@ def render_orchestrate_build(self):
             'source_registry_uri': self.spec.source_registry_uri.value,
             'sources_command': self.spec.sources_command.value,
             'vendor': self.spec.vendor.value,
+            'equal_labels': equal_labels_string,
```
lcarva: Let's set this to `None` if falsy because a couple of lines below, we remove keys that have `None` values.

rcerven: ahh right

tests/build_/test_build_request.py
```diff
@@ -1261,6 +1262,8 @@ def test_render_orchestrate_build(self, tmpdir, platforms, secret, disabled,
             assert kwargs.get('source_registry_uri') == worker_config.get_source_registry_uri()
             assert kwargs.get('sources_command') == worker_config.get_sources_command()
             assert kwargs.get('vendor') == worker_config.get_vendor()
+            assert (kwargs.get('equal_labels') or [] ==
```
vrutkovs: why would we want to have `or` in assert?

rcerven: well it is like on previous lines, just in case you would be testing kwargs without that label

lcarva: I don't see the benefit of splitting it up like that. Keep in mind that additional_kwargs doesn't necessarily have to set `equal_labels` in this test. Coupling the two would not be correct IMO.

vrutkovs: The problem is we're not checking both options when we got `or` there:
`assert ('nonsense' or [] == [])` would return True. Alternatively we could turn it to `kwargs.get('equal_labels', []) == ...`

>Coupling the two would not be correct IMO.

Hmm, agree

lcarva: > The problem is we're not checking both options when we got or there:

We actually do.  If `equal_labels` is not a key in the provided dict, then we verify that the default value produced by `get_equal_labels` is used, which is `[]`.

lcarva: Hmm... @vrutkovs is right, this check is not behaving as expected:
```python
>>> assert ('nonsense' or [] == [])
>>> assert (('nonsense' or []) == [])
AssertionError


```

rcerven: fixed

PR #602, opened 2017-06-19
[WIP] render_add_filesystem: improved multi-arch build support

jarodwilson:
- Pass along an architectures variable to atomic-reactor
- Add new filesystem_koji_task_id parameter to create_worker_build API
- Bump DEFAULT_ARRANGEMENT_VERSION to 2
- Update self-tests with filesystem_koji_task_id usage


inputs/worker_inner:2.json
```diff
@@ -0,0 +1,174 @@
...
+    },
+    {
+      "name": "pull_base_image",
+      "args": {
+        "parent_registry": "{{REGISTRY_URI}}",
+        "parent_registry_insecure": true
+      }
+    },
+    {
+      "name": "bump_release",
```
lcarva: This plugin has been removed from `worker_inner:1.json`, let's remove it here too.

inputs/orchestrator_inner:2.json
```diff
@@ -0,0 +1,48 @@
...
+  "prebuild_plugins": [
+    {
+      "name": "pull_base_image",
+      "args": {
+        "parent_registry": "{{REGISTRY_URI}}",
+        "parent_registry_insecure": true
+      }
+    },
+    {
+      "name": "bump_release",
```
lcarva: The following has been added to `orchestrator_inner:1.json` recently:
```json
    {
      "name": "add_labels_in_dockerfile",
      "args": {
        "labels": "{{IMPLICIT_LABELS}}"
      }
    }
```
Let's add it here too (right before `bump_release`).

inputs/orchestrator_inner:2.json
```diff
@@ -0,0 +1,48 @@
+{
+  "client_version": "{{VERSION}}",
+  "prebuild_plugins": [
```
lcarva: We need `add_filesystem` plugin in here as well.

osbs/build/build_request.py
```diff
@@ -1112,6 +1114,11 @@ def render(self, validate=True):
             self.template['metadata'].setdefault('labels', {})
             self.template['metadata']['labels']['koji-task-id'] = str(koji_task_id)
 
+        filesystem_koji_task_id = self.spec.filesystem_koji_task_id.value
```
lcarva: We also need to set the `from_task_id` parameter for the `add_filesystem` plugin. 
See `render_add_filesystem` in this class.

osbs/build/build_request.py
```diff
@@ -1112,6 +1114,11 @@ def render(self, validate=True):
...
+        if filesystem_koji_task_id is not None:
+            self.template['metadata']['labels']['filesystem-koji-task-id'] = \
```
lcarva: Let's use the helper method `set_label` in this class.
FTW let's do the same for the `koji-task-id` above :)

inputs/orchestrator_inner:2.json
jarodwilson: Ah, whoops, I must have missed those changes when I did a rebase onto current master. I'll fix these bits up.

osbs/build/build_request.py
jarodwilson: Done, for both *-task-id options. filesystem-koji-task-id was missing the setdefault portion, either I missed it when I first hacked on this, or I missed it as part of rebasing.

tests/test_api.py
```diff
@@ -220,6 +221,7 @@ def test_create_worker_build_missing_param(self, osbs, platform, release,
             'git_ref': TEST_GIT_REF,
             'git_branch': TEST_GIT_BRANCH,
             'user': TEST_USER,
+            'filesystem_koji_task_id': TEST_KOJI_TASK_ID,
```
jarodwilson: Yeah, those should be. Will update that.

jarodwilson:
Will squash and clean up commits before submitting a formal pull request.

inputs/orchestrator_inner:2.json
```diff
@@ -0,0 +1,62 @@
...
+    {
+      "name": "add_filesystem",
+      "args": {
+        "hub": "{{KOJI_HUB}}",
+        "repos": [],
+        "from_task_id": "{{FROM_TASK_ID}}"
```
lcarva: Let's remove this since it won't necessarily be overwritten.

inputs/orchestrator_inner:2.json
```diff
@@ -0,0 +1,61 @@
+{
+  "client_version": "{{VERSION}}",
+  "prebuild_plugins": [
+    {
+      "name": "add_filesystem",
+      "args": {
+        "hub": "{{KOJI_HUB}}",
+        "repos": [],
```
lcarva: Trailing comma is invalid json.

inputs/orchestrator_inner:2.json
```diff
@@ -0,0 +1,55 @@
+{
+  "client_version": "{{VERSION}}",
+  "prebuild_plugins": [
+    {
+      "name": "add_filesystem",
+      "args": {
+        "hub": "{{KOJI_HUB}}",
+        "repos": []
+      }
+    },
```
lcarva: @twaugh we had to remove `add_labels_from_dockerfile` plugin from here because the base image is not inspected in this case (filesystem tarball is not streamed into docker engine of orchestrator build).

We can either keep this plugin out of here, or change the plugin [here](https://github.com/projectatomic/atomic-reactor/blob/master/atomic_reactor/plugins/pre_add_labels_in_df.py#L279) to fallback to an empty Labels dict if base image is not inspected.

I think the fallback mechanism makes sense. What do you think?

twaugh: I think it can be removed from the orchestrator build now. That's [where we want to get to](http://osbs.readthedocs.io/en/latest/multiarch.html#anatomy-of-an-orchestrator-build) anyway.

lcarva: Ack. Let's keep `add_labels_from_dockerfile` out of here.

PR #604, opened 2017-06-20
Allow skipping errors when builder is being backed up / restored

vrutkovs:
This adds two separate switches for osbs-client: one would skip
resourcequotas to pause/restore builder - and the other one would
continue if one of the resources has failed to be backed up

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -498,12 +498,20 @@ def cmd_backup(args, osbs):
     else:
         outfile = dirname + ".tar.bz2"
 
-    with paused_builds(osbs, quota_name='pause-backup'):
+    with paused_builds(osbs, quota_name='pause-backup',
+                       ignore_quota_errors=args.ignore_quota_errors):
         with TarWriter(outfile, dirname) as t:
             for resource_type in BACKUP_RESOURCES:
-                logger.info("dumping %s", resource_type)
-                resources = osbs.dump_resource(resource_type)
-                t.write_file(resource_type + ".json", json.dumps(resources).encode('ascii'))
+                try:
+                    logger.info("dumping %s", resource_type)
+                    resources = osbs.dump_resource(resource_type)
+                    t.write_file(resource_type + ".json", json.dumps(resources).encode('ascii'))
+                except Exception as e:
+                    if args.continue_on_error:
+                        logger.warning(
+                            "Error during {} backup".format(resource_type), exc_info=True)
+                    else:
+                        raise e
 
     if not hasattr(outfile, "write"):
         logger.info("backup archive created: %s", outfile)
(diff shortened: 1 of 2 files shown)


osbs/utils.py
```diff
@@ -201,14 +201,14 @@ def checkout_git_repo(git_uri, git_ref, git_branch=None):
 
 
 @contextlib.contextmanager
-def paused_builds(osbs, quota_name=None):
+def paused_builds(osbs, quota_name=None, ignore_quota_errors=False):
     try:
         logger.info("pausing builds")
-        osbs.pause_builds(quota_name=quota_name)
+        osbs.pause_builds(quota_name=quota_name, ignore_quota_errors=ignore_quota_errors)
```
twaugh: Why not wrap this call in this?:
```
try:
    ...
except OsbsResponseException as e:
    if e.status_code != requests.codes.FORBIDDEN:
      raise
```

and likewise for resume_builds below?

vrutkovs: Right, its a good idea - exception may occur when quota json is being loaded etc.

osbs/api.py
```diff
@@ -872,10 +872,19 @@ def _load_quota_json(self, quota_name=None):
...
     @osbsapi
-    def pause_builds(self, quota_name=None):
+    def pause_builds(self, quota_name=None, ignore_quota_errors=True):
         # First, set quota so 0 pods are allowed to be running
         quota_name, quota_json = self._load_quota_json(quota_name)
-        self.os.create_resource_quota(quota_name, quota_json)
+        try:
+            self.os.create_resource_quota(quota_name, quota_json)
+        except OsbsResponseException as e:
+            if e.status_code == 403:
```
rcerven: if e.status_code == 403 and ignore_quota_errors:

would look nicer

osbs/utils.py
vrutkovs: Done

osbs/api.py
vrutkovs: Done, now its `if ignore_quota_errors and (e.status_code == requests.codes.FORBIDDEN):` to avoid magic numbers too

osbs/cli/main.py
```diff
@@ -497,12 +497,20 @@ def cmd_backup(args, osbs):
...
         with TarWriter(outfile, dirname) as t:
             for resource_type in BACKUP_RESOURCES:
-                logger.info("dumping %s", resource_type)
-                resources = osbs.dump_resource(resource_type)
-                t.write_file(resource_type + ".json", json.dumps(resources).encode('ascii'))
+                try:
+                    logger.info("dumping %s", resource_type)
+                    resources = osbs.dump_resource(resource_type)
+                    t.write_file(resource_type + ".json", json.dumps(resources).encode('ascii'))
+                except Exception as e:
```
lcarva: Do we want to catch a more specific error here?

vrutkovs: I think not - a lot of different errors may occur - timeouts, missing permissions, write errors etc. Not sure if there is a more specific base class which would catch all of these

vrutkovs:
Rebased

lcarva:
Updated release notes.

PR #606, opened 2017-06-20
janitorial: clean up whitespace and other linter errors

mlangsdorf:
I ran flake8 on osbs/ and tests/ and fixed all the long lines, incorrect
whitespace, and unused variables.

The only functional change was in tests/build_/test_build_request.py:

@@ -1238,7 +1230,7 @@ class TestBuildRequest(object):
             if isinstance(worker_build_image, type):
                 with pytest.raises(worker_build_image):
                     worker_config_kwargs['build_image']
-                worker_config.get_build_image() == None
+                assert not worker_config.get_build_image()

which converts a nonsense statement into the intended assert.

### DIFF
--- a/tests/build_/test_build_request.py
+++ b/tests/build_/test_build_request.py
@@ -219,10 +219,10 @@ class TestBuildRequest(object):
         assert pull_base_image is not None
         assert ('args' not in pull_base_image or
                 'parent_registry' not in pull_base_image['args'] or
-                pull_base_image['args']['parent_registry'] == None)
+                not pull_base_image['args']['parent_registry'])
 
-        assert plugin_value_get(plugins, "exit_plugins", "store_metadata_in_osv3", "args", "url") == \
-            "http://openshift/"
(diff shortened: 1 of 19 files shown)


tests/build_/test_build_request.py
```diff
@@ -1652,7 +1650,7 @@ def test_prod_is_custom_base_image(self, tmpdir, base_image, is_custom):
         kwargs = get_sample_prod_params()
         kwargs['base_image'] = base_image
         build_request.set_params(**kwargs)
-        build_json = build_request.render()
+        build_json = build_request.render()  # noqa
```
twaugh: Do we want to just not store the return value?

mlangsdorf: That's also a solution, but that was a functional change beyond my remit.

I can make that change though.

twaugh: I'm happy with it either way.

PR #607, opened 2017-06-20
Trigger remove2

bfontecc007:
Remove ImageChange triggers if triggers are removed from prod.json

### DIFF
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -142,6 +142,16 @@ def graceful_chain_del(d, *args):
         pass
 
 
+def has_triggers(build_config):
+    return graceful_chain_get(build_config, 'spec', 'triggers') is not None
+
+
+def clean_triggers(orig, new):
+    if not has_triggers(new) and has_triggers(orig):
+        orig['spec']['triggers'] = [t for t in orig['spec']['triggers']
+                                    if t.get('type', None) != 'ImageChange']
+
+
 def buildconfig_update(orig, new, remove_nonexistent_keys=False):
     """Performs update of given `orig` BuildConfig with values from `new` BuildConfig.
     Both BuildConfigs have to be represented as `dict`s.
@@ -154,6 +164,7 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
       (see https://github.com/projectatomic/osbs-client/pull/273#issuecomment-148038314)
     """
     if isinstance(orig, dict) and isinstance(new, dict):
+        clean_triggers(orig, new)
         if remove_nonexistent_keys:
             missing = set(orig.keys()) - set(new.keys())
             for k in missing:
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -20,7 +20,8 @@ from osbs.utils import (buildconfig_update,
                         git_repo_humanish_part_from_uri,
                         get_time_from_rfc3339, strip_registry_from_image,
                         TarWriter, TarReader, make_name_from_git,
-                        get_instance_token_file_name, Labels, sanitize_version)
+                        get_instance_token_file_name, Labels, sanitize_version,
+                        has_triggers)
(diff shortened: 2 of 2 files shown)


osbs/utils.py
```diff
@@ -142,6 +142,18 @@ def graceful_chain_del(d, *args):
         pass
 
 
+def has_triggers(x):
+    if x.get('spec', {}).get('triggers'):
+        return True
+    return False
```
twaugh: Could use `return graceful_chain_get(x, 'spec', 'triggers') is not None`. But also, can we call it `build_config` instead of `x`?

tests/test_utils.py
```diff
@@ -35,6 +36,85 @@ def test_buildconfig_update():
...
 
 
+def has_ist(x):
+    if not has_triggers(x):
+        return False
+    triggers = x['spec']['triggers']
+    if not triggers:
+        return False
+    for t in triggers:
+        if t.get('type', None) == 'ImageChange':
```
twaugh: Don't need the extra parameter to `.get()`: it defaults to `None`.

tests/test_utils.py
```diff
@@ -35,6 +36,85 @@ def test_buildconfig_update():
...
+        return False
+    for t in triggers:
+        if t.get('type', None) != 'ImageChange':
+            return True
+    return False
+
+
+def add_ist(x):
+    if 'triggers' not in x['spec']:
+        x['spec']['triggers'] = []
```
twaugh: Can use `x['spec'].setdefault('triggers', [])`.

tests/test_utils.py
```diff
@@ -35,6 +36,85 @@ def test_buildconfig_update():
...
+    x['spec']['triggers'].append({
+        "type": "ImageChange",
+        "imageChange": {
+            "from": {
+                "kind": "ImageStreamTag",
+                "name": "foobar"
```
twaugh: I generally like having a trailing comma after every dict entry, even the last, as it makes diffs easier to read later if entries are added/removed.

bfontecc007: I hadn't thought about it that way. Makes sense.

tests/test_utils.py
```diff
@@ -35,6 +36,85 @@ def test_buildconfig_update():
     assert x == {'a': 'A', 'strategy': {'b1': 'newB1', 'b3': 'B3', 'b11': {}}, 'c': 'C', 'd': 'D'}
 
 
+def has_ist(buildconfig):
```
lcarva: What if this was `has_trigger(trigger_type)` ?

```python
def has_trigger(buildconfig, trigger_type):
...
    if t.get('type') == trigger_type):
        return True
...
```

I think it'll make it easier to read the tests:
```python
has_trigger(orig, 'ImageChange')
has_trigger(orig, 'Generic')
```

bfontecc007: Is Generic the only anticipated Other Trigger? Or would Generic return anything but ImageChange?

bfontecc007:
Hi guys this will be broken for a minute. Sorry!

bfontecc007:
OK when you guys get a minute, this is no longer broken

csomh:
hey, could you please make sure to add an empty line after the subject line of commit messages, so that this becomes more readable?

```
# git log --oneline -10
7021c27 Add documentation about mock_jsons
2e69a7e get_platform_node_selector: handle node_selector.platform = None
6ec902c support configMaps
565975f Remove Triggers *gets rid of triggers when triggers are removed from prod.json *ImageChange triggers are affec
2ad0159 backup-builder: don't append namespace in the filename
d1497ef janitorial: clean up whitespace and other linter errors
1024905 pass equal_labels configuration to worker
32fc3d8 Handle Null case for triggers in prod JSON
506ec3a build_request: Set 'release' label when provided
4908818 add node_selector.platform to buildConfig
```

PR #608, opened 2017-06-20
Temporarily clear nodeSelector for explicit builds, or set platform nodeselector

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -217,7 +217,7 @@ class OSBS(object):
         # this should never happen, but if it does, we want to know all the builds
         #  that were running at the time
         builds = ', '.join(['%s: %s' % (b.get_build_name(), b.status) for b in builds])
-        msg = 'Multiple builds for %s running, can\'t proceed: %s' % \
+        msg = "Multiple builds for %s running, can't proceed: %s" % \
             (build_config_name, builds)
         return msg
 
@@ -278,7 +278,7 @@ class OSBS(object):
         if rb_len > 0:
             if rb_len == 1:
                 rb = running_builds[0]
-                msg = 'Build %s for %s in state %s, can\'t proceed.' % \
+                msg = "Build %s for %s in state %s, can't proceed." % \
                     (rb.get_build_name(), build_config_name, rb.status)
             else:
                 msg = self._panic_msg_for_more_running_builds(build_config_name, running_builds)
@@ -351,6 +351,9 @@ class OSBS(object):
         build_config_name = build_json['metadata']['name']
         logger.debug('build config to be named "%s"', build_config_name)
         existing_bc = self._get_existing_build_config(build_json)
+        orchestrator_build = False
+        if build_request.spec.platforms.value:
+            orchestrator_build = True
 
         image_stream, image_stream_tag_name = \
             self._get_image_stream_info_for_build_request(build_request)
@@ -371,6 +374,11 @@ class OSBS(object):
                          build_config_name)
             self._verify_no_running_builds(build_config_name)
 
+            # Remove nodeSelector, will be set from build_json
+            # for worker build, if there is platform nodeselector
+            old_nodeselector = existing_bc['spec'].pop('nodeSelector', None)
+            logger.debug("removing build config's nodeSelector %s", old_nodeselector)
+
             utils.buildconfig_update(existing_bc, build_json)
             # Reset name change that may have occurred during
             # update above, since renaming is not supported.
@@ -379,12 +387,9 @@ class OSBS(object):
                          build_config_name)
 
             self.os.update_build_config(build_config_name, json.dumps(existing_bc))
-            if triggers:
-                # Retrieve updated version to pick up lastVersion
-                existing_bc = self._get_existing_build_config(existing_bc)
 
         else:
-            logger.debug('build config for %s doesn\'t exist, creating...',
+            logger.debug("build config for %s doesn't exist, creating...",
                          build_config_name)
             existing_bc = self.os.create_build_config(json.dumps(build_json)).json()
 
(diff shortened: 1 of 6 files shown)


rcerven:
needs tests

osbs/api.py
```diff
@@ -394,7 +396,25 @@ def _create_build_config_and_build(self, build_request):
                                                        scheduled=True)
             logger.debug('Changed parent ImageStreamTag? %s', changed_ist)
 
+        old_nodeselector = existing_bc['spec'].get('nodeSelector')
+        logger.debug('saving build config\'s nodeSelector %s', old_nodeselector)
```
vrutkovs: It may be easier to use double quotes than escape single quotes

osbs/api.py
```diff
@@ -394,7 +396,25 @@ def _create_build_config_and_build(self, build_request):
...
+
+        if old_nodeselector:
+            existing_bc = self._get_existing_build_config(existing_bc)
+
+            existing_bc['spec']['nodeSelector'] = nodeselector
+            logger.debug('setting build config\'s nodeSelector %s', nodeselector)
+            self.os.update_build_config(build_config_name, json.dumps(existing_bc))
```
vrutkovs: Instead of 'fetch BC, update local json and update BC' we might want to use HTTP PATCH - https://docs.openshift.com/container-platform/3.4/rest_api/openshift_v1.html#partially-update-the-specified-buildconfig

osbs/api.py
```diff
@@ -361,6 +361,8 @@ def _create_build_config_and_build(self, build_request):
         # is properly configured.
         triggers = build_json['spec'].pop('triggers', None)
 
+        nodeselector = build_json['spec'].pop('nodeSelector', None)
```
lcarva: Maybe add a comment explaining why `nodeSelector` is being removed?

osbs/api.py
```diff
@@ -394,9 +393,25 @@ def _create_build_config_and_build(self, build_request):
...
+            self.os.patch_build_config(build_config_name, json.dumps(patch_nodeselector))
+
         if triggers:
-            existing_bc['spec']['triggers'] = triggers
-            self.os.update_build_config(build_config_name, json.dumps(existing_bc))
+            patch_triggers = {'spec': {'triggers': triggers}}
+            self.os.patch_build_config(build_config_name, json.dumps(patch_triggers))
 
         if image_stream and triggers:
             prev_version = existing_bc['status']['lastVersion']
```
lcarva: Won't this be incorrect because we no longer have an updated `lastVersion` value?

vrutkovs: Yes, `lastVersion` would change after patching

osbs/api.py
```diff
@@ -394,9 +393,25 @@ def _create_build_config_and_build(self, build_request):
...
+        new_nodeselector = None
+        logger.debug("saving build config's nodeSelector %s", old_nodeselector)
+
+        if old_nodeselector:
+            patch_nodeselector = {'spec': {'nodeSelector': nodeselector}}
+            logger.debug("setting build config's nodeSelector %s", nodeselector)
+            new_nodeselector = nodeselector
+            self.os.patch_build_config(build_config_name, json.dumps(patch_nodeselector))
+
+        elif nodeselector:
```
vrutkovs: Duplicate code

osbs/api.py
lcarva: Hmm, where's `prev_version` value being used?

rcerven: well lastVersion, will not change after patching exactly, but patching will trigger new build and that will update lastVersion

twaugh: OpenShift will want to update lastVersion as part of processing the patch won't it? As Vadim said.

rcerven: updating & patching updates : resourceVersion

triggered new build, will update config's : lastVersion

vrutkovs: Robert is right here, my comment is not detailed enough.

Patching would change `resourceVersion` in metadata - the version of the buildconfig. `lastVersion` here is the number of the last build. But since patching changes the buildconfig it may trigger the build (I'm not sure about it though), and thus changing `lastVersion`

twaugh: Ah! I was getting the two version fields confused.

osbs/api.py
```diff
@@ -378,13 +379,15 @@ def _create_build_config_and_build(self, build_request):
             logger.debug('build config for %s already exists, updating...',
                          build_config_name)
 
+            # Temporarily removing lowpriority nodeselector for explicit build
+            # it will be set back, once we create build
+            old_low_prio_ns = existing_bc['spec'].get('nodeSelector', {}).pop('lowpriority', None)
```
twaugh: Isn't this making an assumption about the value of the node selector?

rcerven: well even if it won't be defined, this will return None

and later we won't be patching it back because it was None

tests/test_api.py
```diff
@@ -1908,3 +1909,106 @@ def test_do_create_prod_build_branch_required(self, osbs, branch_name):
...
                                            customize_conf=customize_conf)
+
+    @pytest.mark.parametrize('lowpriority', ['yes', None])
+    def test_create_build_config_low_priority(self, caplog, lowpriority):
+        # Tests temporarily removing lowpriority nodeselector
+        with NamedTemporaryFile(mode='wt') as fp:
+            fp.write(dedent("""\
+                [general]
+                build_json_dir = inputs
+                """))
```
twaugh: Don't we want a `[default]` section with a `low_priority_node_selector` item configured?

rcerven: no we don't need that, we are just looking on low_priority already set on buildconfig

osbs/api.py
```diff
@@ -395,18 +400,26 @@ def _create_build_config_and_build(self, build_request):
...
-            prev_version = existing_bc['status']['lastVersion']
             build_id = self.os.wait_for_new_build_config_instance(
-                build_config_name, prev_version)
+                build_config_name, existing_bc['status']['lastVersion'])
             build = BuildResponse(self.os.get_build(build_id).json())
         else:
             response = self.os.start_build(build_config_name)
             build = BuildResponse(response.json())
 
+        if orchestrator_build:
```
lcarva: I'm curious as to why we want to handle orchestrator builds differently in this method.

I understand that orchestrator builds should *not* have a platform node selector, 
but that should be handled when rendering the build request, not in this method.

What this method should be doing instead is temporarily clearing the low priority node selector for non scratch builds.
I don't see a reason to distinguish between orchestrator vs worker builds in this method.

osbs/api.py
```diff
@@ -395,18 +400,26 @@ def _create_build_config_and_build(self, build_request):
...
+            patch_nodeselector = {'spec': {
+                                    'nodeSelector': build_request.low_priority_node_selector}}
+            self.os.patch_build_config(build_config_name, json.dumps(patch_nodeselector))
```
lcarva: If there are triggers set and the image stream exists, this will trigger a new build, won't it?
Not sure how to get around that...

lcarva: Scratch that. Updating the node selector on the build config should be fine. It's updates to triggers that can cause this behavior.

rcerven: exactly, adding trigger will trigger build, which happened already few lines above,
so this is just updating build_config with new nodeselector which doesn't trigger anything

osbs/api.py
twaugh: Is it because the orchestrator is the only one that will have triggers? Would a better test be whether the BuildConfig has triggers?

osbs/api.py
```diff
@@ -395,18 +400,26 @@ def _create_build_config_and_build(self, build_request):
             logger.debug('Changed parent ImageStreamTag? %s', changed_ist)
 
         if triggers:
-            existing_bc['spec']['triggers'] = triggers
-            self.os.update_build_config(build_config_name, json.dumps(existing_bc))
+            patch_triggers = {'spec': {'triggers': None}}
+            self.os.patch_build_config(build_config_name, json.dumps(patch_triggers))
+            patch_triggers = {'spec': {'triggers': triggers}}
+            self.os.patch_build_config(build_config_name, json.dumps(patch_triggers))
```
twaugh: This needs a comment to explain what's going on.

rcerven: will add

osbs/api.py
lcarva: Why don't we want worker builds also being low priority during an auto rebuild?

PR #610, opened 2017-06-21
get_platform_node_selector: handle platform of 'none'

mlangsdorf:
get_platform_node_selector() should return None if the platform is None or
if it is 'none'. Update the code to handle the second instance and add a
test case that fails if get_plaform_node_selector does a search for
'node_selector.none = $SOMETHING'

Signed-of-by: Mark Langsdorf <mlangsdo@redhat.com>

### DIFF
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -545,7 +545,7 @@ class Configuration(object):
         :param platform: str, platform to search for, can be null
         """
         nodeselector = None
-        if platform:
+        if platform and platform != 'none':
             nodeselector_str = self._get_value("node_selector." + platform, self.conf_section,
                                                "node_selector." + platform)
             if nodeselector_str:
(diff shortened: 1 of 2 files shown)


twaugh:
I don't think this is quite right. We want to treat the *value* `none` as 'no node selector required'. We won't be using `none` as a platform name.

vrutkovs:
I agree with Tim - we won't be using platform named 'none', so original `if platform` would work in case plaform is actually set to `None`

mlangsdorf:
I'm really confused then.
If platform is None was already handled in the original PR:
  nodeselector = None
  if platform:
     DO stuiff
  return nodeselector

If platform is None, we don't do stuff, and nodeselector is returned as None. So what did I need to change?

mlangsdorf:
okay, I completely misunderstood. Closing this pull request because it's not salvageable.

PR #611, opened 2017-06-21
get_platform_node_selector: handle node_selector.platform = None

mlangsdorf:
If platform is a valid string, but node_selector.platform = None in the
config file, the existing code will do the wrong thing. Make sure that the
parsed item for node_selector.platform has '=' in it or ignore it.

Add test cases for the node_selector.platform = None path.

### DIFF
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -548,7 +548,7 @@ class Configuration(object):
         if platform:
             nodeselector_str = self._get_value("node_selector." + platform, self.conf_section,
                                                "node_selector." + platform)
-            if nodeselector_str:
+            if nodeselector_str and nodeselector_str != 'none':
                 constraints = nodeselector_str.split(',')
                 raw_nodeselector = dict([constraint.split('=', 1) for constraint in constraints])
                 nodeselector = dict([k.strip(), v.strip()] for (k, v) in raw_nodeselector.items())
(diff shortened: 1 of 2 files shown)


osbs/conf.py
```diff
@@ -548,7 +548,7 @@ def get_platform_node_selector(self, platform):
         if platform:
             nodeselector_str = self._get_value("node_selector." + platform, self.conf_section,
                                                "node_selector." + platform)
-            if nodeselector_str:
+            if nodeselector_str and '=' in nodeselector_str:
```
twaugh: I think it would be nice to explicitly compare with `'none'`, just because I think it might be quite easy to mis-type a node selector (e.g. 'amd64' instead of 'cpu=amd64' or whatever).

mlangsdorf: can you give me the key, value pair you want here?
'node_selector.x86_64': 'amd64' will fail with this implementation, because it doesn't have an '=' in it. So I think I have fixed your case but I don't understand what you're asking for.

twaugh: `node_selector.x86_64 = none` should set spec.nodeSelector to null.
`node_selector.x86_64 = amd64` should fail because 'amd64' is not a valid node selector.

mlangsdorf: okay, got it.

PR #612, opened 2017-06-21
Add documentation for node_selector.(platform)

twaugh:
Also allow literal none for low_priority_node_selector, for consistency with how node_selector.(platform) works.

### DIFF
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -509,7 +509,7 @@ class Configuration(object):
         nodeselector = {}
         nodeselector_str = self._get_value("low_priority_node_selector", self.conf_section,
                                            "low_priority_node_selector")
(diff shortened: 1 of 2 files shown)


osbs/conf.py
```diff
@@ -509,7 +509,7 @@ def get_low_priority_node_selector(self):
         nodeselector = {}
         nodeselector_str = self._get_value("low_priority_node_selector", self.conf_section,
                                            "low_priority_node_selector")
-        if nodeselector_str:
+        if nodeselector_str and nodeselector_str != 'none':
```
lcarva: Should this be a case insensitive comparison?

twaugh: I'd rather it wasn't.

PR #613, opened 2017-06-21
render_add_filesystem: improved multi-arch build support

jarodwilson:
- Pass along an architectures variable to atomic-reactor
- Add new filesystem_koji_task_id parameter to create_worker_build API
- Bump DEFAULT_ARRANGEMENT_VERSION to 2
- Update self-tests with filesystem_koji_task_id usage
- Set from_task_id in render_add_filesystem, make use of set_label
- Make sure filesystem_koji_task_id actually propagates to worker builds

Successfully tested with matching atomic-reactor changes in staging
environment, with both the new arrangement version 2 and the prior
(slightly modified) version 1.


osbs/build/build_request.py
```diff
@@ -1110,8 +1116,11 @@ def render(self, validate=True):
 
         koji_task_id = self.spec.koji_task_id.value
         if koji_task_id is not None:
-            self.template['metadata'].setdefault('labels', {})
-            self.template['metadata']['labels']['koji-task-id'] = str(koji_task_id)
+            self.set_label('koji-task-id', str(koji_task_id))
+
+        filesystem_koji_task_id = self.spec.filesystem_koji_task_id.value
```
lcarva: Let's remove this as we agreed to set this label separately.

osbs/core.py
```diff
@@ -500,12 +500,13 @@ def logs(self, build_id, follow=False, build_json=None, wait_if_missing=False):
         return response.content
 
     def list_builds(self, build_config_id=None, koji_task_id=None,
-                    field_selector=None):
+                    filesystem_koji_task_id=None, field_selector=None):
```
lcarva: Let's remove this as we agreed to set this label separately.

tests/test_api.py
```diff
@@ -313,6 +317,7 @@ def test_create_worker_build(self, osbs, inner_template, outer_template,
             'git_ref': TEST_GIT_REF,
             'git_branch': branch,
             'user': TEST_USER,
+            'filesystem_koji_task_id': TEST_FILESYSTEM_KOJI_TASK_ID,
```
lcarva: I think this is the only test that it makes sense to add `filesystem_koji_task_id` to the list of `expected_kwargs`.

Additionally, test_build_request.py could be enhanced to cover changes to `render_add_filesystem`. See if we can leverage the existing test `test_prod_custom_base_image`.

jarodwilson: Okay, I believe I've got this covered now.

lcarva:
@jarodwilson wanna squash the commits?

@twaugh my only hesitation with merging this is that it moves up the default arrangement. I think we're good with that, but it's worth a second look.

jarodwilson:
Commits squashed and re-pushed

twaugh:
Draft release notes for 0.41 updated.

twaugh:
I don't remember why add_labels_in_dockerfile gets removed from the orchestrator build here. The fact that this also changes arrangement version 1 makes me think this crossed over with https://github.com/projectatomic/osbs-client/pull/581 and the removal is accidental.

lcarva:
> I don't remember why add_labels_in_dockerfile gets removed from the orchestrator build here.

It's because this broke base image builds. Because the add_labels_in_dockerfile plugin requires the parent image to be inspected, which is not the case for base image builds.

PR #614, opened 2017-06-22
Deprecate create_prod_build

twaugh:
Maintaining the code for all-in-one builds will become harder
over time. The sooner we can remove create_prod_build(), the
easier we will make future maintenance.


twaugh:
CC @maxamillion

lcarva:
Updated release notes.

maxamillion:
:+1:

PR #616, opened 2017-06-28
add is_cancelled() to build_response.py

maxamillion:

### DIFF
--- a/tests/build_/test_build_response.py
+++ b/tests/build_/test_build_response.py
@@ -56,6 +56,59 @@ class TestBuildResponse(object):
         assert not build_response.cancelled
         assert 'cancelled' in build_response.json['status']
         assert not build_response.json['status'].get('cancelled')
+        assert not build_response.is_cancelled()
+
+    def test_state_checkers(self):
+        build_response = BuildResponse({
+            'status': {
+                'phase': 'Complete'
+            }
+        })
+
+        build_response.status = 'complete'
+        assert build_response.is_finished()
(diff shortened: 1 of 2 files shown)


maxamillion:
@twaugh this doesn't change state. Also, I couldn't find any tests for any of the `BuildResponse` "status checkers" (`is_failed`, `is_running`, `is_pending`, etc). Should I just write new tests and go ahead and cover all of those for good measure?

twaugh:
We have an existing test for cancelling a build -- it could check what this method says before and after.

maxamillion:
@twaugh I see what you mean, I'll get this added. Thanks

tests/build_/test_build_response.py
```diff
@@ -56,6 +56,31 @@ def test_build_cancel(self):
...
+    def test_state_checkers(self):
+        build_response = BuildResponse({
+            'status': {
+                'phase': 'Complete'
+            }
+        })
+
+        build_response.status = 'complete'
+        assert build_response.is_finished()
+        assert build_response.is_succeeded()
```
twaugh: Could also `assert not build_response.is_failed()` etc.

maxamillion: +1 will do

twaugh:
Would be nice to squash commits I think.

maxamillion:
All squashed, thanks all!

twaugh:
Draft release notes for 0.41 updated.

PR #620, opened 2017-06-29
ConfigMap API: better data encoding

mlangsdorf:
Encode the data as json object pairs to handle nested dictionaries.

### DIFF
--- a/osbs/build/config_map_response.py
+++ b/osbs/build/config_map_response.py
@@ -8,6 +8,7 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import print_function, absolute_import, unicode_literals
 
 import logging
+import json
 
 from osbs.utils import graceful_chain_get
 
@@ -34,10 +35,27 @@ class ConfigMapResponse(object):
         """
         Find the data stored in the config_map
 
-        :return: dict, the json data that was passed into the ConfigMap on creation
+        :return: dict, the json of the data data that was passed into the ConfigMap on creation
         """
         data = graceful_chain_get(self.json, "data")
         if data is None:
             return {}
 
-        return data
+        data_dict = {}
+        for key in data:
+            data_dict[key] = json.loads(data[key])
+
+        return data_dict
+
+    def get_data_by_key(self, name):
+        """
+        Find the object stored by a JSON string at key 'name'
+
+        :return: str or dict, the json of the str or dict stored in the ConfigMap at that location
+        """
+        data = graceful_chain_get(self.json, "data")
+
+        if data is None or name not in data:
+            return {}
+
+        return json.loads(data[name])
--- a/tests/mock_jsons/0.5.4/create_config_map.json
+++ b/tests/mock_jsons/0.5.4/create_config_map.json
@@ -1,8 +1,8 @@
 {
     "apiVersion": "v1",
     "data": {
-        "special.how": "very",
-        "special.type": "charm"
+        "special.how": "\"very\"",
+        "special.type": "{\"quark\": \"charm\"}"
     },
     "kind": "ConfigMap",
     "metadata": {
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1913,29 +1913,34 @@ class TestOSBS(object):
                                            customize_conf=customize_conf)
 
     # osbs is a fixture here
-    @pytest.mark.parametrize(('name', 'data'), [  # noqa
-        ('special-config',
-         {
-             "special.how": "very",
-             "special.type": "charm"
-         }),
-    ])
-    def test_config_map(self, osbs, name, data):
-        obj = {
-            'apiVersion': 'v1',
-            'kind': 'ConfigMap',
-            'metadata': {'name': name},
-            'data': data,
-        }
-        (flexmock(osbs.os)
-            .should_call('create_config_map')
-            .with_args(obj))
+    def test_config_map(self, osbs): # noqa
+        with open(os.path.join(INPUTS_PATH, "config_map.json")) as fp:
+            raw = fp.read()
+        mock = flexmock(sys.modules['__builtin__' if six.PY2 else 'builtins'])
+        mock.should_call('open')  # set the fall-through
+        (mock.should_receive('open')
+            .with_args('inputs/config_map.json')
+            .and_return(flexmock(read=lambda: raw)))
 
+        how_str = "special.how"
+        very_str = "very"
+        type_str = "special.type"
+        quark_dict = {"quark": "charm"}
+        data = {
+                how_str: very_str,
+                type_str: quark_dict
+        }
+        none_str = "special.none"
+        name = 'special-config'
         config_map = osbs.create_config_map(name, data)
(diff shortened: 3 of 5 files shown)


osbs/build/config_map_response.py
```diff
@@ -34,10 +35,23 @@ def get_data(self):
...
+        Find the data stored in the config_map
+
+        :return: json, the json version of the json data that was passed into the ConfigMap on
+                 creation
+        """
+        data = graceful_chain_get(self.json, "data")
+        if data is None:
+            return {}
+
+        return json.dumps(data)
```
twaugh: I think we want `loads` here, i.e.:
"""
Return the object stored in the config map

:param name: str, key name
:return: object, the object described by a JSON string at key 'name'
""""

data = ...

return json.loads(data[name])

tests/test_api.py
```diff
@@ -1925,7 +1925,7 @@ def test_config_map(self, osbs, name, data):
             'apiVersion': 'v1',
             'kind': 'ConfigMap',
             'metadata': {'name': name},
-            'data': data,
+            'data': {key: json.dumps(value) for key, value in data.items()},
```
vrutkovs: The test is duplicating functionality. Instead of mocking `create_config_map` it should mock file open and its contents:
```
mock = flexmock(sys.modules['__builtin__'])
mock.should_call('open')  # set the fall-through
(mock.should_receive('open')
    .with_args('/your/file')
    .and_return( flexmock(read=lambda: 'file contents') ))
```

osbs/build/config_map_response.py
mlangsdorf: do I test that by doing
assert config_map.get_json("special.how") == "very"
or something else?

tests/test_api.py
mlangsdorf: Thanks for clarifying this, but I get travis build failures against __builtin__ when I do this.

vrutkovs: Ah, yes, it should be 'builtins' for Py3

vrutkovs: This should resolve it:
```
mock = flexmock(sys.modules['__builtin__' if six.PY2 else 'builtins'])
```

vrutkovs: Improved part of the code, which uses actual content of the of :
```diff
-        obj = {
-            'apiVersion': 'v1',
-            'kind': 'ConfigMap',
-            'metadata': {}
-        }
+        with open(os.path.join(INPUTS_PATH, "config_map.json")) as fp:
+            obj = json.load(fp)
```

Also we probably should be storing filename in a variable in `tests/constants.py` and use it in this test

osbs/build/config_map_response.py
```diff
@@ -34,10 +35,23 @@ def get_data(self):
...
+        Find the object stored by a JSON string at key 'name'
+
+        :return: json, the json version of the json data that was passed into the ConfigMap on
+                 creation
+        """
+        data = graceful_chain_get(self.json, "data")
+        if data is None:
+            return {}
+
+        return json.loads(data[name])
```
twaugh: This method doesn't have test coverage yet.

osbs/build/config_map_response.py
twaugh: Well, we should improve the test data so that it tries to store structured data there. In mock_jsons/*/create_config_map.json we'd want:

```json
...
"data": {
    "metadata.json": "{\"key\": \"value\"}",
    "special.how": "\"very\"",
    "special.type": "\"charm\""
}
...
```

and `test_config_map()` should be given the additional "metadata.json" data to attempt to store.

Finally, `test_config_map()` should verify the behaviour of `config_map.get_json()`, i.e. that you get back the same objects you put in, for each key ("metadata.json", "special.how", whatever).

tests/mock_jsons/0.5.4/create_config_map.json
```diff
@@ -2,7 +2,7 @@
     "apiVersion": "v1",
     "data": {
         "special.how": "very",
-        "special.type": "charm"
+        "special.type": {"quark": "charm"}
```
twaugh: This doesn't look right.

The *actual ConfigMap* object that OpenShift stories is key:val (str:str).

Our create_config_map(name, data) API stores data objects in that str value by JSON encoding them.

So I would expect this to look like:
```
"special.type": "{\"quark\": \"charm\"}"
```

tests/test_api.py
```diff
@@ -1913,29 +1913,32 @@ def test_do_create_prod_build_branch_required(self, osbs, branch_name):
...
+            'metadata': {}
         }
-        (flexmock(osbs.os)
-            .should_call('create_config_map')
-            .with_args(obj))
+        mock = flexmock(sys.modules['__builtin__' if six.PY2 else 'builtins'])
+        mock.should_call('open')  # set the fall-through
+        (mock.should_receive('open')
+            .with_args('inputs/config_map.json')
+            .and_return(flexmock(read=lambda: json.dumps(obj))))
```
lcarva: Can't we return the real template json here?

tests/test_api.py
```diff
@@ -1913,29 +1913,32 @@ def test_do_create_prod_build_branch_required(self, osbs, branch_name):
...
 
+        data = {
```
lcarva: I think we also want to explicitly define what this data should look like when its json serialized.
Then verify that when it's created, it matches the serialized version.

osbs/build/config_map_response.py
```diff
@@ -30,14 +31,27 @@ def __init__(self, config_map):
...
+        Find the object stored by a JSON string at key 'name'
+
+        :return: str or dict, the json of the str or dict stored in the ConfigMap at that location
+        """
+        data = graceful_chain_get(self.json, "data")
+
+        if data is None or name not in data:
+            return {}
+
+        return data[name]
```
twaugh: `return json.loads(data[name])`

tests/test_api.py
mlangsdorf: I'm not sure what you mean. What is the real template json in this case?

lcarva: The actual file under inputs/config_map.json at the root of this project.

mlangsdorf: so I have that file as fp from here: 
   with open(os.path.join(INPUTS_PATH, "config_map.json")) as fp:

I can't just pass it to .and_return like this:
   .and_return(flexmock(read=lambda: fp.read() ))

because it's not a json object that that point. 

So I'm still confused here.

lcarva: `lambda: json.dumps(obj)` returns a string, not a json object.
`fp.read()` also returns a string.

mlangsdorf: okay, got it.

osbs/build/config_map_response.py
```diff
@@ -8,6 +8,7 @@
 from __future__ import print_function, absolute_import, unicode_literals
 
 import logging
+import json
```
mlangsdorf: Fixed

lcarva:
Update release notes.

PR #625, opened 2017-07-04
pass artifacts_allowed_domains to workers

rcerven:


osbs/build/build_request.py
```diff
@@ -249,6 +249,7 @@ def render_orchestrate_build(self):
             'sources_command': self.spec.sources_command.value,
             'vendor': self.spec.vendor.value,
             'equal_labels': equal_labels_string,
+            'artifacts_allowed_domains': ','.join(self.spec.artifacts_allowed_domains.value or []),
```
vrutkovs: Should have `or None` too, so that this option would not be set to `''` if `.value` is `None`

rcerven: fixed

PR #627, opened 2017-07-04
Fix params passed to plugins when yum proxy is set

vrutkovs:


twaugh:
Good catch. The unit test is also incorrect.

vrutkovs:
Right, updated the unittests

osbs/build/build_request.py
```diff
@@ -563,7 +563,7 @@ def render_add_filesystem(self):
             self.dj.dock_json_set_arg(phase, plugin, 'koji_hub',
                                       self.spec.kojihub.value)
             if self.spec.proxy.value:
-                self.dj.dock_json_set_arg(phase, plugin, 'koji_proxyuser',
+                self.dj.dock_json_set_arg(phase, plugin, 'proxy',
```
lcarva: `proxy` is not a valid param for `add_filesystem` plugin.

vrutkovs:
Right, removed passing this param entirely. `koji_proxyuser` doesn't seem to used anywhere

lcarva:
Not adding to release notes as this seems to be obscure enough and not impact any users.
Please add it to it if you think otherwise.

PR #632, opened 2017-07-07
Don't run fetch_worker_metadata for scratch builds

twaugh:

No unit tests because its presence will be tested properly once the new arrangement including the plugin is created.


twaugh:
Actually I'll close this as a more complete PR is in the works elsewhere.

PR #634, opened 2017-07-11
Additional arrangement testing

twaugh:
Because the point of the test_arrangement.py tests is to verify the renderer changes are correct, and it knows where the templates are, it can also check whether there are missing tests.

Additionally, it can also catch the case where the default arrangement version has been incremented but there is no testing for the new arrangement.


lcarva:
Not adding this to release notes as it only impacts unit tests.
Please update if you feel otherwise.

PR #635, opened 2017-07-11
constants: Add logging format for atomic_reactor

dzickusrh:
OSBS client will need the atomic_reactor logs in a specific format
in order to parse them correctly.  Control that in osbs-client and
make it available for atomic_reactor to import.


twaugh:
I don't think this needs a release notes update.

PR #638, opened 2017-07-21
Build type nodeselectors for worker builds

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -293,9 +293,6 @@ class OSBS(object):
         build_json['spec']['serviceAccount'] = 'builder'
         build_json['metadata']['labels']['scratch'] = 'true'
 
-        if build_request.low_priority_node_selector:
-            build_json['spec']['nodeSelector'].update(build_request.low_priority_node_selector)
-
         builder_img = build_json['spec']['strategy']['customStrategy']['from']
         kind = builder_img['kind']
         if kind == 'ImageStreamTag':
@@ -367,6 +364,10 @@ class OSBS(object):
                          build_config_name)
             self._verify_no_running_builds(build_config_name)
 
+            # Remove nodeSelector, will be set from build_json for worker build
+            old_nodeselector = existing_bc['spec'].pop('nodeSelector', None)
+            logger.debug("removing build config's nodeSelector %s", old_nodeselector)
+
             utils.buildconfig_update(existing_bc, build_json)
             # Reset name change that may have occurred during
             # update above, since renaming is not supported.
@@ -423,6 +424,7 @@ class OSBS(object):
                               arrangement_version=None,
                               filesystem_koji_task_id=None,
                               koji_upload_dir=None,
+                              is_auto=False,
                               **kwargs):
         repo_info = utils.get_repo_info(git_uri, git_ref, git_branch=git_branch)
         df_parser = repo_info.dockerfile_parser
@@ -508,9 +510,12 @@ class OSBS(object):
             arrangement_version=arrangement_version,
             info_url_format=self.build_conf.get_info_url_format(),
             artifacts_allowed_domains=self.build_conf.get_artifacts_allowed_domains(),
-            low_priority_node_selector=self.build_conf.get_low_priority_node_selector(),
             equal_labels=self.build_conf.get_equal_labels(),
             platform_node_selector=self.build_conf.get_platform_node_selector(platform),
+            scratch_build_node_selector=self.build_conf.get_scratch_build_node_selector(),
+            explicit_build_node_selector=self.build_conf.get_explicit_build_node_selector(),
+            auto_build_node_selector=self.build_conf.get_auto_build_node_selector(),
+            is_auto=is_auto,
             filesystem_koji_task_id=filesystem_koji_task_id,
             koji_upload_dir=koji_upload_dir,
             platform_descriptors=self.build_conf.get_platform_descriptors(),
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -61,7 +61,10 @@ class BuildRequest(object):
         # For the koji "scratch" build type
         self.scratch = None
         self.base_image = None
-        self.low_priority_node_selector = None
+        self.scratch_build_node_selector = None
+        self.explicit_build_node_selector = None
+        self.auto_build_node_selector = None
+        self.is_auto = None
         # forward reference
         self.platform_node_selector = None
         self.platform_descriptors = None
@@ -90,9 +93,12 @@ class BuildRequest(object):
         :param distribution_scope: str, distribution scope for this image
                                    (private, authoritative-source-only, restricted, public)
         :param use_auth: bool, use auth from atomic-reactor?
-        :param low_priority_node_selector: dict, a nodeselector for builds with lower priority
         :param platform_node_selector: dict, a nodeselector for a specific platform
         :param platform_descriptors: dict, platforms and their archiectures and enable_v1 settings
+        :param scratch_build_node_selector: dict, a nodeselector for scratch builds
+        :param explicit_build_node_selector: dict, a nodeselector for explicit builds
+        :param auto_build_node_selector: dict, a nodeselector for auto builds
+        :param is_auto: bool, indicates if build is auto build
         """
 
         # Here we cater to the koji "scratch" build type, this will disable
@@ -103,9 +109,12 @@ class BuildRequest(object):
             pass
 
         self.base_image = kwargs.get('base_image')
-        self.low_priority_node_selector = kwargs.get('low_priority_node_selector')
         self.platform_node_selector = kwargs.get('platform_node_selector', {})
         self.platform_descriptors = kwargs.get('platform_descriptors', {})
+        self.scratch_build_node_selector = kwargs.get('scratch_build_node_selector', {})
+        self.explicit_build_node_selector = kwargs.get('explicit_build_node_selector', {})
+        self.auto_build_node_selector = kwargs.get('auto_build_node_selector', {})
+        self.is_auto = kwargs.get('is_auto', False)
 
         logger.debug("setting params '%s' for %s", kwargs, self.spec)
         self.spec.set_params(**kwargs)
@@ -1158,6 +1167,24 @@ class BuildRequest(object):
         # !IMPORTANT! can't be too long: https://github.com/openshift/origin/issues/733
         self.template['metadata']['name'] = name
 
+    def render_node_selectors(self):
+        # for worker builds set nodeselectors
+        if self.spec.platforms.value is None:
+
+            # auto or explicit build selector
+            if self.is_auto:
+                self.template['spec']['nodeSelector'] = self.auto_build_node_selector
+            # scratch build nodeselector
+            elif self.scratch:
+                self.template['spec']['nodeSelector'] = self.scratch_build_node_selector
+            # explicit build nodeselector
+            else:
+                self.template['spec']['nodeSelector'] = self.explicit_build_node_selector
+
+            # platform nodeselector
+            if self.platform_node_selector:
+                self.template['spec']['nodeSelector'].update(self.platform_node_selector)
+
     def render(self, validate=True):
         if validate:
             self.spec.validate()
@@ -1300,8 +1327,6 @@ class BuildRequest(object):
         if koji_task_id is not None:
             self.set_label('koji-task-id', str(koji_task_id))
 
-        self.template['spec']['nodeSelector'] = self.platform_node_selector
-
         use_auth = self.spec.use_auth.value
         self.render_reactor_config()
         self.render_orchestrate_build()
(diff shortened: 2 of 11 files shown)


osbs/api.py
```diff
@@ -602,11 +612,18 @@ def create_worker_build(self, **kwargs):
         for required in ('platform', 'release', 'arrangement_version'):
             if not kwargs.get(required):
                 missing.add(required)
-
         if missing:
             raise ValueError("Worker build missing required parameters: %s" %
                              missing)
 
+        extra = set()
+        for unwanted in ('platforms'):
```
vrutkovs: We're interating over a tuple with single item, why?

rcerven: well it is prepared for future changes if there would be more unwanted things, and same as in orchestrator

osbs/api.py
```diff
@@ -644,6 +661,14 @@ def create_orchestrator_build(self, **kwargs):
             raise OsbsValidationException("can't create orchestrate build "
                                           "when can_orchestrate isn't enabled")
 
+        extra = set()
+        for unwanted in ('platform', 'release'):
+            if kwargs.get(unwanted):
+                extra.add(unwanted)
+        if extra:
```
vrutkovs: `extra = [x for x in kwargs.keys() if x in ['platform', 'release']]` ?

bfontecc007:
LGTM!

osbs/api.py
rcerven: I can't use that because the keys are defined but None

osbs/api.py
```diff
@@ -395,13 +393,21 @@ def _create_build_config_and_build(self, build_request):
             logger.debug('Changed parent ImageStreamTag? %s', changed_ist)
 
         if triggers:
-            existing_bc['spec']['triggers'] = triggers
-            self.os.update_build_config(build_config_name, json.dumps(existing_bc))
```
lcarva: Is this change needed? I'd rather not mess around with this method.

osbs/api.py
```diff
@@ -395,13 +393,21 @@ def _create_build_config_and_build(self, build_request):
...
+            # so in order to completely replace 'triggers',
+            # we have to first empty it by setting it 'null'
+            # and after we can add triggers
+            patch_triggers = {'spec': {'triggers': None}}
+            self.os.patch_build_config(build_config_name, json.dumps(patch_triggers))
+            patch_triggers = {'spec': {'triggers': triggers}}
+            self.os.patch_build_config(build_config_name, json.dumps(patch_triggers))
 
         if image_stream and triggers:
-            prev_version = existing_bc['status']['lastVersion']
```
lcarva: Is this change needed? I'd rather not mess around with this method.

osbs/api.py
```diff
@@ -602,11 +612,13 @@ def create_worker_build(self, **kwargs):
         for required in ('platform', 'release', 'arrangement_version'):
             if not kwargs.get(required):
                 missing.add(required)
-
         if missing:
             raise ValueError("Worker build missing required parameters: %s" %
                              missing)
 
+        if kwargs.get('platforms'):
```
lcarva: I agree with this change, but it seems unrelated to node selectors. Can we split up into different commits?

osbs/api.py
```diff
@@ -643,6 +655,13 @@ def create_orchestrator_build(self, **kwargs):
         if not self.can_orchestrate():
             raise OsbsValidationException("can't create orchestrate build "
                                           "when can_orchestrate isn't enabled")
+        extra = set()
+        for unwanted in ('platform', 'release'):
```
lcarva: I agree with this change, but it seems unrelated to node selectors. Can we split up into different commits?

osbs/build/build_request.py
```diff
@@ -1186,7 +1195,22 @@ def render(self, validate=True):
         if koji_task_id is not None:
             self.set_label('koji-task-id', str(koji_task_id))
 
-        self.template['spec']['nodeSelector'] = self.platform_node_selector
+        # for worker builds set nodeselectors
```
lcarva: Let's move this bonanza to its own method, maybe `render_node_selectors`.

osbs/build/build_request.py
```diff
@@ -1186,7 +1195,22 @@ def render(self, validate=True):
...
+        if self.spec.platforms.value is None:
```
lcarva: I'd rather this was a check for `platform`, not `platforms`. There's also a `build_type` spec value now, too. It's probably better to use `build_type` because we really are checking the build type here. Checking `platform` or `platforms` is just a heuristic for achieving that.

osbs/build/build_request.py
```diff
@@ -1186,7 +1195,22 @@ def render(self, validate=True):
...
+                self.template['spec']['nodeSelector'] = self.auto_build_node_selector
+            else:
+                self.template['spec']['nodeSelector'] = self.explicit_build_node_selector
+
+            # platform nodeselector
+            if self.platform_node_selector:
+                self.template['spec']['nodeSelector'].update(self.platform_node_selector)
+
+            # scratch build nodeselector
+            if self.scratch:
```
lcarva: Would this add both `scratch_build_node_selector` and either `explicit_build_node_selector` or `auto_build_node_selector`.
Be definition, you can't have a scratch autorebuild. Maybe this:

```python
if self.is_scratch:
  ... # scratch build
elif self.is_auto:
  ... # auto build
else:
  ... # explicit build

File: osbs/api.py (hunk 4 above)
Comment (rcerven) [inline osbs/api.py]: it isn't needed, as the lastVersion is still same

File: osbs/build/build_request.py (hunk 8 above)
Comment (rcerven) [inline osbs/build/build_request.py]: platform isn't passed to method only platforms, 'build_type' ??? where is it?

File: osbs/build/build_request.py (hunk 9 above)
Comment (rcerven) [inline osbs/build/build_request.py]: that would add only scratch and not 'scratch & explicit'

yeah I know that by definition you can't have that

File: osbs/build/build_request.py (hunk 8 above)
Comment (rcerven) [inline osbs/build/build_request.py]: and we have already our logic based on platforms

File: osbs/build/build_request.py (hunk 7 above)
Comment (rcerven) [inline osbs/build/build_request.py]: moved to it's own method

File: osbs/api.py (hunk 6 above)
Comment (rcerven) [inline osbs/api.py]: split into 2 commits

File: osbs/api.py (hunk 3 above)
Comment (rcerven) [inline osbs/api.py]: well the code isn't needed, because we are using patch now

Comment (lcarva) [inline osbs/api.py]: Let's remove this change then.

File: osbs/api.py (hunk 4 above)
Comment (lcarva) [inline osbs/api.py]: Let's remove this change then.

File: osbs/api.py (hunk 3 above)
Comment (lcarva) [inline osbs/api.py]: TBH I don't see how switching to patch makes this better. In either case, it's completely irrelevant to this PR.

File: osbs/core.py [hunk 10]
```diff
@@ -200,6 +200,10 @@ def _put(self, url, with_auth=True, **kwargs):
         headers, kwargs = self._request_args(with_auth, **kwargs)
         return self._con.put(url, headers=headers, verify_ssl=self.verify_ssl, **kwargs)
 
+    def _patch(self, url, with_auth=True, **kwargs):
```
vrutkovs: Tests don't hit those lines

osbs/core.py
```diff
@@ -373,6 +377,13 @@ def update_build_config(self, build_config_id, build_config_json):
         check_response(response)
         return response
 
+    def patch_build_config(self, build_config_id, build_config_json):
```
vrutkovs: Tests don't hit those lines

osbs/api.py
vrutkovs: Right, there already is `clean_triggers` in `utils.py`.

The comment should be added why are triggers being replaced at all - it seems pointless here

rcerven: clean triggers is cleaning not ImageChange triggers from existing config

this lines were suppose to update buildconfig with new triggers

osbs/conf.py
```diff
@@ -554,3 +543,36 @@ def get_platform_node_selector(self, platform):
                 nodeselector = dict([k.strip(), v.strip()] for (k, v) in raw_nodeselector.items())
 
         return nodeselector
+
+    def get_scratch_build_node_selector(self):
+        nodeselector = {}
+        nodeselector_str = self._get_value("scratch_build_node_selector", self.conf_section,
+                                           "scratch_build_node_selector")
+        if nodeselector_str and nodeselector_str != 'none':
```
vrutkovs: This piece of code is being copy-pasted across three functions

tests/build_/test_build_request.py
```diff
@@ -2105,3 +2105,70 @@ def test_info_url_format(self):
...
+                'explicit2': 'yes'
+            },
+            'scratch': {
+                'scratch1': 'yes',
+                'scratch2': 'yes'
+            }
+        }
+        nodeselector_expected = {}
+        if not platforms:
+            if platform:
```
vrutkovs: This is repeating the logic of the app. Instead `nodeselector_expected` should be passed as a param to the function

osbs/conf.py
rcerven: yeah as many other methods in conf.py, so I thought is better keep it nice and separate

tests/build_/test_build_request.py
rcerven: so I should write manually all combinations?

vrutkovs: yes, it would be easier to review

osbs/conf.py
vrutkovs: Other methods in `conf.py` are one liners, this one is 7 lines. Whenever the format of the selectors needs to be changed 3 functions would require an update

osbs/build/build_request.py
lcarva: Then let's change the code to reflect that.

osbs/conf.py
lcarva: +1 for making a little helper method to handle parsing of the raw value.

tests/build_/test_arrangements.py
```diff
@@ -510,6 +510,7 @@ class TestArrangementV3(TestArrangementV2):
                 'add_labels_in_dockerfile',
                 'koji_parent',
                 'reactor_config',
+                'check_and_set_rebuild',
```
lcarva: See the comment a few lines above: `# Changing this? Add test methods`

osbs/conf.py
```diff
@@ -547,22 +536,50 @@ def get_equal_labels(self):
...
+    def get_scratch_build_node_selector(self):
+        nodeselector_str = self._get_value("scratch_build_node_selector", self.conf_section,
+                                           "scratch_build_node_selector")
+        nodeselector = self.generate_nodeselector_dict(nodeselector_str)
+        return nodeselector
+
+    def get_explicit_build_node_selector(self):
+        nodeselector_str = self._get_value("explicit_build_node_selector", self.conf_section,
+                                           "explicit_build_node_selector")
+        nodeselector = self.generate_nodeselector_dict(nodeselector_str)
```
vrutkovs: Nitpick: `return self.generate_nodeselector_dict(nodeselector_str)`?

rcerven: changed

osbs/build/build_request.py
```diff
@@ -1138,6 +1147,23 @@ def render_customizations(self):
...
+        if self.spec.platforms.value is None:
+
+            # auto or explicit build selector
+            if self.is_auto:
+                self.template['spec']['nodeSelector'] = self.auto_build_node_selector
+            else:
+                self.template['spec']['nodeSelector'] = self.explicit_build_node_selector
+                # scratch build nodeselector
+                if self.scratch:
+                    self.template['spec']['nodeSelector'].update(self.scratch_build_node_selector)
```
lcarva: why do we want scratch and explicit node selectors for the same build?
The intent was to have either one of the three types set, not a combination of some.

tests/test_conf.py
```diff
@@ -492,3 +493,28 @@ def test_get_registry_api_versions(self, platform, config, expected, valid):
...
+        ({},
+         {}),
+        ({'default': {}},
+         {}),
+        ({'default': {'node_selector': 'breakfast=eggs.com'}},
+         {'breakfast': 'eggs.com'}),
+        ({'default': {'node_selector': 'breakfast=eggs.com, lunch=ham.com'}},
+         {'breakfast': 'eggs.com', 'lunch': 'ham.com'}),
+    ])
+    def test_get_lowprio_node_selectors(self, nodeselector_type, config, expected):
```
lcarva: This method should be renamed.

osbs/build/build_request.py
```diff
@@ -1138,6 +1147,24 @@ def render_customizations(self):
...
         self.dj.dock_json_set_param('client_version', client_version)
 
+    def render_node_selectors(self):
+        # for worker builds set nodeselectors
+        if self.spec.platforms.value is None:
+
+            # auto or explicit build selector
+            if self.is_auto:
+                self.template['spec']['nodeSelector'] = self.auto_build_node_selector
+            else:
```
lcarva: This could be an `elif self.scratch` to reduce the nested statements.

lcarva:
Updated release notes.

PR #639, opened 2017-07-21
Implement retries in osbs-client's HTTP requests

vrutkovs:
Initial version of retries in http.py

How to test:
 * `sed -i '557s/.*/            response.status = 500 if url.startswith("\/oapi") and method == "GET" and retries._observed_errors == 0 else response.status/' /usr/lib/python2.7/site-packages/urllib3/connectionpool.py` in buildroot and on builder
 * start a build via osbs
 * the build should pass and buildlog should have warning that Openshift requests were retried once, as first request would always return HTTP 500

TODO:
 * [x] Verify that it works on Py3 and Fedora
 * [x] Tests
 * [x] contextmanager with `retries_disabled`
 * [ ] Rewrite exising code using contextmanager

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -16,6 +16,7 @@ import sys
 import warnings
 import getpass
 from functools import wraps
+from contextlib import contextmanager
 
 from osbs.build.build_request import BuildRequest
 from osbs.build.build_response import BuildResponse
@@ -1024,3 +1025,13 @@ class OSBS(object):
         """
         response = self.os.delete_config_map(name)
         return response
+
+    @contextmanager
+    def retries_disabled(self):
+        """
+        Context manager to disable retries on requests
+        :returns: OSBS object
+        """
+        self.os.retries_enabled = False
+        yield
+        self.os.retries_enabled = True
--- a/osbs/constants.py
+++ b/osbs/constants.py
@@ -73,3 +73,11 @@ ATOMIC_REACTOR_LOGGING_FMT = \
     '%(asctime)s platform:%(arch)s - %(name)s - %(levelname)s - %(message)s'
 
 REPO_CONFIG_FILE = '.osbs-repo-config'
+
+# number of retries for http requests
+HTTP_MAX_RETRIES = 3
+
+# how many seconds should request wait for in case non-critical error has occurred
+HTTP_BACKOFF_FACTOR = 2
+
+HTTP_RETRIES_STATUS_FORCELIST = [408, 500, 502, 503, 504]
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -83,8 +83,7 @@ class Openshift(object):
                  verbose=False, username=None, password=None, use_kerberos=False,
                  kerberos_keytab=None, kerberos_principal=None, kerberos_ccache=None,
                  client_cert=None, client_key=None, verify_ssl=True, use_auth=None,
-                 token=None,
-                 namespace=DEFAULT_NAMESPACE):
+                 token=None, namespace=DEFAULT_NAMESPACE):
         self.os_api_url = openshift_api_url
         self.k8s_api_url = k8s_api_url
         self._os_api_version = openshift_api_version
@@ -93,6 +92,7 @@ class Openshift(object):
         self.verbose = verbose
         self.verify_ssl = verify_ssl
         self._con = HttpSession(verbose=self.verbose)
+        self.retries_enabled = True
 
         # auth stuff
         self.use_kerberos = use_kerberos
@@ -190,20 +190,27 @@ class Openshift(object):
 
     def _post(self, url, with_auth=True, **kwargs):
         headers, kwargs = self._request_args(with_auth, **kwargs)
-        return self._con.post(url, headers=headers, verify_ssl=self.verify_ssl, **kwargs)
+        return self._con.post(
+            url, headers=headers, verify_ssl=self.verify_ssl,
+            retries_enabled=self.retries_enabled, **kwargs)
 
     def _get(self, url, with_auth=True, **kwargs):
         headers, kwargs = self._request_args(with_auth, **kwargs)
-        return self._con.get(url, headers=headers, verify_ssl=self.verify_ssl, **kwargs)
+        return self._con.get(
+            url, headers=headers, verify_ssl=self.verify_ssl,
+            retries_enabled=self.retries_enabled, **kwargs)
 
     def _put(self, url, with_auth=True, **kwargs):
         headers, kwargs = self._request_args(with_auth, **kwargs)
-        return self._con.put(url, headers=headers, verify_ssl=self.verify_ssl, **kwargs)
+        return self._con.put(
+            url, headers=headers, verify_ssl=self.verify_ssl,
+            retries_enabled=self.retries_enabled, **kwargs)
 
     def _delete(self, url, with_auth=True, **kwargs):
         headers, kwargs = self._request_args(with_auth, **kwargs)
-        return self._con.delete(url, headers=headers, verify_ssl=self.verify_ssl,
-                                **kwargs)
+        return self._con.delete(
+            url, headers=headers, verify_ssl=self.verify_ssl,
+            retries_enabled=self.retries_enabled, **kwargs)
 
     def get_oauth_token(self):
         url = self.os_oauth_url + "?response_type=token&client_id=openshift-challenging-client"
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -23,8 +23,13 @@ except ImportError:
 
 
 from osbs.exceptions import OsbsException, OsbsNetworkException, OsbsResponseException
+from osbs.constants import (
+    HTTP_MAX_RETRIES, HTTP_BACKOFF_FACTOR, HTTP_RETRIES_STATUS_FORCELIST)
 
 import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util import Retry
+from requests.exceptions import HTTPError, RetryError
 from requests.utils import guess_json_utf
 try:
     from requests_kerberos import HTTPKerberosAuth
@@ -59,7 +64,10 @@ class HttpSession(object):
             with stream as s:
                 content = s.req.content
                 return HttpResponse(s.status_code, s.headers, content)
-        except requests.exceptions.HTTPError as ex:
+        except RetryError as ex:
+            raise OsbsNetworkException(url, str(ex), '',
+                                       cause=ex, traceback=sys.exc_info()[2])
+        except HTTPError as ex:
             raise OsbsNetworkException(url, str(ex), ex.response.status_code,
                                        cause=ex, traceback=sys.exc_info()[2])
         except Exception as ex:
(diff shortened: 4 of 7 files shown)


osbs/constants.py
```diff
@@ -71,3 +71,11 @@
...
 ATOMIC_REACTOR_LOGGING_FMT = \
     '%(asctime)s platform:%(arch)s - %(name)s - %(levelname)s - %(message)s'
+
+# number of retries for http requests
+HTTP_MAX_RETRIES = 3
+
+# how many seconds should request wait for in case non-critical error has occurred
+HTTP_BACKOFF_FACTOR = 2
+
+HTTP_RETRIES_STATUS_FORCELIST = [500, 502, 503, 504]
```
lcarva: 408 (Request Timeout) and 409 (Conflict) also?

osbs/http.py
```diff
@@ -59,8 +82,24 @@ def request(self, url, *args, **kwargs):
             with stream as s:
                 content = s.req.content
                 return HttpResponse(s.status_code, s.headers, content)
-        except requests.exceptions.HTTPError as ex:
-            raise OsbsNetworkException(url, str(ex), ex.response.status_code,
+        except (HTTPError, RetryError) as ex:
```
lcarva: Two distinct exception blocks would probably be better here.

osbs/constants.py
lcarva: Although for conflict you may need to fetch the more recent version of the object being updated. Or maybe not.

osbs/http.py
```diff
@@ -34,6 +41,22 @@
...
 
+class OSBSRetry(Retry):
+    def is_forced_retry(self, method, status_code, has_retry_after=False):
+        if status_code in HTTP_RETRIES_STATUS_FORCELIST:
+            logger.debug("Server returned returned %s", status_code)
+            return True
+        else:
+            return False
+
+    def increment(self, method=None, url=None, response=None,
```
lcarva: Is this being overwritten just so we can log? Sure there must be another way.

osbs/http.py
```diff
@@ -34,6 +41,22 @@
 logger = logging.getLogger(__name__)
 
 
+class OSBSRetry(Retry):
+    def is_forced_retry(self, method, status_code, has_retry_after=False):
```
lcarva: Looks like the constructor for `Retry` takes `status_forcelist` which is used for setting the list of status codes which should trigger a retry: http://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry

osbs/http.py
```diff
@@ -59,8 +82,24 @@ def request(self, url, *args, **kwargs):
...
+            if isinstance(ex, RetryError):
+                # RetryError have neither response nor status code set
+                reason = ex.args[0].reason
+                # Try to match it, as ancient python-requests in rhel is ancient
+                if isinstance(reason, ResponseError):
+                    if requests.__version__.startswith('2.6.'):
+                        regexp = r"too many (\d+) error responses"
+                        match = re.search(regexp, reason.message)
+                        if match and len(match.group()) > 0:
+                            status_code = match.group(1)
```
lcarva: Why go through all this trouble?
At this point, this is a fatal failure. Wouldn't the stack trace provide all the needed information for further debugging?

osbs/http.py
```diff
@@ -81,13 +120,21 @@ class HttpStream(object):
...
 
         self.status_code = 0
         self.headers = None
 
+        retry = OSBSRetry(
+            total=HTTP_MAX_RETRIES,
+            backoff_factor=HTTP_BACKOFF_FACTOR
+        )
+        self.session = requests.Session()
+        self.session.mount('http://', HTTPAdapter(max_retries=retry))
```
lcarva: Maybe figure out which one (`http` vs `https`) is actually needed?

osbs/http.py
```diff
@@ -131,7 +178,12 @@ def __init__(self, url, method, data=None, kerberos_auth=False,
             args['stream'] = True
 
         args['headers'] = headers
-        self.req = requests.request(method, url, **args)
+
+        if retries_disabled:
+            request_method = requests.request
+        else:
+            request_method = self.session.request
```
lcarva: Why not always use the session?

osbs/constants.py
vrutkovs: 408 is probably a good idea. @twaugh ?

osbs/http.py
vrutkovs: Right, but that won't let us log retry status

osbs/http.py
vrutkovs: Yep. RHEL7 uses requests 2.6.0, so we can't use a lot of goodies from recent versions

osbs/http.py
vrutkovs: Hmm, good idea

osbs/http.py
vrutkovs: IIRC at this point some parts are comparing status code in `OsbsNetworkException` - this should go away.

I think in case of retries it makes sense to avoid status code there - it may throw different codes in various attempts

osbs/http.py
lcarva: Have you played with setting the logging level? `logging.getLogger("urllib3").setLevel(logging.WARNING)`

osbs/http.py
vrutkovs: Hmm, maybe a generic 'http' would work for both

osbs/http.py
vrutkovs: Right, sessions gives us a connection pool, so its a good idea to rewrite this to:
```
if not retries_disabled:
    self.session.mount('http://', HTTPAdapter(max_retries=retry))
```
and use `self.session` always

osbs/http.py
vrutkovs: Generic `.mount` doesn't seem to work

osbs/http.py
vrutkovs: Implemented

osbs/http.py
vrutkovs: That doesn't seem to do anything. Looking at the code of retry.py in rhel7's urllib3 it doesn't log anything

vrutkovs:
It doesn't look like any existing functionality should be using this new contextmanager yet

osbs/api.py
```diff
@@ -1024,3 +1025,15 @@ def delete_config_map(self, name):
         """
         response = self.os.delete_config_map(name)
         return response
+
+    @osbsapi
```
twaugh: Not sure, but I think we don't want `@osbsapi` here.

osbs/api.py
```diff
@@ -1024,3 +1025,15 @@ def delete_config_map(self, name):
...
+    @contextmanager
+    def retries_disabled(self):
+        """
+        Context manager to disable retries on requests
+        :returns: OSBS object
+        """
+        existing_os = self.os
+        self.os.disable_retries()
+        yield self
+        self.os = existing_os
```
twaugh: What about enabling retries again?

tests/test_api.py
```diff
@@ -1958,3 +1958,12 @@ def test_config_map(self, osbs): # noqa
...
         assert config_map is None
+
+    def test_retries_disabled(self, osbs): # noqa
+        (flexmock(osbs.os._con)
+            .should_call('get')
+            .with_args("/oapi/v1/namespaces/default/builds/", headers={},
+                       verify_ssl=True, retries_disabled=True))
+        with osbs.retries_disabled() as tryonce:
+            response_list = tryonce.list_builds()
+            assert response_list is not None
```
twaugh: Verify retries are no longer disabled after the `with`.

osbs/constants.py
twaugh: Yes, let's add 408.

For 409, we need to re-fetch the object being modified so it should *not* be in the forcelist.

osbs/http.py
twaugh: If we must log (must we?) then let's at least accept any `*args, **kwargs` provided, and pass them through to the superclass's implementation.

But I don't think we necessarily need to log.

osbs/http.py
twaugh: What even calls this method?

osbs/api.py
vrutkovs: This is what last line should do. Gonna add tests for this

osbs/http.py
vrutkovs: This happens inside urrlib3: whenever the request is finished it is evaluating if retry is required: https://github.com/requests/requests/blob/v2.6.0/requests/packages/urllib3/connectionpool.py#L640

osbs/http.py
vrutkovs: I think logging would be very handy in the future, although its not required

tests/test_api.py
vrutkovs: Fixed

osbs/http.py
twaugh: Not in requests 2.13.0.

osbs/http.py
twaugh: If we can't do it in a sane way, let's not log.

osbs/core.py
```diff
@@ -93,6 +92,7 @@ def __init__(self, openshift_api_url, openshift_api_version, openshift_oauth_url
         self.verbose = verbose
         self.verify_ssl = verify_ssl
         self._con = HttpSession(verbose=self.verbose)
+        self.retries_disabled = False
```
twaugh: It's often better to name things the other way around, e.g. `retries_enabled`, otherwise when reading code you end up trying to work out "double negatives" (e.g. "not disabled").

vrutkovs: Agree, renamed everything (except contextmanager) to `retries_enabled` as it makes the code a bit more readable

vrutkovs:
Okay, will remove it and fix tests

vrutkovs:
Tests were removed and adjusted to pass with older python-requests

osbs/http.py
```diff
@@ -81,13 +97,22 @@ class HttpStream(object):
...
                  headers=None, stream=False, username=None, password=None,
-                 client_cert=None, client_key=None, verbose=False):
+                 client_cert=None, client_key=None, verbose=False, retries_enabled=True):
         self.finished = False  # have we read all data?
         self.closed = False    # have we destroyed curl resources?
 
         self.status_code = 0
         self.headers = None
 
+        retry = OSBSRetry(
```
twaugh: This could just be `Retry` now?

vrutkovs: We still have to reimplement `is_forced_retry` with a custom list of statuses

twaugh: We can pass a custom list of statuses in using `Retry(..., status_forcelist=...)` can't we?

osbs/http.py
vrutkovs: Oh, right, it was renamed - https://github.com/shazow/urllib3/blob/master/urllib3/util/retry.py#L294

vrutkovs: Implemented, this now would replace both methods

osbs/http.py
twaugh: What if `is_retry` gets renamed again?

We need to use the documented interface, and that says to pass `status_forcelist` into the constructor.

osbs/api.py
```diff
@@ -1024,3 +1025,13 @@ def delete_config_map(self, name):
...
         return response
+
+    @contextmanager
+    def retries_disabled(self):
+        """
+        Context manager to disable retries on requests
+        :returns: OSBS object
+        """
+        self.os.retries_enabled = False
+        yield self
```
lcarva: No need to actually yield a value btw. Just `yield` will do the trick. Thinking from a usage point of view, you probably want this:

```python
def my_only_once_method(self):
  with self.retries_disabled():
    self.os.something(...)
```

and not this:

```python
def my_only_once_method(self):
  with self.retries_disabled() as disabled:
    disabled.os.something(...)
```

osbs/http.py
```diff
@@ -34,6 +39,21 @@
 logger = logging.getLogger(__name__)
 
 
+class OSBSRetry(Retry):
```
lcarva: I thought the whole point of subclassing `Retry` was so we could hook some logging. This is not being done anymore.
Why not just use `status_forcelist` ?

osbs/api.py
vrutkovs: That makes sense, it also makes the code look cleaner. Since `Openshift` object is a singleton there, there is no point in making an additional object to be used in contextmanager.

@twaugh thoughts?

twaugh: I'm fine with not yielding a value.

vrutkovs: Reworked the contextmanager

PR #640, opened 2017-07-21
Consult git repo when enabling autorebuilds

lcarva:
If configuration file in git repo has been set to enable
autorebuilds, Dockerfile must also not have "release" label set.
If so, build now fails to start to avoid confusion.

Additionally, if configuration has been set to disable autorebuild,
image change triggers are automatically removed from BuildConfig.
This avoids unwanted auto builds.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -57,6 +57,7 @@ class BuildRequest(object):
         self._dj = None
         self._resource_limits = None
         self._openshift_required_version = parse_version('1.0.6')
+        self._repo_info = None
         # For the koji "scratch" build type
         self.scratch = None
         self.base_image = None
@@ -122,6 +123,9 @@ class BuildRequest(object):
     def set_openshift_required_version(self, openshift_required_version):
         if openshift_required_version is not None:
             self._openshift_required_version = openshift_required_version
+
+    def set_repo_info(self, repo_info):
+        self._repo_info = repo_info
 
     @property
     def build_id(self):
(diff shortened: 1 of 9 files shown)


osbs/constants.py
```diff
@@ -71,3 +71,5 @@
 # Logging format used in Atomic Reactor
 ATOMIC_REACTOR_LOGGING_FMT = \
     '%(asctime)s platform:%(arch)s - %(name)s - %(levelname)s - %(message)s'
+
+REPO_CONFIG_FILE = '.osbs-repo-config'
```
vrutkovs: (not related to this PR) Do we want to bring all the repo configs in one file eventually?

osbs/build/build_request.py
```diff
@@ -559,6 +563,19 @@ def adjust_for_custom_base_image(self):
...
+        if not self._repo_info:
+            logger.warning('repo info not set')
+            return
+
+        if not self._repo_info.configuration.is_autorebuild_enabled():
+            logger.info('autorebuild is disabled in repo configuration, removing triggers')
+            self.template['spec'].pop('triggers', None)
+
+        elif 'release' in self._repo_info.dockerfile_parser.labels:
+            raise RuntimeError('when autorebuild is enabled in repo configuration, '
```
rcerven: why is this raising error? shouldn't bump_release take care about bumping_release? why it can't be in Dockerfile?

vrutkovs: It would do that just once, without writing the changes back to Dockerfile. As a result, next autorebuild of this would cause exception in `koji_promote`/`koji_tag_build` - as the build with the bumped release already exists.

osbs/constants.py
lcarva: I believe that's the idea of behind using `container.yaml`.

lcarva:
Updated release notes.

osbs/build/build_request.py
csomh: Should the deprecated 'Release' label be also considered here?

lcarva: I didn't think it was needed. Perhaps that was an incorrect assumption. @twaugh, thoughts?

lcarva: Probably safer to check aliases as well. Created #644 .

twaugh: Agreed.

PR #641, opened 2017-07-25
[WIP] attempt to build registries for pulp_pull plugin

jarodwilson:
We need to provide secret to the pulp_pull plugin, so that it can properly query registries to look for v2 schema 2 manifests.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -859,6 +859,13 @@ class BuildRequest(object):
         if not pulp_registry:
             logger.info("removing %s from request, requires pulp_registry", pulp_registry)
             self.dj.remove_plugin(phase, plugin)
+            return
+
+        # Verify that we have a secret
+        if self.spec.pulp_secret.value is None:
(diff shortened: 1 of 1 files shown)


osbs/build/build_request.py
```diff
@@ -859,6 +859,13 @@ def render_pulp_pull(self):
         if not pulp_registry:
             logger.info("removing %s from request, requires pulp_registry", pulp_registry)
             self.dj.remove_plugin(phase, plugin)
+            return
+
+        # Verify that we have a secret
+        if self.spec.pulp_secret.value is None:
+            raise OsbsValidationException("Pulp registry specified with no secret")
```
vrutkovs: Secrets are not required for unsecured registries, which we still want to support. Maybe a simple warning should be sufficient?

twaugh:
No, `pulp_secret` is the secret holding the certificate for using the Pulp API, not for using the Docker API on Crane.

I think we can just drop this PR.

jarodwilson:
Concur with @twaugh, let's just drop this.

PR #645, opened 2017-07-26
Rework travis to make it run in different Fedora/CentOS containers

vrutkovs:
Benefits of this approach:
 * Travis failures can be reproduced locally: `sudo env OS=centos OS_VERSION=7 ./test.sh` or `sudo env OS=fedora OS_VERSION=26 PYTHON_VERSION=3 ./test.sh`. By default it would run tests on CentOS 6
 * Tests would run against actual packages in distributions, not latest from pypi
 * Spec files dependencies are tested as well, so less surprises for Fedora packagers during `%check` phase

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,36 @@
 language: python
-python:
-  - "2.6"
-  - "2.7"
-  - "3.4"
-  - "3.5"
-# command to install dependencies
-install:
-  - "pip install -U pip"
-  - "pip install -r requirements.txt"
-  - "pip install -r tests/requirements.txt"
-  - "pip install pytest-cov coveralls"
-  - "if [[ ${TRAVIS_PYTHON_VERSION%.*} -gt 2 ]]; then pip install -r requirements-py3.txt; fi"
-# command to run tests
-script: py.test -vv tests --cov osbs
-# run in a docker container
-sudo: false
+sudo: "required"
+services:
+  - docker
+env:
+  matrix:
+    - OS=centos
+      OS_VERSION=6
+      PYTHON_VERSION=2
+    - OS=centos
+      OS_VERSION=7
+      PYTHON_VERSION=2
+    - OS=fedora
+      OS_VERSION=25
+      PYTHON_VERSION=2
+    - OS=fedora
+      OS_VERSION=25
+      PYTHON_VERSION=3
+    - OS=fedora
+      OS_VERSION=26
+      PYTHON_VERSION=2
+    - OS=fedora
+      OS_VERSION=26
+      PYTHON_VERSION=3
+    - OS=fedora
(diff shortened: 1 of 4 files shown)


twaugh:
@csomh, what do you think of this approach compared with tox? I can definitely see the advantages.

tests/test_core.py
```diff
@@ -162,10 +162,13 @@ def test_stream_logs_not_decoded(self, caplog):
 
         fake_response = flexmock(status_code=httplib.OK, headers={})
 
+        kwargs = {'decode_unicode': False}
+        if requests.__version__.startswith('2.6.'):
+            kwargs['chunk_size'] = 1
```
twaugh: Is this a test fix that's needed regardless?

twaugh:
Is there any way of preserving python-2.6 testing? The CentOS 6 run used 2.7.

csomh:
@twaugh I definitely see the benefits of this. Compared to tox the only difference I see is that with tox the matrix would be defined in the tox config, so one running the tests locally would not need to think about the matrix.

With this approach only travis knows about the matrix. Could the test script be updated to consider the matrix from the travis config? This would result in the same behavior as tox - although I am not really sure how important this would be (and I don't really have preferences regarding this).

vrutkovs:
>With this approach only travis knows about the matrix. Could the test script be updated to consider the matrix from the travis config?

@csomh, Travis matrix could be moved in the script instead, yes, but I'm concerned about the time it would take to run all the configs (its now 8 configurations and it doesn't include Fedora 27, for instance)

vrutkovs:
PR was rebased, test fix was removed - the test is now removed in master

twaugh:
There's still no python-2.6 testing.

vrutkovs:
It does test python2.6 in https://travis-ci.org/projectatomic/osbs-client/jobs/273221732

>Using /usr/lib64/python2.6/site-packages

and 

>platform linux2 -- Python 2.6.6, pytest-3.2.2, py-1.4.34, pluggy-0.4.0 -- /usr/bin/python

vrutkovs:
More fixes to the code are required:
* there is no `is_retry` in `python[2,3]-requests-2.10.0-4.fc25.noarch`
* Mysterious `AttributeError: executable` in Fedora 26 Py3
* Missing RPM keys in Rawhide image - Dockerfile should be updated

twaugh:
Ah! Now I see it. :)

lcarva:
@vrutkovs, on the python2.6 job, it looks like it does a bunch of unnecessary pre-processing for python 2.7. It'd be nice to avoid that.

vrutkovs:
@lcarva that's because default python in Travis is 2.7 and we need it for coveralls reporting. It doesn't affect the container builds though

PR #650, opened 2017-07-28
Development setup improvements

owtaylor:
Here are some improvements to development-setup.md based on what I had to do to get prod and orchestrated builds working locally.

Perhaps it's better to just point people to osbs-box than to have development-setup.md, but as long as there is a development-setup.md, hopefully these changes are useful!

### DIFF
--- a/docs/development-setup.md
+++ b/docs/development-setup.md
@@ -203,3 +203,74 @@ container with credentials to push to the registry:
 ```
 $ osbs --config osbs.conf --instance local build -g https://github.com/TomasTomecek/hello-world-container -b master -u ${USER} -c hello-world
 ```
+
+## Configuring orchestration
+
+To test out orchestrated builds, you'll need another namespace to act as a
+test worker, and add the necessary permissions to it:
+
+```
+$ oc new-project worker01
+
+$ oc login -u system:admin
+$ oc policy -n worker01 add-role-to-user edit system:serviceaccount:myproject:builder
+$ oc policy -n worker01 add-role-to-user edit system:serviceaccount:worker01:builder
+$ oc policy -n worker01 add-role-to-group system:build-strategy-custom system:authenticated
+
+$ oc login -u developer
+$ oc project myproject
+```
+
+Then you'll need to create two configuration files. Create a file `reactor-conf/config.yaml`
+with the contents
+
+```
+version: 1
+clusters:
+  x86_64:
+  - name: worker01
+    max_concurrent_builds: 4
+    enabled: true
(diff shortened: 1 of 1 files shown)


docs/development-setup.md
```diff
@@ -17,8 +17,8 @@ We'll use the `oc cluster up` method, since it's very easy to use:
     ```
     $ oc cluster up
     ```
-    (If you are running into some connection issues, please flush your iptables
-    rules bufore starting the cluster with `iptables -F`)
```
lcarva: The linked documentation does not mention `iptables -F`. This works for me whenever a build fails to lookup github.com, for example. I don't think we should remove that. Maybe:
> flush your iptables rules before starting the cluster with `iptables -F`, for further assistance see...

docs/development-setup.md
```diff
@@ -192,3 +200,70 @@ container with credentials to push to the registry:
...
+use_kerberos = false
+verify_ssl = false
+use_auth = true
+registry_uri = <registry URL>
+```
+
+And create corresponding secrets:
+
+```
+$ oc secrets new osbs-client-conf osbs-client-con/osbs.conf
```
lcarva: At this point, it's probably safer to make these commands use: `oc -n myproject`
This will make these instructions more resilient as we may come back to it for reference and not execute command in exact order.

docs/development-setup.md
owtaylor: Clearly, 'oc cluster' doesn't work very well with a laptop with a dynamic IP and Fedora's default firewall rules. (Origin 3.6 has some patches in it that are supposed to improve the dynamic IP situation.) I felt that things went a bit better when I followed the instructions in that document, and stopped removing all the iptables rules, restarting docker, restarting firewalld, etc. I'll do a bit more experimentation and see if I can figure out some wording that suggests 'iptables -F' as another thing to try.

owtaylor: I did some more testing and with the firewall rules added, things seemed to be robust when I did 'oc cluster down', 'oc cluster up', changed IP, rebooted, etc. No firewall problems. I'm hestitant to recommend iptables -F because I was having a painful time when I was trying to use it at the recommendation of this document - one thing would start working, something else would break, I'd need to restart docker, etc. I think there are rules in the FORWARD chain that are actually needed for correct operation. But I've added some wording suggesting iptables -F as a last resort. See what you think.

docs/development-setup.md
owtaylor: Fixed.

PR #651, opened 2017-07-28
plugins: Move delete_from_registry from worker to orchestrator

dzickusrh:
This needs the corresponding atomic_reactor change first.

For arrangement version 3, the delete_from_registry will move from the worker
node to the orchestrator node.  This is to allow the orchestrator to do some
post-processing of the registries before they are deleted.

In order to make this work successfully, the render_pulp_sync needed to split
out the delete_from_registry pieces into its own render_delete_from_registry().

This would allow the rendering to happen regardless of which node the plugin
actually ran in.  Further some tweaks were added to the render function based
on which arrangement version was being used.  This is for backwards compatibility.

Finally the worker_inner json file for inputs and tests were updated to remove
the delete_from_registry plugin on the worker node in arrangement version 3 as
everything should be working correctly now.

### DIFF
--- a/inputs/worker_inner:3.json
+++ b/inputs/worker_inner:3.json
@@ -136,12 +136,6 @@
   ],
   "exit_plugins": [
     {
-      "name": "delete_from_registry",
-      "args": {
-        "registries": {}
-     }
-    },
-    {
       "name": "store_metadata_in_osv3",
       "args": {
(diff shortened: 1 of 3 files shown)


dzickusrh:
Replaced with PR 664.  Dropping this request.

PR #652, opened 2017-07-31
Remove pulp_pull from scratch build request or when kojhub is not specified

vrutkovs:
pulp_pull is required to set koji metadata only

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -548,7 +548,8 @@ class BuildRequest(object):
         if self.scratch:
             remove_plugins = [
                 ("prebuild_plugins", "koji_parent"),
-                ("postbuild_plugins", "compress"),  # only for Koji
+                ("postbuild_plugins", "compress"),  # required only to make an archive for Koji
+                ("postbuild_plugins", "pulp_pull"),  # required only to make an archive for Koji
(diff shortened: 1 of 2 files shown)


twaugh:
Just a note: one other thing it also does is ensure that the image is available to pull. That is important to any images further down the layer chain... however, I believe pull_base_image ought to retry in any case, so I'm happy for pulp_pull to be dependent on Koji integration as in this PR.

vrutkovs:
>Just a note: one other thing it also does is ensure that the image is available to pull.

`pulp_pull` would verify that image is available to pull in a limited number of cases:
1) only right after it was pushes. It can be remove/moved by Pulp admins generally
2) it would do the actual pull only on schema1-only pulp now. In case of newer pulps atomic-reactor would check the manifest only.

As Tim has pointed out I think it would be safe to merge after atomic-reactor would get retry capabilities

lcarva:
This seems reasonable. But I can see how someone in the future might be confused about this change.
It would be nice to add a comment in the code, and/or commit message, indicating that `pulp_pull` only provides values that are used when `koji` integration is enabled.

vrutkovs:
Added a comment, but words fail me, please verify it makes sense. Also, commit message has a note about it

lcarva:
Updated release notes.

PR #653, opened 2017-07-31
move required, unwanted logic to api

rcerven:


tests/test_api.py
```diff
@@ -1880,7 +1894,7 @@ def __init__(self, platform, release, platforms, arrangement_version,
                     .and_raise(CustomTestException))
 
         if raises_exception:
-            with pytest.raises(ValueError):
+            with pytest.raises(OsbsException):
```
vrutkovs: The test should verify that underlying exception is `ValueError`

tests/test_api.py
```diff
@@ -1812,10 +1823,10 @@ def test_reactor_config_secret(self):
         ("plat", 'rel', None, True, False, 1, False),
         ("plat", 'rel', None, True, False, None, True),
         # orchestrator build
-        (None, None, 'platforms', False, True, 1, True),
+        (None, None, 'platforms', False, True, 1, False),
```
vrutkovs: Why wouldn't it raise exception here?

tests/test_api.py
```diff
@@ -1812,10 +1823,10 @@ def test_reactor_config_secret(self):
...
         (None, None, 'platforms', False, True, None, False),
         # prod build
-        (None, None, None, False, False, 1, True),
+        (None, None, None, False, False, 1, False),
```
vrutkovs: Why wouldn't it raise exception here?

tests/test_api.py
rcerven: arrangement_version is optional param for orchestrator, at least according docstring

- arrangement_version param may be used to select which                       
  orchestrator_inner:n.json template to use

tests/test_api.py
rcerven: added check

tests/fake_api.py
```diff
@@ -364,6 +368,9 @@ def osbs(openshift, kwargs=None, additional_config=None):
             koji_root = http://koji.example.com/kojiroot
             koji_hub = http://koji.example.com/kojihub
             use_auth = false
+            git_branch = {git_branch}
```
lcarva: No, let's not set git information in config file.

tests/test_api.py
```diff
@@ -1212,6 +1212,9 @@ def test_create_build_config_label_mismatch(self):
             has_ist_trigger=lambda: False,
             scratch=False)
 
+        spec = BuildSpec()
```
lcarva: Why is this required now?

tests/test_api.py
```diff
@@ -1840,18 +1851,21 @@ def __init__(self, platform, release, platforms, arrangement_version,
...
             'target': None,
             'architecture': None,
             'yum_repourls': None,
         }
-        if worker:
+        if arrangement_version:
             expected_kwargs.update({
                 'arrangement_version': arrangement_version,
+            })
+        if koji_upload_dir:
```
lcarva: ??

tests/test_api.py
lcarva: I'm trying to understand what has changed that would change this behavior.

lcarva: As @rcerven pointed out on IRC, this should not fail now because the CLI check incorrectly prohibited use of `arrangement_version` for orchestrator builds.

osbs/api.py
```diff
@@ -652,6 +655,13 @@ def create_orchestrator_build(self, **kwargs):
         if not self.can_orchestrate():
             raise OsbsValidationException("can't create orchestrate build "
                                           "when can_orchestrate isn't enabled")
+        extra = set()
```
vrutkovs: `extra = [x for x in ('platform', 'release') if kwargs.get(x)]`?

tests/test_api.py
vrutkovs: Oh, I see, agreed

tests/test_api.py
vrutkovs: Same here - agreed

tests/test_api.py
```diff
@@ -1345,7 +1345,7 @@ def test_create_build_config_create(self):
             .should_receive('create_build_config')
             .with_args(json.dumps(build_json))
             .once()
-            .and_return(flexmock(json=lambda: {'spam': 'maps'})))
+            .and_return(flexmock(json=lambda: {'spam': 'maps', 'spec': {}})))
```
vrutkovs: Not sure why this change is required - nothing in current PR changes the output of `create_build_config`

tests/test_api.py
```diff
@@ -1885,9 +1893,10 @@ def __init__(self, platform, release, platforms, arrangement_version,
                     .and_raise(CustomTestException))
 
         if raises_exception:
-            with pytest.raises(ValueError):
+            with pytest.raises(OsbsException):
                 cmd_build(MockArgs(platform, release, platforms, arrangement_version,
                           worker, orchestrator), osbs)
+                assert 'OsbsException: ValueError(' in caplog.text()
```
vrutkovs: A better check would be:
```
with pytest.raises(OsbsException) as exc_info:
...
    assert isinstance(exc_info.cause, ValueError)
```

tests/test_api.py
rcerven: it doesn't anymore, will remove

tests/test_api.py
rcerven: sure

osbs/api.py
rcerven: sure

tests/test_api.py
rcerven: on line 1829: koji_upload_dir = 'upload' if worker else None
koji_upload_dir is defined if worker

condition before was if worker, but that changed, so in order not to pass None koji_upload_dir, there is new condition for it

rcerven: https://github.com/projectatomic/osbs-client/blob/master/osbs/cli/main.py#L384

koji_upload_dir is passed ONLY when is defined

vrutkovs: It doesn't seem to be related to this PR, a separate commit/PR would be better

lcarva: @rcerven, that makes sense. Otherwise, comparison of `expected_kwargs`, would fail because the `koji_upload_dir` key is set.

lcarva:
Updated release note.

PR #654, opened 2017-07-31
conf: add support for "platform:..." sections

mlangsdorf:
add support to Configuration to parse platform: sections, as described at
http://osbs.readthedocs.io/en/latest/multiarch.html#platform-description.

change get_registry_api_versions to take a platform argument. If there
is a platform argument, only return 'v1' if the configured api version
supports it and that platform value has enable_v1 set to True. Otherwise,
return v2. If the platform doesn't enable v1 and the registry is v1 only,
raise an exception.

Also raise an exception if two platforms have enable_v1 set to true.

A stub instance of render_group_manifests() is included, but since
group_manifests isn't in any of the existing arrangements, it will do
nothing for now in practice.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -469,7 +469,7 @@ class OSBS(object):
             registry_uris=self.build_conf.get_registry_uris(),
             registry_secrets=self.build_conf.get_registry_secrets(),
             source_registry_uri=self.build_conf.get_source_registry_uri(),
-            registry_api_versions=self.build_conf.get_registry_api_versions(),
+            registry_api_versions=self.build_conf.get_registry_api_versions(platform),
             openshift_uri=self.os_conf.get_openshift_base_uri(),
             builder_openshift_url=self.os_conf.get_builder_openshift_url(),
             kojiroot=self.build_conf.get_kojiroot(),
@@ -515,6 +515,7 @@ class OSBS(object):
             platform_node_selector=self.build_conf.get_platform_node_selector(platform),
             filesystem_koji_task_id=filesystem_koji_task_id,
             koji_upload_dir=koji_upload_dir,
+            platform_descriptors=self.build_conf.get_platform_descriptors(),
         )
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
         build_request.set_repo_info(repo_info)
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -64,6 +64,7 @@ class BuildRequest(object):
         self.low_priority_node_selector = None
         # forward reference
         self.platform_node_selector = None
+        self.platform_descriptors = None
 
     def set_params(self, **kwargs):
         """
@@ -91,6 +92,7 @@ class BuildRequest(object):
         :param use_auth: bool, use auth from atomic-reactor?
         :param low_priority_node_selector: dict, a nodeselector for builds with lower priority
         :param platform_node_selector: dict, a nodeselector for a specific platform
+        :param platform_descriptors: dict, platforms and their archiectures and enable_v1 settings
         """
 
         # Here we cater to the koji "scratch" build type, this will disable
@@ -103,6 +105,7 @@ class BuildRequest(object):
         self.base_image = kwargs.get('base_image')
         self.low_priority_node_selector = kwargs.get('low_priority_node_selector')
         self.platform_node_selector = kwargs.get('platform_node_selector', {})
+        self.platform_descriptors = kwargs.get('platform_descriptors', {})
 
         logger.debug("setting params '%s' for %s", kwargs, self.spec)
         self.spec.set_params(**kwargs)
@@ -962,6 +965,42 @@ class BuildRequest(object):
             self.dj.remove_plugin("postbuild_plugins", "pulp_sync")
             self.dj.remove_plugin("exit_plugins", "delete_from_registry")
 
+    def render_group_manifests(self):
+        """
+        Configure the group_manifests plugin. Group is always set to false for now.
+        """
+        if not self.dj.dock_json_has_plugin_conf('postbuild_plugins',
+                                                 'group_manifests'):
+            return
+
+        pulp_registry = self.spec.pulp_registry.value
+        if pulp_registry:
+            self.dj.dock_json_set_arg('postbuild_plugins', 'group_manifests',
+                                      'pulp_registry_name', pulp_registry)
+
+            # Verify we have either a secret or username/password
+            if self.spec.pulp_secret.value is None:
+                conf = self.dj.dock_json_get_plugin_conf('postbuild_plugins',
+                                                         'group_manifests')
+                args = conf.get('args', {})
+                if 'username' not in args:
+                    raise OsbsValidationException("Pulp registry specified "
+                                                  "but no auth config")
+
+            self.dj.dock_json_set_arg('postbuild_plugins', 'group_manifests',
+                                      'group', False)
+            goarch = {}
+            for platform in self.platform_descriptors:
+                goarch[platform] = self.platform_descriptors[platform]['architecture']
+            self.dj.dock_json_set_arg('postbuild_plugins', 'group_manifests',
+                                      'goarch', goarch)
+
+        else:
+            # If no pulp registry is specified, don't run the pulp plugin
+            logger.info("removing group_manifests from request, "
+                        "requires pulp_registry")
+            self.dj.remove_plugin("postbuild_plugins", "group_manifests")
+
     def render_import_image(self, use_auth=None):
         """
         Configure the import_image plugin
@@ -1226,6 +1265,7 @@ class BuildRequest(object):
         self.render_pulp_pull()
         self.render_pulp_push()
         self.render_pulp_sync()
+        self.render_group_manifests()
         self.render_koji_promote(use_auth=use_auth)
         self.render_koji_upload(use_auth=use_auth)
         self.render_koji_import(use_auth=use_auth)
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -266,12 +266,22 @@ class Configuration(object):
         else:
             return []
 
-    def get_registry_api_versions(self):
+    def get_registry_api_versions(self, platform=None):
         value = self._get_value("registry_api_versions",
                                 self.conf_section,
                                 "registry_api_versions",
                                 default='v1,v2')
-        return value.split(',')
+        versions = value.split(',')
+        section = 'platform:{0}'.format(platform)
+        enable_v1 = self._get_value("enable_v1", section, "enable_v1",
+                                    default=False, is_bool_val=True)
+        if enable_v1:
+            return versions
+        else:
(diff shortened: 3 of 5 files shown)


osbs/conf.py
```diff
@@ -266,12 +267,22 @@ def get_registry_secrets(self):
...
                                 self.conf_section,
                                 "registry_api_versions",
                                 default='v1,v2')
-        return value.split(',')
+        if not platform:
+            return value.split(',')
+        else:
+            self.get_platform_descriptors()
+            if platform in self.platform_descriptors and self.platform_descriptors[platform]['enable_v1']:  # noqa
+                return value.split(',')
```
lcarva: Shouldn't there be an error if value is just `v2`?

osbs/conf.py
```diff
@@ -554,3 +565,22 @@ def get_platform_node_selector(self, platform):
                 nodeselector = dict([k.strip(), v.strip()] for (k, v) in raw_nodeselector.items())
 
         return nodeselector
+
+    def get_platform_descriptors(self):
```
lcarva: This method is doing more than `get`. Let's name it something like: `resolve_platform_descriptors`
Alternatively, I think it's fine to not store the result in `self.platform_descriptors` and simply resolve and return the whole object, leaving it named as `get_platform_descriptors`.

osbs/api.py
```diff
@@ -515,6 +515,7 @@ def _do_create_prod_build(self, git_uri, git_ref,
             platform_node_selector=self.build_conf.get_platform_node_selector(platform),
             filesystem_koji_task_id=filesystem_koji_task_id,
             koji_upload_dir=koji_upload_dir,
+            platform_descriptors=self.build_conf.get_platform_descriptors,
```
lcarva: `self.build_conf.get_platform_descriptors` is a reference to the method, did you mean `self.build_conf.get_platform_descriptors()`

osbs/conf.py
```diff
@@ -266,12 +267,22 @@ def get_registry_secrets(self):
...
+    def get_registry_api_versions(self, platform=None):
         value = self._get_value("registry_api_versions",
                                 self.conf_section,
                                 "registry_api_versions",
                                 default='v1,v2')
-        return value.split(',')
+        if not platform:
+            return value.split(',')
+        else:
+            self.get_platform_descriptors()
```
lcarva: How about just looking up the section directly?
```python
section = 'platform:{0}'.format(platform)
enable_v1 = self._get_value("enable_v1", section, "enable_v1", False, True)
```

`_get_value` should handle if either the config section or option is not defined.

osbs/conf.py
mlangsdorf: I don't think so. enable_v1 is "support for the Docker Registry HTTP API v1 may be included for this platform." v1 support is not mandatory.

osbs/conf.py
mlangsdorf: I think I was being overly prepared. I'll remove self.get_platform_descriptors since it isn't used anywhere anyway.

osbs/api.py

osbs/conf.py
mlangsdorf: Fixed.

osbs/conf.py
```diff
@@ -266,12 +267,22 @@ def get_registry_secrets(self):
...
+            if platform in self.platform_descriptors and self.platform_descriptors[platform]['enable_v1']:  # noqa
```
mlangsdorf: Fixed.

osbs/conf.py
lcarva: Hmm, maybe a log message?

osbs/conf.py
```diff
@@ -554,3 +565,23 @@ def get_platform_node_selector(self, platform):
...
+        for section in self.scp.sections():
+            if "platform:" not in section:
+                continue
+            platform = section.split("platform:")[1]
+            platform_descriptor = {}
+            arch = self._get_value("architecture", section, "architecture") or platform
+            enable_v1 = self._get_value("enable_v1", section, "enable_v1", False, True)
+            if enable_v1:
+                if has_v1:
+                    raise OsbsValidationException('two platforms have enable_v1 as true')
```
lcarva: This method is pretty straight forward. A minor enhancement would be to report the multiple platforms that have requested v1. How about turning `has_v1` into a list, and add all the platforms that have requested v1 to it. Then after the for-loop executes, you can simply check the length, if it's greater than 1, an error can be raised indicating exactly which platforms are involved. This also has the benefit of checking if there's possibly a third platform that need to be dealt with.

mlangsdorf: Fixed.

osbs/conf.py
mlangsdorf: I don't think that's necessary. Returning v2 only, even if the platform has enable_v1, is the desired behavior.

lcarva: agree

osbs/build/build_request.py
```diff
@@ -91,6 +92,7 @@ def set_params(self, **kwargs):
         :param use_auth: bool, use auth from atomic-reactor?
         :param low_priority_node_selector: dict, a nodeselector for builds with lower priority
         :param platform_node_selector: dict, a nodeselector for a specific platform
+        :param platform_descriptors: dict, a nodeselector for a specific platform
```
twaugh: This description needs updating.

osbs/build/build_request.py
```diff
@@ -962,6 +965,42 @@ def render_pulp_sync(self):
             self.dj.remove_plugin("postbuild_plugins", "pulp_sync")
             self.dj.remove_plugin("exit_plugins", "delete_from_registry")
 
+    def render_group_manifests(self):
+        """
+        If a pulp registry is specified, use the pulp plugin
```
twaugh: This description needs updating.

osbs/conf.py
```diff
@@ -266,12 +266,23 @@ def get_registry_secrets(self):
...
-    def get_registry_api_versions(self):
+    def get_registry_api_versions(self, platform=None):
         value = self._get_value("registry_api_versions",
                                 self.conf_section,
                                 "registry_api_versions",
                                 default='v1,v2')
-        return value.split(',')
+        if not platform:
+            return value.split(',')
+        else:
```
twaugh: We could lose the `else`: here and de-indent the contained block. If `platform` is falsy, execution will not get here.

osbs/conf.py
```diff
@@ -266,12 +266,23 @@ def get_registry_secrets(self):
...
+            section = 'platform:{0}'.format(platform)
+            enable_v1 = self._get_value("enable_v1", section, "enable_v1", False, True)
```
twaugh: Can we explicitly name the keyword arguments here to save having to look around for the method signature? e.g. `default=False, is_bool_val=True`.

osbs/conf.py
```diff
@@ -266,12 +266,23 @@ def get_registry_secrets(self):
...
+            if enable_v1:
+                return value.split(',')
+            else:
+                if 'v2' in value:
```
twaugh: How about splitting the value right at the start? `versions = value.split(',')`

That would save repeating it, and make this line 'if 'v2' in versions:`, i.e. testing to see whether "v2" is one of the comma-separated values, rather than testing that it's a substring of the unsplit string (which would also match "v20,v21").

osbs/conf.py
```diff
@@ -554,3 +565,25 @@ def get_platform_node_selector(self, platform):
...
+    def get_platform_descriptors(self):
+        has_v1 = []
+        platform_descriptors = {}
+        for section in self.scp.sections():
+            if "platform:" not in section:
+                continue
+            platform = section.split("platform:")[1]
+            platform_descriptor = {}
+            arch = self._get_value("architecture", section, "architecture") or platform
+            enable_v1 = self._get_value("enable_v1", section, "enable_v1", False, True)
```
twaugh: `default=False, is_bool_val=True`

osbs/conf.py
mlangsdorf: Fixed.

osbs/conf.py
mlangsdorf: Fixed.

osbs/conf.py
mlangsdorf: Fixed.

osbs/build/build_request.py
mlangsdorf: Fixed

osbs/conf.py
mlangsdorf: Fixed

osbs/build/build_request.py
```diff
@@ -962,6 +965,42 @@ def render_pulp_sync(self):
...
+                                      'pulp_registry_name', pulp_registry)
+
+            # Verify we have either a secret or username/password
+            if self.spec.pulp_secret.value is None:
+                conf = self.dj.dock_json_get_plugin_conf('postbuild_plugins',
+                                                         'group_manifests')
+                args = conf.get('args', {})
+                if 'username' not in args:
+                    raise OsbsValidationException("Pulp registry specified "
+                                                  "but no auth config")
```
twaugh: This block doesn't get hit during tests.

osbs/build/build_request.py
```diff
@@ -962,6 +965,42 @@ def render_pulp_sync(self):
...
+            for platform in self.platform_descriptors:
+                goarch[platform] = self.platform_descriptors[platform]['architecture']
+            self.dj.dock_json_set_arg('postbuild_plugins', 'group_manifests',
+                                      'goarch', goarch)
+
+        else:
+            # If no pulp registry is specified, don't run the pulp plugin
+            logger.info("removing group_manifests from request, "
+                        "requires pulp_registry")
+            self.dj.remove_plugin("postbuild_plugins", "group_manifests")
```
twaugh: This block doesn't get hit during tests.

mlangsdorf: Fixed.

osbs/build/build_request.py
mlangsdorf: Fixed.

lcarva:
Updated release notes.

osbs/conf.py
```diff
@@ -266,12 +266,22 @@ def get_registry_secrets(self):
...
 
-    def get_registry_api_versions(self):
+    def get_registry_api_versions(self, platform=None):
         value = self._get_value("registry_api_versions",
                                 self.conf_section,
                                 "registry_api_versions",
                                 default='v1,v2')
-        return value.split(',')
+        versions = value.split(',')
+        section = 'platform:{0}'.format(platform)
```
twaugh: If `platform is None` we should not perform this check and instead return `versions` unchanged.

We should only remove v1 support if a platform is specified for which `enable_v1` is false.

This is because the orchestrator build, which is not platform-specific, needs to know the full set of API versions required. Only the worker builds, for which a platform will be specified, will need to know to disable v1 support.

mlangsdorf: Fixed in https://github.com/projectatomic/osbs-client/pull/661

PR #655, opened 2017-07-31
Add new arrangement 4

lcarva:
Signed-off-by: Luiz Carvalho <lucarval@redhat.com>

PR #656, opened 2017-08-01
Add Flatpak support

owtaylor:
This is a pretty complete patch to make osbs-client create Flatpaks using projectatomic/atomic-reactor#747. The main caveat is the module_compose_url workaround for not having support for automatically composing repositories for successful module builds. The interaction with Koji also hasn't been tested - while it's supposed to be "just the same" as images, that needs to be validated.


vrutkovs:
This should probably wait for #655 to be merged and update arrangement 4 so that it wouldn't change the arrangement in the released version

owtaylor:
The latest push push adds support for looking up a yum repository for a module build via the "On Demand Compose Server". I left this as separate commits for clarity, though it involves some amount of restructuring of the earlier commits.

This also rebases on top of the addition of arrangement version 4 - the patch here changes *both* the version 3 inputs (for testing, since the version 4 json doesn't work yet), and the version 4 inputs (probably the final landing place of the additions). Will fix up appropriately at the point this is ready to land.

docs/configuration_file.md
```diff
@@ -59,6 +59,22 @@ Some options are also mandatory.
 
 * `koji_kerberos_principal` (*optional*, `string`) - kerberos principal for the keytab provided in `koji_kerberos_keytab`
 
+* `flatpak` (*optional*, `boolean`) - If true, will build a Flatpak as an OCI image rather than a docker image
+
+* `module` (*optional*, `string`) - Module to build the image against, in the form `NAME:STREAM[:VERSION]`. Currently used only for flatpaks.
```
twaugh: Shouldn't this be a per-build parameter rather than something in a configuration file in the build system?

owtaylor: Yeah - while the way that the code works most things can be configured (even target/git_uri, etc.) these don't make much sense as configuration parameters and shouldn't be documented here.

owtaylor:
Latest push removes the excess parameters from configuration_file.md and also adds a commit to add the append parameter to bump_release (see corresponding commit added to the atomic-reactor Flatpak PR.)

owtaylor:
Changes:
 * rebased to latest master
 * remove the changes to arrangement version 3 and only modify arrangement version 4
 * forbid the combination of `--flatpak --isolate` since `--isolated` doesn't make much sense with the autogenerated `<module-build-version>.<n>` releases we use for flatpaks.

osbs/build/build_request.py
```diff
@@ -747,6 +823,12 @@ def render_bump_release(self):
 
         self.dj.dock_json_set_arg(phase, plugin, 'hub', hub)
 
+        # For flatpak, we want to an an name-version-release of
```
twaugh: Typo? Extra 'an'

tests/build_/test_arrangements.py
```diff
@@ -1040,3 +1049,54 @@ def test_inject_parent_image(self, osbs, build_type):
             'koji_hub': osbs.build_conf.get_kojihub()
         }
         assert args == expected_args
+
+    @pytest.mark.parametrize('scratch', [False, True])  # noqa:F811
+    def test_flatpak(self, osbs, scratch):
```
twaugh: Would also be nice to add a test to make sure those plugins are removed for non-flatpak arrangement-version-4 builds.

osbs/api.py
```diff
@@ -451,35 +484,33 @@ def _do_create_prod_build(self, git_uri, git_ref,
...
-                             labels.get_name(label))
+        if flatpak:
+            if module is None:
+                raise ValueError("Flatpak build missing required parameter 'module'")
+            if isolated:
+                # Flatpak builds from a particular stream autogenerate the release
+                # as <module_version>.<n>; it doesn't make sense to make a fix
+                # from specific one of these autogenerated version. What an isolated
+                # fix for module requires will have to be determined from experience.
+                raise ValueError("Flatpak build cannot be isolated")
```
twaugh: Do we want test coverage for either of these two failure paths?

osbs/conf.py
```diff
@@ -194,6 +194,33 @@ def get_namespace(self):
         return self._get_value("namespace", self.conf_section, "namespace",
                                default=DEFAULT_NAMESPACE)
 
+    def get_flatpak(self):
+        return self._get_value("flatpak", self.conf_section, "flatpak",
+                               is_bool_val=True)
```
twaugh: This doesn't seem to get coverage during tests.

osbs/conf.py
```diff
@@ -194,6 +194,33 @@ def get_namespace(self):
...
+
+    def get_module(self):
+        return self._get_value("module", self.conf_section, "module")
```
twaugh: This doesn't seem to get coverage during tests.

twaugh:
There are merge conflicts.

osbs/build/build_request.py
owtaylor: Fixed

tests/build_/test_arrangements.py
owtaylor: Done

osbs/api.py
owtaylor: Add a a test case that exercises the command-line => build args pathj and tests these two cases and also causes the conf.py code to be called.

osbs/conf.py
owtaylor: Forcing it to be covered turned out that it wasn't getting called at all. I made it work because 'flatpak' and 'module' are analogous to 'git_url', 'git_branch' - it is additional specification of *what* to build. And those are parameters are in the config file, though not documented.

owtaylor:
Think all of your comments are now addressed.

lcarva:
Updated release notes draft:
> - Support for building Flatpaks has been added

PR #657, opened 2017-08-02
Adjust max retries and backoff factor to make sure retries are covering an inaccessibility period of 15 mins

vrutkovs:


twaugh:
Did Travis CI get hiccups? Maybe it needs a `git commit --amend --reset-author && git push -f` to poke it.

twaugh:
Am I reading http://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry wrong, or are we still only going to sleep for 96 seconds in total? (3 * (2 ^ (6 - 1))

vrutkovs:
Right, I've used the wrong formula. So amount of time to pause is calculated using:
```
{backoff factor} * (2 ^ ({number of total retries} - 1))
```
`number of total retries` here is current number retries urllib3 has already performed. The final formula would be:
```
4 * (2 ^ (1 - 1)) + 4 * (2 ^ (2 - 1)) + 4 * (2 ^ (3 - 1)) + 4 * (2 ^ (4 - 1)) + 4 * (2 ^ (5 - 1)) + 4 * (2 ^ (6 - 1)) + 4 * (2 ^ (7 - 1)) + 4 * (2 ^ (8 - 1)) == 1020 seconds = 17 mins```

**Comment by vrutkovs** [comment]
Travis tests fail as no output was received in 10 mins, so .. retries are working :)

Updated Travis to wait for 20 mins without output, but probably the tests should be modifying constants

**Comment by twaugh** [comment]
Ah, now I understand, thank you. The current parameters look fine then.

Do we really want every test run to take the full time though? Not sure how to proceed here -- it's valuable to have unit testing for this, but problematic to have it run in real time.

Is there some way we can fake the responses without using httpbin.org?

**Comment by vrutkovs** [comment]
Had to move the tests, which use `new_instances` to a separate module to fight with https://github.com/bkabrda/flexmock/issues/13

File: tests/test_retries.py [hunk 1]
```diff
@@ -0,0 +1,96 @@
...
+import six
+from requests.packages.urllib3.util import Retry
+from osbs.http import HttpSession, HttpStream
+from osbs.exceptions import OsbsNetworkException, OsbsResponseException
+from osbs.constants import HTTP_RETRIES_STATUS_FORCELIST, HTTP_RETRIES_METHODS_WHITELIST
+from osbs.core import Openshift
+
+try:
+    # py2
+    import httplib
```
lcarva: from `six.moves import httplib` ?

vrutkovs: Good idea, implemented

PR #659, opened 2017-08-03
osbs/api: add decode parameter to get_build_logs

jarodwilson:
This adds support for (utf-8 only) log decoding within the
get_build_logs() api method. Ideally, we would clean up the return types
to be more consistent (None or iterable bytes), but for interoperability
with current atomic-reactor, we'll continue to return either None, bytes,
or an iterable of bytes, entirely dependent on what we're fed by the
self.os.logs() call within get_build_logs().

Note: osbs/cli/main.py's calls to get_build_logs have also been updated to
use decode=True, which is a bit of a behavioral change for end-users, but
probably for the better, since the caller mostly just print out whatever
they're handed back, which should look better if fully decoded. The
behavior of any external caller should remain the same.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -17,6 +17,7 @@ import warnings
 import getpass
 from functools import wraps
 from contextlib import contextmanager
+from types import GeneratorType
 
 from osbs.build.build_request import BuildRequest
 from osbs.build.build_response import BuildResponse
@@ -670,22 +671,39 @@ class OSBS(object):
 
             raise
 
+    def _decode_build_logs_generator(self, logs):
+        for line in logs:
+            line = line.decode("utf-8")
+            yield line
+
     @osbsapi
-    def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missing=False):
+    def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missing=False,
+                       decode=False):
         """
         provide logs from build
 
-        NOTE: since there is no way to determine the text encoding in use
-        the bytes are returned exactly as they came from the container.
+        NOTE: Since atomic-reactor 1.6.25, logs are always in UTF-8, so if
+        asked to decode, we assume that is the encoding in use. Otherwise, we
+        return the bytes exactly as they came from the container.
 
         :param build_id: str
         :param follow: bool, fetch logs as they come?
         :param build_json: dict, to save one get-build query
         :param wait_if_missing: bool, if build doesn't exist, wait
+        :param decode: bool, whether or not to decode logs as utf-8
         :return: None, bytes, or iterable of bytes
         """
-        return self.os.logs(build_id, follow=follow, build_json=build_json,
+        logs = self.os.logs(build_id, follow=follow, build_json=build_json,
                             wait_if_missing=wait_if_missing)
+
+        if decode and isinstance(logs, GeneratorType):
+            return self._decode_build_logs_generator(logs)
+
+        # str or None returned from self.os.logs()
+        if decode and logs is not None:
+            logs = logs.decode("utf-8")
+
+        return logs
 
     @osbsapi
     def get_docker_build_logs(self, build_id, decode_logs=True, build_json=None):
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -389,7 +389,7 @@ def cmd_build(args, osbs):
     # we need to wait for kubelet to schedule the build, otherwise it's 500
     build = osbs.wait_for_build_to_get_scheduled(build_id)
     if not args.no_logs:
-        build_logs = osbs.get_build_logs(build_id, follow=True)
+        build_logs = osbs.get_build_logs(build_id, follow=True, decode=True)
         if not isinstance(build_logs, collections.Iterable):
             logger.error("'%s' is not iterable; can't display logs", build_logs)
             return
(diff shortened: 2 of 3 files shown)


osbs/api.py
```diff
@@ -682,10 +683,18 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
         :param follow: bool, fetch logs as they come?
```
lcarva: Probably need to update the `NOTE` comment right above.

osbs/api.py
```diff
@@ -682,10 +683,18 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
-        :return: None, bytes, or iterable of bytes
+        :param decode: bool, whether or not to decode logs as utf-8
+        :return: None, str, bytes, or iterable of bytes
         """
-        return self.os.logs(build_id, follow=follow, build_json=build_json,
+        logs = self.os.logs(build_id, follow=follow, build_json=build_json,
                             wait_if_missing=wait_if_missing)
+        # we can't decode a Generator
+        if decode and follow:
+            warnings.warn("support for decode when using follow has not been implemented")
```
lcarva: How about something like this:
```python
for line in logs:
   if decode:
      line = line.decode('utf-8')
   yield line
```

osbs/api.py
jarodwilson: Done

osbs/api.py
jarodwilson: It was a little bit more involved than this, because of the possibility we're returned a string instead of a generator, but I think I've got it figured out now.

osbs/api.py
```diff
@@ -671,22 +672,37 @@ def create_orchestrator_build(self, **kwargs):
...
         :param wait_if_missing: bool, if build doesn't exist, wait
-        :return: None, bytes, or iterable of bytes
+        :param decode: bool, whether or not to decode logs as utf-8
+        :return: None, iterable of bytes or iterable of strs
         """
-        return self.os.logs(build_id, follow=follow, build_json=build_json,
+        logs = self.os.logs(build_id, follow=follow, build_json=build_json,
                             wait_if_missing=wait_if_missing)
 
+        if isinstance(logs, GeneratorType):
```
lcarva: What are the possible types for `logs`? Seems like one for them is a generator.
I'm surprised that the other type is str/unicode, and not a list of str/unicode.

osbs/api.py
```diff
@@ -671,22 +672,37 @@ def create_orchestrator_build(self, **kwargs):
...
+            for line in logs:
+                if decode:
+                    line = line.decode("utf-8")
+                    logger.debug("line is %s", line)
+                yield line
+        else:
+            if decode:
+                logs = logs.decode("utf-8")
+                logger.debug("logs is %s", logs)
```
lcarva: Let's not log the logs to avoid it being logged twice.

osbs/api.py
lcarva: Never mind I see your comment above now.

osbs/api.py
jarodwilson: Oops, that was a relic of testing, so that I got output from py.test failures, meant to remove it.

osbs/api.py
twaugh: How about this?:
```
if decode:
    # Return a generator that decodes each line
    return (line.decode("utf-8") for line in logs)

return logs
```

jarodwilson: This version results in an AssertionError from test_api.py::TestOSBS::test_build_logs_api, what you end up with for 'content' is just 'l'.

jarodwilson: One nice side-effect I'm seeing from the way I'm doing things is that the data returned stays consistent -- it's always an iterable generator, and the content matches what we're expecting. I've tried a few different variations, and they result in sometimes returning a string instead of an iterable with the string inside it, or the string gets hacked up oddly like I see with the variant you suggested. It's not the cleanest thing in the world, but the added complexity inside get_build_logs makes the test cases cleaner, as well as making follow-on work (get_orchestrator_build_logs) cleaner, since they know they'll always get a generator (or None) back.

twaugh: Fair enough.

osbs/api.py
```diff
@@ -671,22 +672,38 @@ def create_orchestrator_build(self, **kwargs):
...
-        return self.os.logs(build_id, follow=follow, build_json=build_json,
+        logs = self.os.logs(build_id, follow=follow, build_json=build_json,
                             wait_if_missing=wait_if_missing)
 
+        if isinstance(logs, GeneratorType):
+            for line in logs:
+                if decode:
+                    line = line.decode("utf-8")
+                yield line
+        # str or None returned from self.os.logs()
```
lcarva: Hey if `self.os.logs()` returns `None`, the code will throw an exception at `logs.decode("utf-8")`.
Is the comment lying? or do we need `if decode and logs` check?

lcarva:
Let's remove the `[WIP]` marker from PR title and commit message.

osbs/api.py
jarodwilson: yeah, we should probably have a check for None, it is still possible (as far as I know) for self.os.logs() to return None, if it was possible before. I've added something to cover this.

osbs/api.py
```diff
@@ -671,22 +672,41 @@ def create_orchestrator_build(self, **kwargs):
...
-        :return: None, bytes, or iterable of bytes
+        :param decode: bool, whether or not to decode logs as utf-8
+        :return: None, iterable of bytes or iterable of strs
         """
-        return self.os.logs(build_id, follow=follow, build_json=build_json,
+        logs = self.os.logs(build_id, follow=follow, build_json=build_json,
                             wait_if_missing=wait_if_missing)
 
+        if logs is None:
+            yield None
```
lcarva: You probably want `return` here, otherwise it'll continue from this line on the next iteration.

jarodwilson: Good catch. Changed that to a return instead of yield None, had originally tried return None, which at least python 2.x doesn't like.

lcarva:
Updated release notes.

PR #661, opened 2017-08-08
conf: fix get_registry_api_version for orchestrator

mlangsdorf:
orchestrator builds never have platforms, but should always return the
registry_api_versions values from their config file, or ['v1', 'v2'] if
that is undefined.

### DIFF
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -272,6 +272,9 @@ class Configuration(object):
                                 "registry_api_versions",
                                 default='v1,v2')
         versions = value.split(',')
+        if platform is None:
+            return versions
+
         section = 'platform:{0}'.format(platform)
         enable_v1 = self._get_value("enable_v1", section, "enable_v1",
                                     default=False, is_bool_val=True)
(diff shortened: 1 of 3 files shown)


osbs/conf.py
```diff
@@ -272,6 +272,9 @@ def get_registry_api_versions(self, platform=None):
                                 "registry_api_versions",
                                 default='v1,v2')
         versions = value.split(',')
+        if not platform:
```
twaugh: How about `if platform is None`, to do an explicit check against the default value for that parameter?

mlangsdorf: Sure.

mlangsdorf: will registry_api_versions never show up in the orchestrator conf?

twaugh: osbs.conf at the client side, e.g. the Koji builder, will have registry_api_versions, yes. To support v1 it needs to be:

* osbs.conf on Koji builder says registry_api_versions=v1,v2
* client-config-secret says registry_api_versions=v1,v2 *and* platform section has enable_v1=true

lcarva:
@twaugh, @mlangsdorf, so what happens if orchestrator build thinks both v1 and v2 content will be built, but none of the platforms actually enable it. Will this cause an issue for one of the orchestrator plugins?

twaugh:
As long as `pulp_tag` and `pulp_publish` are happy with no worker builds having uploaded v1 images to Pulp, it should be fine.

lcarva:
Not updating releases notes: bugfix to unreleased feature.

PR #662, opened 2017-08-08
Release value

bfontecc007:
Let release value through (at least at this level).

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -708,16 +708,16 @@ def cli():
                               help="set yum proxy to repos from koji/add-yum-repo params")
     build_parser.add_argument('--koji-parent-build', action='store', required=False,
                               help='overwrite parent image with image from koji build')
+    build_parser.add_argument('--release', action='store', required=False,
+                              help='release value to use')
+    build_parser.add_argument('--arrangement-version', action='store', required=False,
+                              help='version of inner template to use')
 
     worker_group = build_parser.add_argument_group(title='arguments for --worker',
                                                    description='Required arguments for creating a '
                                                    'worker build')
     worker_group.add_argument('--platform', action='store', required=False,
                               help='platform name to use')
-    worker_group.add_argument('--release', action='store', required=False,
-                              help='release value to use')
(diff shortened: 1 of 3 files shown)


bfontecc007:
This is all that's needed.

osbs/cli/main.py
```diff
@@ -704,16 +704,16 @@ def cli():
                               help="perform a scratch build")
     build_parser.add_argument("--yum-proxy", action='store', required=False,
                               help="set yum proxy to repos from koji/add-yum-repo params")
+    build_parser.add_argument('--release', action='store', required=False,
+                              help='release value to use')
+    build_parser.add_argument('--arrangement-version', action='store', required=False,
+                              help='version of inner template to use')
```
twaugh: I'm not entirely sure it makes sense to allow `--arrangement-version` for the build parse. We still support 'prod' (all-in-one) builds, which have no concept of a plugin arrangement.

We should at least disallow `--arrangement-version` if neither orchestrator nor worker build is being created.

lcarva:
> How about this part?

That's addressed in this PR.

osbs/cli/main.py
lcarva: The group parser is only used for grouping the help text. Although it looked like it, we actually never enforced that `arrangement-version` was not given to a prod build.

We can either move `arrangement-version` back to where it was. Or add the restriction to `create_prod_build` (which btw actually lists `arrangement_version` as valid parameter).

twaugh: Huh.

Well, I don't really mind. Given that it's not changing what we had anyway, maybe this PR is fine as-is, moving `--arrangement-version` out of the group parser.

lcarva:
@bfontecc007, merge conflicts - my bad :smiling_imp:

twaugh:
@vrutkovs, it's no worse than the existing situation -- those parameters are not actually required. The help text says they are required, but if you look at the actual add_argument() calls they are not really required. It's because we're just grouping together the worker parameters for the purpose of display, rather than having them in their own parser. They're all still in the build parser.

vrutkovs:
@twaugh I think there is no point to remove the args from worker cli to fix that

PR #665, opened 2017-08-10
osbs/api: add get_orchestrator_build_logs API method

jarodwilson:
The method is a generator function that returns objects with these
attributes:

platform: str, platform name if worker build, else None
line: str, log line (decoded utf-8)

The method leverages an updated get_build_logs(), calling it with it's
newly acquired parameter, decode=True.

Basic self-tests of the method are included.


osbs/api.py
```diff
@@ -713,6 +713,45 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
 
         return logs
 
+    def _parse_build_log_entry(self, entry):
+        items = entry.split()
+        if len(items) < 4:
+            raise OsbsValidationException("This is not a valid build log entry")
```
lcarva: Maybe just a warning? Otherwise this may cause the build to fail.

osbs/api.py
```diff
@@ -713,6 +713,45 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
 
         return logs
 
+    def _parse_build_log_entry(self, entry):
+        items = entry.split()
+        if len(items) < 4:
+            logger.warning("This is not a valid build log entry: %s", entry)
```
twaugh: Return `(None, entry)` in this case?

jarodwilson: Ah, good call, otherwise, we'll potentially error out just a bit further down when trying to access the items by indices that aren't there.

twaugh:
Looks great! Just needs some additional test coverage for the error cases, and for the case where we get a generator from `get_build_logs()`.

osbs/api.py
```diff
@@ -713,6 +713,46 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+            return (None, entry)
+        date = items[0]
+        time = items[1]
+        platform = items[2]
+        plen = len(date) + len(time) + len(platform) + 3
```
vrutkovs: `3` here is the length of separator - ` - `? It would be better to add a comment or use `self.separator = ' - '` and `len(self.separator)` here

jarodwilson: No, it's actually the number of spaces also in the entire string to sum up the chars to chop off the front end to get just the log sub-string without date, time and platform included.

twaugh: Would this make that clearer?:
```
plen = sum(len(field) + 1  # include following space
           for field in [date, time, platform])
```

jarodwilson: Yeah, I'll go with that, it does make it much clearer what that constant is there for.

osbs/api.py
```diff
@@ -681,7 +681,7 @@ def create_orchestrator_build(self, **kwargs):
 
     def _decode_build_logs_generator(self, logs):
         for line in logs:
-            line = line.decode("utf-8")
+            line = line.decode("utf-8").strip()
```
twaugh: What's the stripping for?

osbs/api.py
```diff
@@ -713,6 +713,47 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+        :param build_id: str
+        :param follow: bool, fetch logs as they come?
+        :param wait_if_missing: bool, if build doesn't exist, wait
+        :return: tuple with fields 'platform' and 'line'
+        """
+        logs = self.get_build_logs(build_id=build_id, follow=follow,
+                                   wait_if_missing=wait_if_missing, decode=True)
+
+        if not isinstance(logs, GeneratorType):
+            for entry in logs.split("\n"):
```
twaugh: Could use `logs.splitlnes()` -- that would avoid a final empty string.

osbs/api.py
```diff
@@ -713,6 +713,47 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+
+        :param build_id: str
+        :param follow: bool, fetch logs as they come?
+        :param wait_if_missing: bool, if build doesn't exist, wait
+        :return: tuple with fields 'platform' and 'line'
+        """
+        logs = self.get_build_logs(build_id=build_id, follow=follow,
+                                   wait_if_missing=wait_if_missing, decode=True)
+
+        if not isinstance(logs, GeneratorType):
```
twaugh: Any reason for the `not` here? When there's a `if` with an `else`, it might make it slightly easier to read if the condition isn't negated.

osbs/api.py
```diff
@@ -713,6 +713,47 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+            for entry in logs.split("\n"):
+                yield self._parse_build_log_entry(entry)
+        else:
+            for entries in logs:
+                for entry in entries.split("\n"):
```
twaugh: I think the generator function will always return "a log entry", in as much as it calls `iterlines()` on the streamed `requests.Response` object and yields what it parses from each JSON object. Given that, I don't think we need to then iterate over each line within that as it will only be a single line anyway.

osbs/api.py
```diff
@@ -713,6 +713,47 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
 
         return logs
 
+    def _parse_build_log_entry(self, entry):
```
twaugh: Very minor: since this method doesn't use `self`, it could be made into a "static method", i.e. a function defined in the class but without adding the instance as the first argument:
```
@staticmethod
def _parse_build_log_entry(entry):
```
This is what landscape.io is on about here: https://landscape.io/github/projectatomic/osbs-client/pulls/665

osbs/api.py
```diff
@@ -713,6 +713,47 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+            logger.warning("This is not a valid build log entry: %s", entry)
+            return (None, entry)
+        date = items[0]
+        time = items[1]
+        platform = items[2]
+        plen = sum(len(field) + 1  # include trailing space
+                   for field in [date, time, platform])
+        if not platform.startswith("platform:"):
+            logger.warning("Line logged without using the appropriate LoggerAdapter")
+            platform = None
```
twaugh: These two lines aren't covered by unit testing.

osbs/api.py
jarodwilson: In some cases, I'm getting newline characters along with the line, so the strip was necessary to get the output fully sanitized.

osbs/api.py
jarodwilson: Ah, okay, will update accordingly.

osbs/api.py
jarodwilson: I'll throw a 3rd log line in the faked log return data that should match that.

osbs/api.py
jarodwilson: I think there was an earlier iteration of the code I had locally that didn't have the else, so it's legacy. I'll invert those for readability.

osbs/api.py
jarodwilson: Done.

osbs/api.py
jarodwilson: Similar to needing the .strip() to remove newline chars on decode, the newline chars also seem to get passed through in a single entry, at least with the self-tests, so I was actually getting multiple lines as a single 'entries', thus the nested for loop to split that up. In actual use, maybe it's not necessary, but it definitely was for the self-tests, so... I dunno.

osbs/api.py
jarodwilson: Side note: I'm not quite sure how to verify in the self-test that it actually does trip that warning, other than to cause a failure and look at the py.test debug and warning messages near the failure (which are indeed showing that line tripped w/my latest addition).

osbs/api.py
```diff
@@ -713,6 +713,57 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+        if len(items) < 4:
+            logger.warning("This is not a valid build log entry: %s", entry)
+            return (None, entry)
+
+        platform = items[2]
+        if not platform.startswith("platform:"):
+            logger.warning("Line logged without using the appropriate LoggerAdapter")
+            return (None, entry)
+
+        platform = platform.split(":")[1]
```
twaugh: We won't have platform names that themselves contain ":" I'm sure, but just to avoid adding that additional constraint could we use `split(':', 1)[1]` here?

jarodwilson: Yeah, simple enough to add. Did that and squashed the formatting rework, pushed again.

twaugh:
Draft release notes updated.

osbs/api.py
```diff
@@ -713,6 +713,57 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+
+        platform = platform.split(":", 1)[1]
+        if platform == "-":
+            return (None, entry)  # proper orchestrator build log entry
+
+        # Anything else should be a worker build log entry, so we strip off
+        # the leading 8 wrapping orchestrator log fields:
+        # <date> <time> <platform> - <name> - <level> -
+        plen = sum(len(items[i]) + 1  # include trailing space
+                   for i in range(8))
```
twaugh: This fails with an IndexError if there are fewer than 8 fields.

osbs/api.py
```diff
@@ -713,6 +713,57 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+        line = entry[plen:]
+        # if the 3rd field is "platform:-", we strip it out
+        items = line.split()
+        if items[2] == "platform:-":
```
twaugh: This should be `if len(items) >= 2 and items[2] == 'platform:-':`. Example log line that triggers failure:
```
2017-08-17 05:04:51,334 platform:x86_64 - atomic_reactor.plugins.orchestrate_build - INFO - "ContainersPaused": 0,
```

osbs/api.py
```diff
@@ -713,6 +713,57 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
 
         return logs
 
+    @staticmethod
+    def _parse_build_log_entry(entry):
+        items = entry.split()
+        if len(items) < 4:
+            logger.warning("This is not a valid build log entry: %s", entry)
```
twaugh: I think we should remove this. It will happen very frequently.

osbs/api.py
```diff
@@ -713,6 +713,57 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+            return (None, entry)
+
+        platform = items[2]
+        if not platform.startswith("platform:"):
+            logger.warning("Line logged without using the appropriate LoggerAdapter")
```
twaugh: I think we should remove this one too. Replace them with comments maybe?

osbs/api.py
```diff
@@ -713,6 +713,57 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+        plen = sum(len(items[i]) + 1  # include trailing space
+                   for i in range(8))
+        line = entry[plen:]
+        # if the 3rd field is "platform:-", we strip it out
+        items = line.split()
+        if items[2] == "platform:-":
+            plen = sum(len(items[i]) + 1  # include trailing space
+                       for i in range(3))
+            line = "%s %s %s" % (items[0], items[1], line[plen:])
+        return (platform, line)
```
twaugh: The docs say this we'll yield an object with attributes `platform` and `line`. One way of doing that is to declare a namedtuple (at module scope):
```
from collections import namedtuple
LogEntry = namedtuple('LogEntry', ['platform', 'line'])
```

Then you can replace this return value with an instance of that new class:
```
return LogEntry(platform, line)
```

osbs/api.py
```diff
@@ -713,6 +713,57 @@ def get_build_logs(self, build_id, follow=False, build_json=None, wait_if_missin
...
+
+    @osbsapi
+    def get_orchestrator_build_logs(self, build_id, follow=False, wait_if_missing=False):
+        """
+        provide logs from orchestrator build
+
+        :param build_id: str
+        :param follow: bool, fetch logs as they come?
+        :param wait_if_missing: bool, if build doesn't exist, wait
+        :return: tuple with fields 'platform' and 'line'
```
twaugh: For 'return' it should be: "generator yielding objects with attributes 'platform' and 'line'"

PR #667, opened 2017-08-11
remove trigger plugins, for scratch builds

rcerven:

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -547,6 +547,8 @@ class BuildRequest(object):
         and should not be imported into Koji.
         """
         if self.scratch:
+            self.template['spec'].pop('triggers', None)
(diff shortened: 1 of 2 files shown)


twaugh:
Why not make these changes?:
1. update `adjust_for_scratch()` so it removes any triggers -- we never want triggers in scratch builds, and in any case they are Builds not BuildConfigs so triggers don't even make sense
2. update `render()` so it runs `adjust_for_scratch()` prior to `adjust_for_triggers()` -- to reflect the fact that adjust_for_scratch might create work for adjust_for_triggers

That way the triggers will be removed, and then check_and_set_rebuild will be removed as a consequence.

rcerven:
changed

rcerven:
updated release note

PR #670, opened 2017-08-17
Fixes for get_orchestrator_build_logs

twaugh:


twaugh:
@jarodwilson, please review.

jarodwilson:
I'm a late-riser, was already merged by the time I was conscious today, but for the sake of completeness, looks good to me. Didn't realize any of the warnings would be triggered frequently in actual use, and the self-test changes make that look a whole lot cleaner.

twaugh:
Great, thanks for taking a look.

PR #671, opened 2017-08-22
Run pulp_pull before koji_upload so that an updated image id would be stored in koji

vrutkovs:


lcarva:
Changes look good for arrangement 3.

What about arrangement 4?
In that case, `pulp_pull` only runs on orchestrator, which happens after worker's `koji_upload`.

vrutkovs:
Arrangement 4 already have those in place, but would require a fix in atomic-reactor - https://github.com/projectatomic/atomic-reactor/pull/780

twaugh:
No release note required.

PR #672, opened 2017-08-24
Strip log lines of trailing spaces only - this breaks exception traceback formatting

vrutkovs:


twaugh:
@jarodwilson, is removing this strip() a problem? https://github.com/projectatomic/osbs-client/pull/665#discussion_r132914664

twaugh:
Would be nice to include a test for this.

vrutkovs:
Will do, I'm thinking `rstrip()` would be a better solution, although I haven't seen extra newline chars yet

jarodwilson:
If I'm remembering correctly, not having the strip there caused some minor formatting oddness with self-tests, but I think they could be adjusted slightly to compensate if that strip is really only useful for the self-test.

jarodwilson:
Okay, never mind, the .strip() seems to be entirely superfluous now. I recall it being needed for self-tests for some iteration of this code, but it seems to not be now, so yeah, have at it, remove it entirely!

lcarva:
Not updating release notes. This is a bugfix to unreleased changes.

PR #673, opened 2017-08-24
build_request: render_group_manifests now supports true grouping

mlangsdorf:
set group for post_group_manifests to true when the configuration file
has the pulp_manifest_lists value set to true to indicate the pulp
repository supports manifests lists.

### DIFF
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -178,6 +178,7 @@ class BuildSpec(object):
     koji_upload_dir = BuildParam("koji_upload_dir", allow_none=True)
     yum_proxy = BuildParam("yum_proxy", allow_none=True)
     koji_parent_build = BuildParam("koji_parent_build", allow_none=True)
+    group_manifests = BuildParam("group_manifests", allow_none=True)
 
     def __init__(self):
         self.required_params = [
@@ -229,7 +230,7 @@ class BuildSpec(object):
                    token_secrets=None, arrangement_version=None,
                    info_url_format=None, artifacts_allowed_domains=None,
                    equal_labels=None, koji_upload_dir=None, yum_proxy=None,
-                   koji_parent_build=None,
+                   koji_parent_build=None, group_manifests=None,
                    **kwargs):
         self.git_uri.value = git_uri
         self.git_ref.value = git_ref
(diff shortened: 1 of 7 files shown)


osbs/conf.py
```diff
@@ -286,6 +286,10 @@ def get_source_registry_uri(self):
     def get_pulp_registry(self):
         return self._get_value("pulp_registry_name", self.conf_section, "pulp_registry_name")
 
+    def get_pulp_manifest_lists(self):
+        return self._get_value("pulp_manifest_lists", self.conf_section,
```
lcarva: Is `pulp_manifest_lists` the correct name here? 
One could run atomic-reactor without pulp and this setting is still relevant.
Maybe something like `use_manifest_list` provides a clearer intent?

mlangsdorf: as I understand it, we only use manifest lists in atomic reactor if the pulp registry supports them. I'd like to keep pulp in the name to remind people that this is a pulp configuration issue.

vrutkovs: Manifest list functionality can be used with just docker registry

mlangsdorf: as a general issue, yes, manifest lists are independent of pulp registry. As a practical matter, if you enable manifest lists in atomic_reactor and your pulp registry doesn't support them, things will break and the point of this entry to is to keep those things from breaking.

mlangsdorf: changing it to group_manifests which is simpler and somewhat clearer

osbs/build/spec.py
```diff
@@ -286,6 +287,7 @@ def set_params(self, git_uri=None, git_ref=None,
         self.nfs_dest_dir.value = nfs_dest_dir
         self.git_branch.value = git_branch
         self.name.value = make_name_from_git(self.git_uri.value, self.git_branch.value)
+        self.group_manifests.value = group_manifests or False
```
lcarva: Is the `or False` needed here?

vrutkovs: I think its a good idea - this protects from accidental `None`

lcarva: No other BuildParam has that, trying to understand the special case.

lcarva: Anyways, this is a nitpick. I'm fine as is.

lcarva:
Updated release notes.

PR #674, opened 2017-08-24
Isolated Build Type

bfontecc007:
Allow for isolated build type.

In accordance with:
http://osbs.readthedocs.io/en/latest/isolated_builds.html

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -30,7 +30,8 @@ from osbs.constants import (BUILD_RUNNING_STATES, WORKER_OUTER_TEMPLATE,
                             ORCHESTRATOR_CUSTOMIZE_CONF, BUILD_TYPE_WORKER,
                             BUILD_TYPE_ORCHESTRATOR, BUILD_FINISHED_STATES)
 from osbs.core import Openshift
-from osbs.exceptions import OsbsException, OsbsValidationException, OsbsResponseException
+from osbs.exceptions import (OsbsException, OsbsValidationException, OsbsResponseException,
+                             OsbsOrchestratorNotEnabled)
 # import utils in this way, so that we can mock standalone functions with flexmock
 from osbs import utils
 
@@ -682,8 +683,8 @@ class OSBS(object):
             raise ValueError('Orchestrator build requires platforms param')
 
         if not self.can_orchestrate():
-            raise OsbsValidationException("can't create orchestrate build "
-                                          "when can_orchestrate isn't enabled")
+            raise OsbsOrchestratorNotEnabled("can't create orchestrate build "
+                                             "when can_orchestrate isn't enabled")
         extra = [x for x in ('platform',) if kwargs.get(x)]
         if extra:
             raise ValueError("Orchestrator build called with unwanted parameters: %s" %
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -28,7 +28,7 @@ from osbs.constants import (BUILD_RUNNING_STATES, WORKER_OUTER_TEMPLATE,
                             WORKER_INNER_TEMPLATE, WORKER_CUSTOMIZE_CONF,
                             ORCHESTRATOR_OUTER_TEMPLATE, ORCHESTRATOR_INNER_TEMPLATE,
                             ORCHESTRATOR_CUSTOMIZE_CONF, BUILD_TYPE_WORKER,
-                            BUILD_TYPE_ORCHESTRATOR)
+                            BUILD_TYPE_ORCHESTRATOR, BUILD_FINISHED_STATES)
 from osbs.core import Openshift
 from osbs.exceptions import OsbsException, OsbsValidationException, OsbsResponseException
 # import utils in this way, so that we can mock standalone functions with flexmock
@@ -105,7 +105,8 @@ class OSBS(object):
         self._bm = None
 
     @osbsapi
-    def list_builds(self, field_selector=None, koji_task_id=None):
+    def list_builds(self, field_selector=None, koji_task_id=None, running=None,
+                    labels=None):
         """
         List builds with matching fields
 
@@ -114,12 +115,20 @@ class OSBS(object):
         :return: BuildResponse list
         """
 
+        if running:
+            running_fs = ",".join(["status!={status}".format(status=status.capitalize())
+                                  for status in BUILD_FINISHED_STATES])
+            if not field_selector:
+                field_selector = running_fs
+            else:
+                field_selector = ','.join([field_selector, running_fs])
         response = self.os.list_builds(field_selector=field_selector,
-                                       koji_task_id=koji_task_id)
+                                       koji_task_id=koji_task_id, labels=labels)
         serialized_response = response.json()
         build_list = []
         for build in serialized_response["items"]:
             build_list.append(BuildResponse(build))
+
         return build_list
 
     def watch_builds(self, field_selector=None):
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -295,11 +295,17 @@ class OSBS(object):
             raise OsbsException(msg)
 
     def _create_scratch_build(self, build_request):
+        return self._create_build_directly(build_request)
+
+    def _create_isolated_build(self, build_request):
+        return self._create_build_directly(build_request,
+                                           unique=('git-repo-name', 'git-branch', 'isolated'))
+
+    def _create_build_directly(self, build_request, unique=None):
         logger.debug(build_request)
         build_json = build_request.render()
         build_json['kind'] = 'Build'
         build_json['spec']['serviceAccount'] = 'builder'
-        build_json['metadata']['labels']['scratch'] = 'true'
 
         builder_img = build_json['spec']['strategy']['customStrategy']['from']
         kind = builder_img['kind']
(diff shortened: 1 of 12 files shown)


bfontecc007:
This fails the arrangement version unit test which is expected behavior until we can remove the workaround code when arrangement 4 is ready.

inputs/orchestrator_inner:3.json
```diff
@@ -28,13 +28,6 @@
       }
     },
     {
-      "name": "koji_parent",
```
vrutkovs: This is certainly required

lcarva:
For clarification, the workaround code is in its own commit.

osbs/api.py
```diff
@@ -303,6 +318,16 @@ def _create_scratch_build(self, build_request):
             builder_img['kind'] = 'DockerImage'
             builder_img['name'] = ref
 
+        if unique:
+            # Query for a running build that matches the unique labels given.
+            # Probably need to enhance list_builds to accept a "running" parameter
```
lcarva: Let's remove this line and the one below ("Probably ... parameter.")

osbs/build/build_request.py
```diff
@@ -98,22 +100,34 @@ def set_params(self, **kwargs):
...
         self.explicit_build_node_selector = kwargs.get('explicit_build_node_selector', {})
         self.auto_build_node_selector = kwargs.get('auto_build_node_selector', {})
-        self.is_auto = kwargs.get('is_auto', False)
 
         logger.debug("setting params '%s' for %s", kwargs, self.spec)
         self.spec.set_params(**kwargs)
 
+    def validate_build_variation(self):
+        variations = (self.scratch, self.is_auto, self.isolated)
+        enabled_variations = list(filter(lambda x: x, variations))
```
vrutkovs: This `filter` looks useless to me, what does it do?

osbs/api.py
```diff
@@ -303,6 +318,16 @@ def _create_scratch_build(self, build_request):
...
+            # to avoid duplication from cli/main.py; and also take a labels parameter.
+            unique_labels = {}
+            for u in unique:
+                unique_labels[u] = build_json['metadata']['labels'][u]
+            if len(self.list_builds(running=True, labels=unique_labels)):
+                raise Exception('Matching build already running')
```
lcarva: It might be worth listing the ID/name of the build(s) already running.

osbs/api.py
```diff
@@ -520,13 +546,17 @@ def _do_create_prod_build(self, git_uri, git_ref,
...
         )
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
         build_request.set_repo_info(repo_info)
         if build_request.scratch:
             response = self._create_scratch_build(build_request)
+        elif build_request.isolated:
+            response = self._create_isolated_build(build_request)
         else:
             response = self._create_build_config_and_build(build_request)
+        # TODO: if not response ...
```
lcarva: Let's remove this.

osbs/build/build_request.py
```diff
@@ -890,9 +922,13 @@ def render_tag_from_config(self):
         unique_tag = self.spec.image_tag.value.split(':')[-1]
         tag_suffixes = {'unique': [unique_tag], 'primary': []}
 
-        if self.spec.build_type.value == BUILD_TYPE_ORCHESTRATOR and not self.scratch:
+        # These changes are just temporary
+        # TODO: revert this code block to run on orchestrator
+        if self.spec.build_type.value == BUILD_TYPE_WORKER and not (self.scratch or self.isolated):
```
lcarva: I'd like these if conditionals to be a little bit cleaner to read.
Something like this:

```python
if self.spec.build_type.value == BUILD_TYPE_WORKER:  # Will be BUILD_TYPE_ORCHESTRATOR once workaround is  not needed
  if self.scratch:
    # No additional tags for scratch builds
    pass 
  elif self.isolated:
    tag_suffixes['primary'].extend(['{version}-{release}'])
  else:
    tag_suffixes['primary'].extend(['latest', '{version}', '{version}-{release}'])
    tag_suffixes['primary'].extend(self._repo_info.additional_tags.tags)
```

osbs/build/build_request.py
```diff
@@ -1250,7 +1286,17 @@ def render_name(self):
 
             _, salt, timestamp = name.rsplit('-', 2)
 
-            name = 'scratch-{0}-{1}'.format(salt, timestamp)
+            if self.scratch:
+                name = 'scratch-{0}-{1}'.format(salt, timestamp)
+            elif self.isolated:
+                # Maybe this doesn't belong here?
```
lcarva: Looking for alternatives, maybe a helper method in util?

inputs/orchestrator_inner:3.json
lcarva: It's part of the workaround commit.

osbs/build/build_request.py
lcarva: It removes Falsey values.

vrutkovs: What about
```
variations = (self.scratch, self.is_auto, self.isolated)
if variations.count(True) > 1:
    raise ValueError('Build variations are mutually exclusive. '
                               'Must set either scratch, is_auto, isolated, or none. ')
```?

Comment (lcarva) [reply osbs/build/build_request.py]: Sure, that would produce a similar result.

File: osbs/api.py [hunk 8]
```diff
@@ -303,6 +318,17 @@ def _create_scratch_build(self, build_request):
             builder_img['kind'] = 'DockerImage'
             builder_img['name'] = ref
 
+        if unique:
+            unique_labels = {}
+            for u in unique:
+                unique_labels[u] = build_json['metadata']['labels'][u]
+            running_builds = self.list_builds(running=True, labels=unique_labels)
+            if running_builds:
+                for build in running_builds:
```
lcarva: No need to iterate over this and log each one. Put the information right in the exception text.
This has the benefit of propagating that information back to Koji.
Easier to debug instead of digging through logs.

osbs/build/build_request.py
```diff
@@ -890,9 +921,14 @@ def render_tag_from_config(self):
         unique_tag = self.spec.image_tag.value.split(':')[-1]
         tag_suffixes = {'unique': [unique_tag], 'primary': []}
 
-        if self.spec.build_type.value == BUILD_TYPE_ORCHESTRATOR and not self.scratch:
-            tag_suffixes['primary'].extend(['latest', '{version}', '{version}-{release}'])
-            tag_suffixes['primary'].extend(self._repo_info.additional_tags.tags)
+        if self.spec.build_type.value == BUILD_TYPE_ORCHESTRATOR:
+            if self.scratch:
+                pass
+            if self.isolated:
```
lcarva: This should be `elif`

osbs/api.py
```diff
@@ -114,12 +115,20 @@ def list_builds(self, field_selector=None, koji_task_id=None):
         :return: BuildResponse list
         """
 
+        if running:
+            running_fs = ",".join(["status!={status}".format(status=status.capitalize())
+                                  for status in BUILD_FINISHED_STATES])
+            if not field_selector:
+                field_selector = running_fs
+            else:
+                field_selector += running_fs
```
lcarva: this needs a comma: 

```python
field_selector = ','join([field_selector, running_fs])
```

osbs/api.py
```diff
@@ -104,8 +104,17 @@ def __init__(self, openshift_configuration, build_configuration):
                             namespace=self.os_conf.get_namespace())
         self._bm = None
 
+    def _has_each_label(self, labels_needle, labels_haystack):
+        for (k, v) in labels_needle.iteritems():
+            if k not in labels_haystack:
+                return False
+            if labels_haystack[k] != v:
+                return False
+        return True
```
twaugh: This (and the labels parameter for list_builds) is removed in a later commit. Does this need to be squashed/revised?

osbs/utils.py
```diff
@@ -387,6 +387,22 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
...
+    """
+    wraps the result of make_name_from_git in a suffix and postfix
+    adding separators for each.
+
+    see docstring for make_name_from_git for a full list of parameters
+    """
+    # 64 is maximum length allowed by OpenShift
+    # 2 is the number of dashes that will be added
+    prefix = ''.join(filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(prefix)))
+    suffix = ''.join(filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(suffix)))
```
twaugh: landscape.io suggests `''.join([VALID_BUILD_CONFIG_NAME_CHARS.match(ch) for ch in prefix])` etc
Is that easier to read? I tend to think what's already there is better.

osbs/constants.py
```diff
@@ -89,3 +90,5 @@
 
 BUILD_TYPE_ORCHESTRATOR = object()
 BUILD_TYPE_WORKER = object()
+
+ISOLATED_RELEASE_FORMAT = re.compile('^\d+\.\d+(\..+)?$')
```
twaugh: Use 'r' prefix to calm landscape.io: `re.compile(r'^\d+\.\d+(\..+)?$')`

osbs/utils.py
lcarva: Agree, the existing one seems cleaner to me.

osbs/api.py
lcarva: Yup. Let me work that out.

lcarva: Done.

osbs/constants.py
lcarva: Done.

osbs/api.py
```diff
@@ -286,11 +296,17 @@ def _verify_no_running_builds(self, build_config_name):
             raise OsbsException(msg)
 
     def _create_scratch_build(self, build_request):
+        return self._create_build_directly(build_request)
```
rcerven: shouldn't it be passing unique('scratch') ?

lcarva: Nope. That would allow only a single scratch build at a time in the whole system.
Also, if we do `unique=('git-repo-name', 'git-branch', 'scratch')` it will restrict a single scratch build per repo-branch at a time, which we do not want.
Isolated builds are different, as it should only allow a single build at a time for any given repo-branch.

lcarva:
Release notes updated.

PR #679, opened 2017-08-29
Add prefer_schema1_digest configuration option

lcarva:
When defined, this configuration option is passed to koji_upload
plugin so it can be used in deciding which manifest should be
used in the build output files.
This option is also forwarded to worker builds.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -274,6 +274,7 @@ class BuildRequest(object):
             'koji_root': self.spec.kojiroot.value,
             'openshift_required_version': sanitize_version(self._openshift_required_version),
             'pulp_registry_name': self.spec.pulp_registry.value,
+            'prefer_schema1_digest': self.spec.prefer_schema1_digest.value,
             'registry_api_versions': ','.join(self.spec.registry_api_versions.value or []) or None,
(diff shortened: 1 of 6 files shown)


osbs/build/build_request.py
```diff
@@ -737,24 +738,29 @@ def render_koji_promote(self, use_auth=None):
...
+
+        set_arg('kojihub', self.spec.kojihub.value)
+        set_arg('url', self.spec.builder_openshift_url.value)
+        set_arg('build_json_dir', self.spec.builder_build_json_dir.value)
+        set_arg('koji_upload_dir', self.spec.koji_upload_dir.value)
+
+        if use_auth is not None:
+            set_arg('use_auth', use_auth)
+
+        if self.spec.prefer_schema1_digest.value:
```
twaugh: Shouldn't this be `self.spec.prefer_schema1_digest.value is not None`? The parameter defaults to True... how is it set to False?

lcarva: Good catch! It also points out my tests aren't good enough... fixing it.

lcarva: Fixed.

twaugh:
Conflicts already!

lcarva:
> Conflicts already!

Fixed.

lcarva:
Release notes updated.

PR #682, opened 2017-08-31
arrangement 4: make arrangement 4 the default

mlangsdorf:
set DEFAULT_ARRANGEMENT_VERSION to 4 and adjust some test cases to
account for it.

### DIFF
--- a/tests/build_/test_build_request.py
+++ b/tests/build_/test_build_request.py
@@ -954,7 +954,7 @@ class TestBuildRequest(object):
         ['v2'],
     ])
     @pytest.mark.parametrize('platform', [None, 'x86_64'])
-    @pytest.mark.parametrize('arrangement_version', [3, 4])
+    @pytest.mark.parametrize('arrangement_version', range(3, DEFAULT_ARRANGEMENT_VERSION + 1))
     @pytest.mark.parametrize('scratch', [False, True])
     def test_render_prod_request_v1_v2(self, registry_api_versions, platform, arrangement_version,
                                        scratch):
@@ -1450,10 +1450,8 @@ class TestBuildRequest(object):
         (['x86_64', 'ppc64le'], None, False),
         (None, None, True),
     ))
-    @pytest.mark.parametrize('arrangement_version', [
-        # Only one version defined so far
-        DEFAULT_ARRANGEMENT_VERSION,
-    ])
+    @pytest.mark.parametrize('arrangement_version',
+                             range(3, DEFAULT_ARRANGEMENT_VERSION + 1))
     @pytest.mark.parametrize('koji_parent_build', ['fedora-26-9', None])
     @pytest.mark.parametrize(('build_image', 'build_imagestream', 'worker_build_image', 'valid'), (
         ('fedora:latest', None, 'fedora:latest', True),
@@ -1469,6 +1467,7 @@ class TestBuildRequest(object):
             'kojihub': 'kojihub',
             'kojiroot': 'kojiroot',
             'pulp_registry': 'pulp_registry',
+            'pulp_secret': 'pulp_secret',
             'registry_api_versions': ['v1', 'v2'],
             'smtp_additional_addresses': ['spam@food.bz', 'bacon@food.bz'],
             'smtp_email_domain': 'smtp_email_domain',
--- a/osbs/constants.py
+++ b/osbs/constants.py
@@ -1,5 +1,5 @@
 """
-Copyright (c) 2015 Red Hat, Inc
+Copyright (c) 2015, 2016, 2017 Red Hat, Inc
 All rights reserved.
 
 This software may be modified and distributed under the terms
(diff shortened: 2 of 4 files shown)


tests/build_/test_build_request.py
```diff
@@ -954,7 +954,7 @@ def test_render_prod_request_requires_newer(self):
         ['v2'],
     ])
     @pytest.mark.parametrize('platform', [None, 'x86_64'])
-    @pytest.mark.parametrize('arrangement_version', [3, 4])
+    @pytest.mark.parametrize('arrangement_version', [3, DEFAULT_ARRANGEMENT_VERSION])
```
lcarva: Instead of `[3, DEFAULT_ARRANGEMENT_VERSION]`, should this be: `range(3, DEFAULT_ARRANGEMENT_VERSION+1)` ?

That way when we up to arrangement 5, 4 is not left untested.

mlangsdorf: I think range(3, DEFAULT_ARRANGEMENT) since we shouldn't generally test an arrangement greater than DEFAULT which is normally the highest available. I'll make the change.

tests/build_/test_build_request.py
```diff
@@ -1469,6 +1468,7 @@ def test_render_optional_plugins(self, tmpdir):
             'kojihub': 'kojihub',
             'kojiroot': 'kojiroot',
             'pulp_registry': 'pulp_registry',
+            'pulp_secret': 'pulp_secret',
```
lcarva: This is odd... Why is it needed now? Seems to be that pulp_secret shouldn't be passed along to worker builds.

lcarva:
I wonder if we should undo this [change](https://github.com/projectatomic/osbs-client/pull/655/files#diff-625efbbf35a2290861f596fee1b19838) in `tests/test_api.py`
@twaugh thoughts?

tests/build_/test_build_request.py
lcarva: The limit of the `range` function is non-inclusive thus the `+1`:
```python
>>> range(3, 4)
[3]
>>> range(3, 4+1)
[3, 4]
```

tests/build_/test_build_request.py
mlangsdorf: It's needed for the orchestrator build to render pulp_tag, apparently.

mlangsdorf:
Ah, I missed that. Fixed.

tests/build_/test_build_request.py
```diff
@@ -954,7 +954,7 @@ def test_render_prod_request_requires_newer(self):
         ['v2'],
     ])
     @pytest.mark.parametrize('platform', [None, 'x86_64'])
-    @pytest.mark.parametrize('arrangement_version', [3, 4])
+    @pytest.mark.parametrize('arrangement_version', range(3, DEFAULT_ARRANGEMENT_VERSION))
```
twaugh: Missing a "+ 1" on the end limit.

```
>>> range(3,4)
[3]
```

tests/build_/test_build_request.py
```diff
@@ -1450,10 +1450,8 @@ def test_render_optional_plugins(self, tmpdir):
         (['x86_64', 'ppc64le'], None, False),
         (None, None, True),
     ))
-    @pytest.mark.parametrize('arrangement_version', [
-        # Only one version defined so far
-        DEFAULT_ARRANGEMENT_VERSION,
-    ])
+    @pytest.mark.parametrize('arrangement_version',
+                             range(3, DEFAULT_ARRANGEMENT_VERSION))
```
twaugh: Missing a "+ 1" on the end limit.

tests/build_/test_build_request.py
twaugh: Was this not needed after all?

tests/build_/test_build_request.py
mlangsdorf: Fixed.

tests/build_/test_build_request.py
mlangsdorf: Fixed.

tests/build_/test_build_request.py
```diff
@@ -1525,6 +1523,9 @@ def test_render_orchestrate_build(self, tmpdir, platforms, secret, disabled,
             kwargs['koji_parent_build'] = koji_parent_build
         if prefer_schema1_digest is not None:
             kwargs['prefer_schema1_digest'] = prefer_schema1_digest
+        # required for pulp_ plugins in the orchestrator build
+        if arrangement_version > 3:
+            kwargs['pulp_secret'] = 'pulp_secret'
```
lcarva: I'm assuming this is needed because orchestrator build uses pulp registry starting on arrangement version 4, and whenever a `pulp_registry_name` is provided, `pulp_secret` is required.

What's puzzling me is that I don't see how `pulp_registry_name` is being set for this test run which make me wonder if we have an incorrect check somewhere.

lcarva: Aha! It does get set by `additional_kwargs` param.
It didn't complain before because pulp registry was actually never used by orchestrator build.
I think `pulp_secret` should be set alongside it, around line 1469.

lcarva:
Updated release notes draft.

PR #683, opened 2017-08-31
koji_upload takes platform parameter instead of buildstep_logs

twaugh:


twaugh:
Draft release notes updated.

PR #684, opened 2017-09-07
Perform stricter validation of additional tags

lcarva:
When processing of additional-tags file was moved to
osbs-client from atomic-reactor, the validation check
became more permissive. This commit ports the validation
algorithm from atomic-reactor's tag_from_config plugin.


lcarva:
Updated release notes draft:
> - Stricter validation when processing tags in `additional-tags` file

PR #687, opened 2017-09-12
Make BuildParams instance attributes in BuildSpec

lcarva:
The BuildParams in BuildSpec were defined as class attributes.
This caused those attributes to not be unique among multiple
worker builds, muddling the values. This changes makes them
instance attributes to guarantee isolation.


lcarva:

PR #689, opened 2017-09-14
new remove worker metadata plugin

rcerven:

might be merged after this is merged https://github.com/projectatomic/atomic-reactor/pull/831

### DIFF
--- a/inputs/orchestrator_inner:4.json
+++ b/inputs/orchestrator_inner:4.json
@@ -163,6 +163,10 @@
     },
     {
       "name": "remove_built_image"
+    },
+    {
+      "name": "remove_worker_metadata",
+      "required": false
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


inputs/orchestrator_inner:4.json
```diff
@@ -163,6 +163,9 @@
     },
     {
       "name": "remove_built_image"
+    },
+    {
+      "name": "remove_worker_metadata"
```
twaugh: Since removing the configmap after use is not essential for a successful build, I would suggest adding:
```
"required": False
```

so that osbs-client can still create successful builds even when using a builder image that does not have the remove_worker_metadata plugin present.

rcerven: added

rcerven:
release note updated

PR #690, opened 2017-09-18
spec.py: don't track base image for flatpak builds

vrutkovs:
Flatpak builds fail with `OsbsValidationException: base_image must be provided`

### DIFF
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -315,31 +315,33 @@ class BuildSpec(object):
         self.name.value = make_name_from_git(self.git_uri.value, self.git_branch.value)
         self.group_manifests.value = group_manifests or False
         self.prefer_schema1_digest.value = prefer_schema1_digest
-        if not base_image:
-            raise OsbsValidationException("base_image must be provided")
-        self.trigger_imagestreamtag.value = get_imagestreamtag_from_image(base_image)
-        self.builder_build_json_dir.value = builder_build_json_dir
-        if not name_label:
-            raise OsbsValidationException("name_label must be provided")
-        self.imagestream_name.value = name_label.replace('/', '-')
-        # The ImageStream should take tags from the source registry
-        # or, if no source registry is set, the first listed registry
-        imagestream_reg = self.source_registry_uri.value
-        if not imagestream_reg:
-            try:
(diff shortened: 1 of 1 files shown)


lcarva:
Flatpak builds don't need a base_image? ever?

vrutkovs:
>Flatpak builds don't need a base_image? ever?

Flatpak builds use `flatpak_base_image` param from osbs-client

osbs/build/spec.py
```diff
@@ -315,31 +315,33 @@ def set_params(self, git_uri=None, git_ref=None,
...
-            self.imagestream_url.value = os.path.join(imagestream_reg.docker_uri,
-                                                      name_label)
-            logger.debug("setting 'imagestream_url' to '%s'",
-                         self.imagestream_url.value)
-            insecure = imagestream_reg.uri.startswith('http://')
-            self.imagestream_insecure_registry.value = insecure
-            logger.debug("setting 'imagestream_insecure_registry' to %r", insecure)
+
+        if not flatpak:
+            if not base_image:
```
lcarva: Shouldn't this be `if not base_image and not flatpak_base_image`?

Why is everything else not applicable to flatpaks?

vrutkovs: Flatpaks have updatable runtime these don't need to be rebuild whenever base image change. See [docs](http://docs.flatpak.org/en/latest/introduction.html#runtimes)

lcarva:

osbs/build/spec.py
```diff
@@ -315,31 +315,33 @@ def set_params(self, git_uri=None, git_ref=None,
         self.name.value = make_name_from_git(self.git_uri.value, self.git_branch.value)
         self.group_manifests.value = group_manifests or False
         self.prefer_schema1_digest.value = prefer_schema1_digest
-        if not base_image:
-            raise OsbsValidationException("base_image must be provided")
-        self.trigger_imagestreamtag.value = get_imagestreamtag_from_image(base_image)
-        self.builder_build_json_dir.value = builder_build_json_dir
```
vrutkovs: This one doesn't belong here - and breaks worker builds for flatpaks :( Fixed in #692

PR #691, opened 2017-09-19
spec.py: remove group_manifests in flatpak builds

vrutkovs:
Making a multi-arch aware flatpak is more complex than making a manifest list,
so group_manifests should be disabled for now

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -1230,6 +1230,12 @@ class BuildRequest(object):
         if not self.dj.dock_json_has_plugin_conf('postbuild_plugins', 'group_manifests'):
             return
 
+        if self.spec.flatpak.value:
+            logger.info("removing group_manifests from request "
+                        "because this is a Flatpak built from a module")
+            self.dj.remove_plugin('postbuild_plugins', 'group_manifests')
+            return
+
         push_conf = self.dj.dock_json_get_plugin_conf('postbuild_plugins',
                                                       'group_manifests')
         args = push_conf.setdefault('args', {})


osbs/build/build_request.py
```diff
@@ -1230,6 +1230,12 @@ def render_group_manifests(self):
         if not self.dj.dock_json_has_plugin_conf('postbuild_plugins', 'group_manifests'):
             return
 
+        if self.spec.flatpak.value:
+            logger.info("removing group_manifests from request "
+                        "because this is a Flatpak built from a module")
+            self.dj.remove_plugin('postbuild_plugins', 'group_manifests')
```
twaugh: I don't think we want to remove the plugin altogether. Wouldn't we instead want to ensure that its "group" parameter is set to False?

Better still would be to raise an OsbsValidationException if `flatpak` is set for the build and `group_manifests` is set for the instance.

vrutkovs: It was failing for me when it was set to `False`. There is not reason to have it here anyway:
* `group=True` doesn't make sense for flatpaks - these don't use manifest lists (flatpak automatically splits arch-specific image in a branch and pulls the branch based on client's arch
* `group=False` is not vital for flatpak workflow - currently only koji archives are used as a source.

This, however, is a temporary solution, a coordination with Fedora folks is required to proceed, as its not yet clear if OCI registry would be used as a source in future

lcarva: Let's add a unit test for this.

lcarva:
I was under the impression that flatpaks still care about container registry tags.
Without group_manifestswon't the image be tagged solely with the arch specific tag.

There are two plugins that handle this: `tag_and_push` and `group_manifests`. With 
this change neither will run in orchestrator build for arrangement version 4.

owtaylor:
The expectation is that for Fedora (and RHEL) we will be distributing Flatpaks via the Docker registry protocol. While the network efficiency is less than with OSTree, it keeps things a lot simpler by not adding a totally new type of artifact that has to be managed by release engeering internally and sysadmins externally.

There is work that needs to be done in the client tooling to be able to download from a Docker registry, and also on the server side to provide an index to let Flatpak and GNOME Software browse the registry to see what is available for installation (prototype https://github.com/owtaylor/metastore/). The temporary test scripts download the Koji artifacts for simplicity, but that's not the final plan.

My plan has been to use manifest lists (or rather the equivalent OCI Image indexes) for multi-architecture support. If you were importing from an OCI image into a OSTree registry, then things would expand into multiple branches at that point, but it would be a single tag in the docker registry pointing to an image index.

I'll try to figure out what would be involved in making group_manifests work properly - if it's too hard to do while the multi-arch support is still evolving, then we can remove it for the time being as in this PR.

twaugh:
Does https://github.com/projectatomic/atomic-reactor/pull/846 make this no longer relevant?

vrutkovs:
Yep, closing

PR #692, opened 2017-09-19
spec.py: set builder_build_json_dir value for flatpaks too

vrutkovs:


lcarva:

PR #693, opened 2017-09-20
timeout for requests

rcerven:

### DIFF
--- a/osbs/constants.py
+++ b/osbs/constants.py
@@ -88,6 +88,9 @@ HTTP_RETRIES_STATUS_FORCELIST = [408, 500, 502, 503, 504]
 # HTTP methods that we should retry on
 HTTP_RETRIES_METHODS_WHITELIST = ['GET', 'PUT', 'POST', 'DELETE']
 
+# requests timeout in seconds
+HTTP_REQUEST_TIMEOUT = 600
+
 BUILD_TYPE_ORCHESTRATOR = object()
 BUILD_TYPE_WORKER = object()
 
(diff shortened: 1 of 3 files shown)


osbs/http.py
```diff
@@ -43,15 +43,19 @@ def __init__(self, verbose=False):
         self.verbose = verbose
 
     def get(self, url, **kwargs):
+        kwargs.setdefault('timeout', HTTP_REQUEST_TIMEOUT)
```
vrutkovs: All of those call `self.request`, it would be best to be set there

rcerven: yeah I've just realized it now :)

osbs/constants.py
```diff
@@ -88,6 +88,9 @@
 # HTTP methods that we should retry on
 HTTP_RETRIES_METHODS_WHITELIST = ['GET', 'PUT', 'POST', 'DELETE']
 
+# requests timeout in seconds
+HTTP_REQUEST_TIMEOUT = 120
```
twaugh: Please can we increase this to something like 600? cf. https://github.com/openshift/origin/blob/v3.6.0/pkg/cmd/server/api/types.go#L716-L718

rcerven:
increased timeout, and added catching 'Timeout'

as for the logs()
HttpSession will catch RetryError and raise OsbsNetworkException,
and logs() are catching only OsbsResponseException which would be raised by HttpResponse, if there wouldn't be any exception in HttpSession

rcerven:
test added

vrutkovs:
Release notes updated

PR #694, opened 2017-09-21
Always leave import_image enabled

lcarva:
When enabled in *_inner.json, import_image plugin needs to run
for all builds, even the ones that do not have autorebuild
enabled. Importing image into OpenShift affects the layered
images which may have autorebuild enabled.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -535,7 +535,6 @@ class BuildRequest(object):
 
         If there are triggers defined, and this is a custom
         base image, some trigger-related plugins do not apply.
-        All but import_image are disabled in this case.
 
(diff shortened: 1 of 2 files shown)


lcarva:
> add a test (that the plugin is present after rendering) to test_arrangements.py

The `import_image` plugin is only enabled in `prod_inner.json`. 
It needs a couple more tweaks before we can enable in `orchestrator_inner:4.json`.
It's not enabled in any of the arrangements.

twaugh:
Ah, ok.

lcarva:
Updated release notes draft:
> - `import_image` does not explicitly track `{version}-{release}` tags

PR #695, opened 2017-09-25
Make import_image remove dockerRepositoryCheck

lcarva:
The API method import_image is used to force OpenShift to
update the tags tracked in the corresponding ImageStream
object. To do so, the following annotation is cleared:
    "openshift.io/image.dockerRepositoryCheck"
The problem addressed by this commit is that in the case
an ImageStream does not pick up new tags, the API method
may hang indefinitely. Removing this annotation causes
OpenShift to actually "force" the import and reset the
annotation.

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -830,7 +830,7 @@ class Openshift(object):
         # Mark it as needing import
         imagestream_json['metadata'].setdefault('annotations', {})
         check_annotation = "openshift.io/image.dockerRepositoryCheck"
-        imagestream_json['metadata']['annotations'][check_annotation] = ''
+        imagestream_json['metadata']['annotations'].pop(check_annotation, None)
         response = self._put(url, data=json.dumps(imagestream_json),
                              use_json=True)
         check_response(response)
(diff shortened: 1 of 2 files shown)


lcarva:
```go
if hasAnnotation && !hasChanges {                                                           
    glog.V(4).Infof("stream did not change: %#v", stream)                                   
    obj, err = original.(*imageapi.ImageStream), nil                                        
} else {                                                                                    
    if glog.V(4) {                                                                          
        glog.V(4).Infof("updating stream %s", diff.ObjectDiff(original, stream))            
    }                                                                                       
    stream.Annotations[imageapi.DockerImageRepositoryCheckAnnotation] = now.UTC().Format(time.RFC3339)
    obj, _, err = r.internalStreams.Update(ctx, stream.Name, rest.DefaultUpdatedObjectInfo(stream, kapi.Scheme))
}                                                                                           
```

https://github.com/openshift/origin/blob/master/pkg/image/registry/imagestreamimport/rest.go#L331-L340

lcarva:
Updated release notes draft:
> - `import_image` hang while importing images has been fixed

PR #696, opened 2017-09-27
updating BuildConfig retry

rcerven:

### DIFF
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -6,7 +6,7 @@ This software may be modified and distributed under the terms
 of the BSD license. See the LICENSE file for details.
 """
 from __future__ import print_function, absolute_import, unicode_literals
-
+from functools import wraps
 import contextlib
 import copy
 import logging
@@ -19,12 +19,15 @@ import subprocess
 import sys
 import tempfile
 import tarfile
+import time
 import requests
 from collections import namedtuple
 from datetime import datetime
 from io import BytesIO
 from hashlib import sha256
 from osbs.repo_utils import RepoConfiguration, RepoInfo, AdditionalTagsConfig
+from osbs.constants import OS_CONFLICT_MAX_RETRIES, OS_CONFLICT_WAIT
+from six.moves import http_client
 
 try:
     # py3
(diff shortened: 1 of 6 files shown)


rcerven:
working on test

twaugh:
Can't `retry_on_conflict` help somehow? Seems a shame to duplicate logic.

osbs/utils.py
```diff
@@ -554,3 +556,22 @@ def get_name_and_value(self, label_type):
...
+def retry_on_conflict(func,
+                      retry_times=OS_CONFLICT_MAX_RETRIES,
+                      retry_delay=OS_CONFLICT_WAIT):
+    @wraps(func)
+    def retry(*args, **kwargs):
+        for counter in range(retry_times + 1):
+            try:
+                return func(*args, **kwargs)
+            except OsbsResponseException as ex:
+                if ex.status_code == 409 and counter != retry_times:
```
vrutkovs: Use `http_client.CONFLICT` instead of `409`

rcerven: ok

osbs/utils.py
```diff
@@ -554,3 +557,22 @@ def get_name_and_value(self, label_type):
...
             return (label_type, self._df_labels[label_type])
+
+
+def retry_on_conflict(func,
+                      retry_times=OS_CONFLICT_MAX_RETRIES,
+                      retry_delay=OS_CONFLICT_WAIT):
+    @wraps(func)
+    def retry(*args, **kwargs):
+        for counter in range(retry_times + 1):
+            try:
```
vrutkovs: We should log retries, that would help with debugging

tests/test_core.py
```diff
@@ -265,7 +266,7 @@ def test_get_multiple_build_config_by_labels(self, openshift):  # noqa
         ([http_client.CONFLICT, http_client.OK], False),
         ([http_client.CONFLICT, http_client.CONFLICT, http_client.UNAUTHORIZED], True),
         ([http_client.UNAUTHORIZED], True),
-        ([http_client.CONFLICT for _ in range(10)], True),
+        ([http_client.CONFLICT for _ in range(OS_CONFLICT_MAX_RETRIES + 1)], True),
```
vrutkovs: Is there a test case where it fails after all attempts are depleted?

rcerven: that test is in test_utils

and I'll make 1 more test which will conflict and success after

PR #701, opened 2017-10-09
conf.py: strip spaces when parsing values, separated by comma

vrutkovs:
This might lead to a wrong `OsbsValidationException('v2 only platform in v1 only instance')` error if space is put after a comma

### DIFF
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -276,14 +276,14 @@ class Configuration(object):
                                 self.conf_section,
                                 "registry_uri")
         if value:
-            return value.split(',')
+            return [x.strip() for x in value.split(',')]
         else:
             return []
 
     def get_registry_secrets(self):
         value = self._get_value("registry_secret", self.conf_section, "registry_secret")
         if value:
-            return value.split(',')
+            return [x.strip() for x in value.split(',')]
         else:
             return []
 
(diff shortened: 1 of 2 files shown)


osbs/conf.py
```diff
@@ -292,7 +292,7 @@ def get_registry_api_versions(self, platform=None):
                                 self.conf_section,
                                 "registry_api_versions",
                                 default='v1,v2')
-        versions = value.split(',')
+        versions = [x.strip() for x in value.split(',')]
```
twaugh: This pattern is used in two more places in this module. Maybe we should fix those as well?
```
$ git grep -n "value\.split(',')"
conf.py:279:            return value.split(',')
conf.py:286:            return value.split(',')
conf.py:295:        versions = [x.strip() for x in value.split(',')]
conf.py:396:            return [x.strip() for x in value.split(',')]
conf.py:403:            return [x.strip() for x in value.split(',')]
conf.py:540:            return [x.strip() for x in value.split(',')]
```

vrutkovs: Good idea, replaced other places where `split(',')` results were not stripped

lcarva:
Updated release notes draft:
> - Spaces are stripped when parsing values separated by comma

PR #704, opened 2017-10-12
Remove koji-task-id from build config

bfontecc007:
Remove koji task id from build config if build was created explicitly.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -375,6 +375,13 @@ class OSBS(object):
         old_nodeselector = existing_bc['spec'].pop('nodeSelector', None)
         logger.debug("removing build config's nodeSelector %s", old_nodeselector)
 
+        # Remove koji_task_id
+        koji_task_id = utils.graceful_chain_get(existing_bc, 'metadata', 'labels',
+                                                'koji-task-id')
+        if koji_task_id is not None:
+            logger.debug("removing koji-task-id %r", koji_task_id)
+            utils.graceful_chain_del(existing_bc, 'metadata', 'labels', 'koji-task-id')
+
         utils.buildconfig_update(existing_bc, build_json)
         # Reset name change that may have occurred during
         # update above, since renaming is not supported.
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -219,6 +219,79 @@ class TestOSBS(object):
                                           customize_conf=customize_conf)
         assert isinstance(response, BuildResponse)
 
+    @pytest.mark.parametrize('has_task_id', [True, False])
+    def test_create_prod_build_remove_koji_task_id(self, has_task_id):
+
+        build_config = {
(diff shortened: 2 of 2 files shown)


twaugh:
Shouldn't this happen in `_update_build_config_when_exist()`, before the `self.os.update_build_config()` call?

lcarva:
> Shouldn't this happen in _update_build_config_when_exist(), before the self.os.update_build_config() call?

Yes.

bfontecc007:
The recent changes threw me off a little

osbs/api.py
```diff
@@ -375,6 +375,12 @@ def _update_build_config_when_exist(self, build_json):
         old_nodeselector = existing_bc['spec'].pop('nodeSelector', None)
         logger.debug("removing build config's nodeSelector %s", old_nodeselector)
 
+        # Remove koji_task_id
+        koji_task_id = utils.graceful_chain_get(existing_bc, 'metadata', 'labels',
+                                                'koji-task-id')
+        logger.debug("removing koji-task-id %s", str(koji_task_id))
```
lcarva: maybe use `'%r'`, and not convert koji_task_id to str?

bfontecc007:
Added unit testing.

osbs/api.py
```diff
@@ -375,6 +375,12 @@ def _update_build_config_when_exist(self, build_json):
         old_nodeselector = existing_bc['spec'].pop('nodeSelector', None)
         logger.debug("removing build config's nodeSelector %s", old_nodeselector)
 
+        # Remove koji_task_id
+        koji_task_id = utils.graceful_chain_get(existing_bc, 'metadata', 'labels',
+                                                'koji-task-id')
+        logger.debug("removing koji-task-id %r", koji_task_id)
+        utils.graceful_chain_del(existing_bc, 'metadata', 'labels', 'koji-task-id')
```
twaugh: Wonder whether to guard these two lines with `if koji_task_id is not None:`, to avoid logs being too chatty.

tests/test_api.py
```diff
@@ -219,6 +219,76 @@ def test_create_prod_build_build_request(self, osbs, inner_template,
                                           customize_conf=customize_conf)
         assert isinstance(response, BuildResponse)
 
+    def test_create_prod_build_remove_koji_task_id(self):
```
twaugh: Can we verify that it behaves gracefully when there is already no "koji-task-id" annotation?

osbs/api.py
lcarva: I think in most cases, this will not be None though.
Only when the previous build was submitted via `osbs-client` directly.

bfontecc007:
I modified the test to address Tim's concerns

lcarva:
Updated release notes draft:
> - Previous `koji-task-id` is cleared

PR #705, opened 2017-11-03
don't run import_image for scratch build

rcerven:


vrutkovs:
Wouldn't it be more explicit to update `def adjust_for_scratch` instead?

rcerven:
right will do that, thanks

vrutkovs:
>ahh I can't do that, adjust_for_scratch is running before render_import_image

That's exactly the point of this change - remove the plugin as early as possible

rcerven:
ok I was looking wrong, will remove it in adjust_for_scratch

PR #710, opened 2017-11-22
Add SSL auth support for ODCS in resolve_composes

lcarva:


lcarva:
Updated release notes draft:
> - Add SSL auth support for ODCS

PR #712, opened 2017-11-28
Convert the value of --compose-id to int

csomh:
When triggering builds with osbs-client cli and using the
--compose-id option to specify composes to be used,
the following TypeError is raised when logging the ID of
a compose which is going to be renewed:

TypeError: %d format: a number is required, not unicode

This is due to the compose IDs being received as unicode
strings from the cli.

Setting the type for the argument to int solves this issue.

Signed-off-by: Hunor Csomortáni <csomh@redhat.com>


csomh:
@rcerven @lcarva ptal

lcarva:

PR #713, opened 2017-11-30
[do not merge] enable chained rebuilds in orchestrator workflow

vrutkovs:


lcarva:
Ok to drop this? 

Was addressed by https://github.com/projectatomic/osbs-client/pull/720.

vrutkovs:
Yes, sure

PR #714, opened 2017-11-30
Tests: freeze pytest and pytest-* versions

vrutkovs:


lcarva:
Looks like that fixes centos:7 but breaks everything else :angry:

vrutkovs:
Yeah, `setuptools` needs to be updated in centos :( Other packages are pinned though, so it should be more stable

lcarva:
> Yeah, setuptools needs to be updated in centos :( Other packages are pinned though, so it should be more stable

They are pinned when the file `tests/requirements.py` is being used directly. On the line I commented on, that gets lost. It should be fine as setuptools wouldn't require pytest.

PR #715, opened 2017-12-06
use random string for isolated build name

rcerven:


lcarva:
This looks good to me. Let's just fix up the unit tests.

rcerven:
unit tests doesn't need fix, but fedora image can't be pulled

lcarva:
Registry is going through maintenance, it appears: https://www.fedorastatus.org/q4maint.html

osbs/build/build_request.py
```diff
@@ -1421,8 +1421,7 @@ def render_name(self):
             if self.scratch:
                 name = 'scratch-{0}-{1}'.format(salt, timestamp)
             elif self.isolated:
-                name = wrap_name_from_git('isolated', timestamp, self.spec.git_uri.value,
-                                          self.spec.git_branch.value)
+                name = 'isolated-{0}-{1}'.format(salt, timestamp)
```
vrutkovs: A bit better version:
```
prefix = "scratch" if self.scratch else "isolated"
name = "-".join(prefix, salt, timestamp)
```

rcerven:
release note updated

PR #717, opened 2017-12-08
allow more isolated builds with different release

rcerven:


lcarva:
Updated release notes draft:
> - Multiple isolated builds of same branch-repo are allowed if different release values are used

PR #720, opened 2017-12-14
Enable rebuild triggers

bfontecc007:
Enable autorebuild by adding ImageChange trigger and activating import_image plugin.

### DIFF
--- a/inputs/orchestrator_inner:5.json
+++ b/inputs/orchestrator_inner:5.json
@@ -129,6 +129,16 @@
         "dockpulp_loglevel": "INFO",
         "publish": false
       }
+    },
+    {
+      "name": "import_image",
+      "args": {
+        "imagestream": "{{IMAGESTREAM}}",
+        "docker_image_repo": "{{DOCKER_IMAGE_REPO}}",
+        "url": "{{OPENSHIFT_URI}}",
+        "verify_ssl": false,
+        "build_json_dir": "{{PATH}}"
+      }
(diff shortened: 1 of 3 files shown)


inputs/orchestrator_inner:5.json
```diff
@@ -129,6 +129,16 @@
...
+    },
+    {
+      "args": {
+        "imagestream": "{{IMAGESTREAM}}",
+        "docker_image_repo": "{{DOCKER_IMAGE_REPO}}",
+        "url": "{{OPENSHIFT_URI}}",
+        "verify_ssl": false,
+        "build_json_dir": "{{PATH}}"
+      },
+      "name": "import_image"
```
lcarva: Let's list the name first like the other plugins.

tests/build_/test_arrangements.py
```diff
@@ -1198,6 +1198,7 @@ class TestArrangementV5(TestArrangementV4):
                 'group_manifests',
                 'pulp_tag',
                 'pulp_sync',
+                'import_image',
```
lcarva: Do we need a test to ensure the plugin parameters were rendered correctly?

bfontecc007: I'll code it if you think we need it.

lcarva:
Updated release notes draft:
> - ImageChangeTriggers enabled
> - `import_image` plugin enabled

PR #721, opened 2017-12-18
Check logs is not None before performing string operation

yahzaa:
Signed-off-by: Chandrakant Kumar <k.03chandra@gmail.com>

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -831,6 +831,8 @@ class OSBS(object):
         logs = self.get_build_logs(build_id=build_id, follow=follow,
                                    wait_if_missing=wait_if_missing, decode=True)
 
+        if logs is None:
+            return
         if isinstance(logs, GeneratorType):
             for entries in logs:
                 for entry in entries.splitlines():
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -947,6 +947,12 @@ class TestOSBS(object):
         assert isinstance(content, six.string_types)
         assert content == u"   líne 1"
 
+    def test_orchestrator_build_logs_no_logs(self, osbs):  # noqa:F811
+        flexmock(osbs).should_receive('get_build_logs').and_return(None)
+        logs = osbs.get_orchestrator_build_logs(TEST_BUILD)
+        assert isinstance(logs, GeneratorType)
+        assert list(logs) == []
+
     # osbs is a fixture here
     def test_pause_builds(self, osbs):  # noqa
         osbs.pause_builds()


lcarva:
Thanks for this!
It would be great to add a unit test to cover this case so it prevents it from breaking in the future. 
Let us know if you need any help with that.

osbs/api.py
```diff
@@ -827,10 +827,13 @@ def get_orchestrator_build_logs(self, build_id, follow=False, wait_if_missing=Fa
         :param follow: bool, fetch logs as they come?
         :param wait_if_missing: bool, if build doesn't exist, wait
         :return: generator yielding objects with attributes 'platform' and 'line'
+                 None, if logs doesn't exist
```
twaugh: Technically it returns a generator which yields no objects. In other words, the caller can always iterate over the value returned, but this line makes it appear as though they would need to check for identity with `None` before iterating.

yahzaa: In that case I should change the return statement to return an empty generator?

yahzaa:
> It would be great to add a unit test to cover this case so it prevents it from breaking in the future.
> Let us know if you need any help with that.
What could be the test case for this change?

osbs/api.py
twaugh: No. The method is already a generator function because it uses the `yield` keyword. This is different from a regular function in that calling the function immediately returns a generator. When the caller iterates over the generator, the called function runs up until the next `yield` statement.

When a generator function returns None, this means the iteration stops. The caller will see an empty iterator.

So what I described in the comment above is already the behaviour your change has, and in fact is exactly the behaviour we'd like to see... just the docstring doesn't describe it correctly.

twaugh:
> What could be the test case for this change?

Take a look at this test case, which is similar:
https://github.com/projectatomic/osbs-client/blob/0.44/tests/test_api.py#L966-L973

What you'll want to do is add a new method alongside that one, something like this:
```
    def test_orchestrator_build_logs_no_logs(self, osbs):
        flexmock(osbs).should_receive('get_build_logs').and_return(None)
        logs = osbs.get_orchestrator_build_logs(TEST_BUILD)
        assert isinstance(logs, GeneratorType)
        assert list(logs) == []
```

Because the method name starts with "test_" it will get called automatically when the test suite runs (when you push your changes). The `flexmock` line arranges for the `get_build_logs(...)` call to return `None`. It is followed by the actual call to the method we're testing. The remaining lines verify that it returns the result we expect, i.e. a generator which generates nothing.

tests/test_api.py
```diff
@@ -947,6 +947,12 @@ def test_orchestrator_build_logs_api_badlog(self, osbs):  # noqa
         assert isinstance(content, six.string_types)
         assert content == u"   líne 1"
 
+    def test_orchestrator_build_logs_no_logs(self, osbs):  # noqa
```
twaugh: You can use `# noqa:F811` to avoid a diagnostic about `osbs` being redefined. That way we'll still see any other diagnostics there might be. (I know other methods in that file also use `# noqa` without any specific diagnostic being named... they ought to be corrected too, but that's outside the scope of this PR.)

yahzaa: done

lcarva:
Updated release notes draft:
> - Check logs is not None before performing string operation

PR #722, opened 2018-01-09
Fail if image name begins with component containing '.'

twaugh:
Addresses part of OSBS-4908.


twaugh:
Re-based, now the tests run clean.

lcarva:
Updated release notes draft:
> Fail if image name begins with component containing '.'

PR #724, opened 2018-02-01
cleanup flatpak

rcerven:


inputs/orchestrator_inner:6.json
```diff
@@ -0,0 +1,204 @@
+{
```
lcarva: These changes break arrangement 5 which is fine. But it makes adding a new arrangement not needed.

twaugh: As long as we say in the release notes for osbs-client that it will require a newer version of atomic-reactor (i.e. one in wihch the plugin won't expect those parameters), then it's fine to stick with the existing plugin arrangement.

In any case, adding a new arrangement doesn't necessarily help the plugins to know whether to expect the parameters in question, as they don't necessarily know the arrangement version they are running for.

rcerven: ok removing 6, and modifying 5

requirements-py3.txt
```diff
@@ -1 +1,2 @@
 python-dateutil
+python3-PyYAML
```
twaugh: "python3-PyYAML" isn't a PyPI module name. Just PyYAML is fine, that works on both Python 2 and Python 3.

rcerven: ok deleting

osbs/repo_utils.py
```diff
@@ -52,6 +54,11 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
         if os.path.exists(config_path):
             self._config_parser.read(config_path)
 
+        file_path = os.path.join(dir_path, REPO_CONTAINER_CONFIG)
+        if os.path.exists(file_path):
+            with open(file_path) as f:
+                self.compose_data = (yaml.load(f) or {}).get('compose')
```
twaugh: These two lines don't get coverage during unit testing.

rcerven: ok will add

twaugh:
@mgoldmann Note that this PR introduces a tiny amount of parsing of container.yaml into osbs-client, and I expect this will be replaced by your work relating to tag name specification.

goldmann:
@twaugh Thanks!

@dbecvarik FYI

rcerven:
rn updated:

- Removed 'module' and 'module-compose-url'

PR #725, opened 2018-02-05
removal of unused annotations

rcerven:

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -227,21 +227,7 @@ def cmd_get_build(args, osbs):
         digests_str = make_digests_str(build.get_digests())
 
         logs_str = ''
-        build_logs = build.get_logs()
-        if build_logs:
-            logs_str = dedent("""\
-                BUILD LOGS
-
-                {logs}""").format(logs=build_logs)
-
         packages_str = ''
-        packages_list = build.get_rpm_packages()
-        if packages_list:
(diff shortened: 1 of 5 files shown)


osbs/cli/main.py
```diff
@@ -399,21 +385,13 @@ def cmd_build_logs(args, osbs):
     build_id = args.BUILD_ID[0]
     follow = args.follow
 
-    if follow and args.from_docker_build:
-        print("Can't use --follow and --from-docker-build. "
-              "Logs from docker build are part of metadata of a already built image.")
+    logs = osbs.get_build_logs(build_id, follow=follow,
+                               wait_if_missing=args.wait_if_missing,
+                               decode=True)
```
lcarva: Any concerns with using `decode=True` if follow if `False` ?

rcerven: ahh ok I'll put it bellow that if as it should be

rcerven: mmm this logic wasn't changed, should those 2 things be conflicting?

lcarva: Yes, you're right. Logic is the same.

I vaguely remember a while back there was an issue with follow and decode.

rcerven:
rn updated:

## API changes
- Removal of get_docker_build_logs method 

## Improvements
- Removal of methods related to logs and rpm-packages annotations

PR #726, opened 2018-02-06
unify build_image and build_imagestream by build_from

rcerven:

### DIFF
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -11,7 +11,7 @@ import logging
 import os
 import re
 import random
-from osbs.constants import DEFAULT_GIT_REF, DEFAULT_BUILD_IMAGE, RAND_DIGITS
+from osbs.constants import DEFAULT_GIT_REF, RAND_DIGITS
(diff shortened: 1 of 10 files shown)


osbs/build/spec.py
```diff
@@ -271,11 +272,21 @@ def set_params(self, git_uri=None, git_ref=None,
         self.yum_repourls.value = yum_repourls or []
         self.use_auth.value = use_auth
 
-        if build_imagestream and build_image:
+        unique_build_args = (build_imagestream, build_image, build_from)
+        if sum(bool(a) for a in unique_build_args) != 1:
```
twaugh: This might break existing configuration files which do not specify any of them and instead use the default value. Can we keep those working, perhaps by having a new default value "image:buildroot:latest" for build_from when no values are provided?

twaugh:
Robert reminded me that this work was intended to remove the default value as well, so now my only request is to add a warning.

rcerven:
warning added

lcarva:
Updated release notes draft:
> unify build_image and build_imagestream by build_from

PR #728, opened 2018-02-12
Added possibility to override image tags via container.yaml file.

dbecvarik:
If container.yaml file in the git repository contains tags section,
resulting image is created with these tags only.

Signed-off-by: David Becvarik <dbecvari@redhat.com>

As this is working, if the tags are missing image-version tag it will die in this [this assertion](https://github.com/projectatomic/atomic-reactor/blob/56299e323d9904248746282569a9c32f5353c547/atomic_reactor/plugins/exit_koji_import.py#L294).
I don't have a deep understanding how to solve this issue, because I'm missing bigger picture how whole OSBS/koji works inside. 
I've noticed that for scratch build this plugin is not used? 
Can we also disable it for this custom tags builds (but this feels very weird)?

@goldmann Or maybe, we can add this tag as always present ?

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -1072,6 +1072,8 @@ class BuildRequest(object):
             elif self.isolated:
                 tag_suffixes['primary'].extend(['{version}-{release}'])
             elif tags:
+                assert '{version}-{release}' in tags, \
+                    'Unable to find {version}-{release} in tags from container.yaml'
                 tag_suffixes['primary'].extend(tags)
             else:
                 tag_suffixes['primary'].extend(['latest', '{version}', '{version}-{release}'])
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -1060,25 +1060,14 @@ class BuildRequest(object):
         unique_tag = self.spec.image_tag.value.split(':')[-1]
         tag_suffixes = {'unique': [unique_tag], 'primary': []}
 
-        # this part is very ugly but otherwise ~900 tests will fail
-        # as self.repo_info object is not injected properly
-        tags = []
-        if self._repo_info:
-            tags = self._repo_info.configuration.container.get('tags', [])
-
         if self.spec.build_type.value == BUILD_TYPE_ORCHESTRATOR:
             if self.scratch:
                 pass
             elif self.isolated:
                 tag_suffixes['primary'].extend(['{version}-{release}'])
-            elif tags:
+            elif self._repo_info.additional_tags.from_container_yaml:
                 tag_suffixes['primary'].extend(['{version}-{release}'])
-                for tag in tags:
-                    assert '-' not in tag, \
-                        'Tag from container.yaml cannot contain "-" character. ' \
-                        'Provided tag was "%s"' % tag
-
-                tag_suffixes['primary'].extend(tags)
+                tag_suffixes['primary'].extend(self._repo_info.additional_tags.tags)
             else:
                 tag_suffixes['primary'].extend(['latest', '{version}', '{version}-{release}'])
                 tag_suffixes['primary'].extend(self._repo_info.additional_tags.tags)
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -1072,8 +1072,12 @@ class BuildRequest(object):
             elif self.isolated:
                 tag_suffixes['primary'].extend(['{version}-{release}'])
             elif tags:
-                assert '{version}-{release}' in tags, \
-                    'Unable to find {version}-{release} in tags from container.yaml'
+                tag_suffixes['primary'].extend(['{version}-{release}'])
+                for tag in tags:
+                    assert '-' not in tag, \
+                        'Tag from container.yaml cannot contain "-" character. ' \
+                        'Provided tag was "%s"' % tag
+
                 tag_suffixes['primary'].extend(tags)
             else:
                 tag_suffixes['primary'].extend(['latest', '{version}', '{version}-{release}'])
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -1060,11 +1060,19 @@ class BuildRequest(object):
         unique_tag = self.spec.image_tag.value.split(':')[-1]
         tag_suffixes = {'unique': [unique_tag], 'primary': []}
 
+        # this part is very ugly but otherwise ~900 tests will fail
+        # as self.repo_info object is not injected properly
+        tags = []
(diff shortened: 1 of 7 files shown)


dbecvarik:
Sample [container.yaml](https://github.com/dbecvarik/docker-hello-world/blob/osbs-box-demo/container.yaml) using tags.

@twaugh can you take a quick look if this way of implementation is correct one?

After speaking with @goldmann we can add {image}-{version} tag even if tags are specified in container.yaml. Is it ok to do it?

osbs/build/build_request.py
```diff
@@ -1060,11 +1060,19 @@ def render_tag_from_config(self):
         unique_tag = self.spec.image_tag.value.split(':')[-1]
         tag_suffixes = {'unique': [unique_tag], 'primary': []}
 
+        # this part is very ugly but otherwise ~900 tests will fail
+        # as self.repo_info object is not injected properly
```
twaugh: Maybe it just needs initialising to `RepoInfo()`?

csomh:
@dbecvarik I am not sure if my understanding is 100% correct (@twaugh might have a better explanation), but I *think*, the `{version}-{release}` tag is needed for manifests to have mapping between manifest lists and Koji builds (the later uses NVR as unique identifier for a build).

But: the pull spec in the build metadata will have a spec by digest, which should be enough for mapping.

Looping in @Allda, as this might impact tooling around RHCC.

goldmann:
@csomh We would like to minimize impact of this change. If it would be better to add the `{version}-{release}` tag for now and plan to make it optional later then that would be my preference since as you mentioned, removing this requirement would (possibly) impact other tooling.

Allda:
@csomh Yes, {version}-{release} tag is important for mapping between image manifest and image build in Koji.

twaugh:
From the point of view of OSBS, I don't think there is anything special about `{version}-{release}` in particular, only that it needs to report "a" tag (as well as the digest) in the build metadata for the `extra.image.index.pull` field.

About mapping: given a container image, we can map it back to the Koji build relating to it by examining the `com.redhat.component`, `version`, and `release` labels. You can do that even with an image not in a registry, so registry tags are not important for that.

goldmann:
@twaugh @Allda Can we agree how to proceed here? We want to wrap up this PR very soon so there will be time to include it in the release. We are now a bit worried that we will be short on time if there will be no decision soon.

Please let us know how would you like to proceed.

dbecvarik:
Ideally I would suggest to inject {version}-{release} tag for now and mark this feature as technology preview if possible to deliver it soon and we can clean it up later? Because the information about metadata mapping and registry is far behind my understanding what OSBS does and how to test it if everything is still correct.
So my suggestion is, that I'll finish schema checking for the container.yaml file and inject the {version}-{release} tag for, is it feasible @twaugh ?

twaugh:
How about leaving `{version}-{release}` up to the `tags` field in the YAML file, but for the moment saying that it **must** contain `{version}-{release}` (i.e. assert it)? That leaves us the option later of removing that requirement.

goldmann:
@twaugh I think this is great suggestion and I prefer this over anything else.

twaugh:
We may need to think about how/whether the import_image plugin should distinguish transient tags from non-transient tags. Currently it looks for "-" in the tag name. @lcarva

goldmann:
Anything we can do on our side? I'm still hoping for February release.

lcarva:
> We may need to think about how/whether the import_image plugin should distinguish transient tags from non-transient tags. Currently it looks for "-" in the tag name.

If we're allowing `'-'` in the override tags that's going to cause problems. Otherwise, `import_image` should be able to treat override tags as transient tags as expected.

twaugh:
> If we're allowing '-' in the override tags that's going to cause problems. Otherwise, import_image should be able to treat override tags as transient tags as expected.

So it should be this?: the list of tags must contain `"{version}-{release}"`, and no tag except that one may contain `"-"`

lcarva:
> So it should be this?: the list of tags must contain "{version}-{release}", and no tag except that one may contain "-"

That's the safest way to go about this IMO. Also, if we're going with that it should be done automatically by code; there's no need to make users add that entry themselves.

dbecvarik:
@lcarva  @twaugh What is the final decision? Which of these changes should be implemented?
* check if {version}-{release} tag is specified?
* auto inject {version}-{release} tag?
* check that provided tags doesn't contain '-' character?

twaugh:
@lcarva, the idea behind making users add `{version}-{release}` explicitly was to allow us an obvious way to relax the requirement of including that tag in future. (See discussion above.)

However, I don't have a strong preference either way, and if we want to do that sort of thing in the future without making users explicitly add `{version}-{release}`, we can always use an initial otherwise-invalid character to mean "not this tag", e.g. `-{version}-{release}`.

So my suggestion is:
* make `{version}-{release}` implicit
* disallow "-" in tag names provided via container.yaml

@lcarva, do we agree?

twaugh:
Are we moving the JSON Schema for container.yaml into osbs-client as part of this pull request?

dbecvarik:
@twaugh I would prefer not to implement schema now as its not strictly related to this PR. I can open different PR for schema checking this wee if we can get this in as its really high prio for us.

twaugh:
> I would prefer not to implement schema now as its not strictly related to this PR. I can open different PR for schema checking this wee if we can get this in as its really high prio for us.

Yes, ok.

dbecvarik:
@twaugh I;'ve updated the PR. So hope its ok this way.

osbs/build/build_request.py
twaugh: You don't think `BuildRequest.__init__()` ought to initialise it to `RepoInfo()`?

osbs/build/build_request.py
```diff
@@ -1060,11 +1060,25 @@ def render_tag_from_config(self):
...
+
         if self.spec.build_type.value == BUILD_TYPE_ORCHESTRATOR:
             if self.scratch:
                 pass
             elif self.isolated:
                 tag_suffixes['primary'].extend(['{version}-{release}'])
+            elif tags:
+                tag_suffixes['primary'].extend(['{version}-{release}'])
+                for tag in tags:
+                    assert '-' not in tag, \
```
lcarva: Validation should extend beyond this. `AdditionalTagsConfig` uses a regex to do so. Having this validation in two distinct places is error prone.

lcarva:
Updated release notes draft:
> - Added possibility to override image tags via `container.yaml` file

PR #729, opened 2018-02-14
import_image will be as well exit plugin

rcerven:

### DIFF
--- a/inputs/orchestrator_inner:5.json
+++ b/inputs/orchestrator_inner:5.json
@@ -127,16 +127,6 @@
         "dockpulp_loglevel": "INFO",
         "publish": false
       }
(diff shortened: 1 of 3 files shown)


twaugh:
I'd like to see the commit message explain a little more why we need to run this as an exit plugin (so it runs after pulp_publish), and why we still need to handle configuring it as a post-build plugin (for 'prod' builds).

rcerven:
Improved commit message

lcarva:
Once merged, https://github.com/projectatomic/osbs-client/pull/730 might have to be rebased.

rcerven:
rn updated : 
import_image is now also an ExitPlugin

PR #730, opened 2018-02-15
render_import_image: fix logic for checking imagestream_name/url

owtaylor:
The code was checking for spec.imagestream_name is None not
spec.imagestream_name.value is None - this was never noticed because
spec.imagestream_name is now always set except for Flatpak builds and
previously, a different check was removing the plugin first.

Check in tests that the plugin is properly removed for Flatpak builds.

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -1303,7 +1303,7 @@ class BuildRequest(object):
         plugin = 'import_image'
 
         for phase in phases:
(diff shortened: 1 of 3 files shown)


mlangsdorf:
This needs to be rebased before merging. It looks pretty straightforward but it's not obvious enough that I want to do the fixup myself.

twaugh:
Draft release notes updated.

PR #731, opened 2018-02-20
source_registry is string and not dict

rcerven:

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -1171,7 +1171,7 @@ class BuildRequest(object):
 
             source_registry = self.spec.source_registry_uri.value
             perform_delete = (source_registry is None or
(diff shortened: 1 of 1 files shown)


osbs/build/build_request.py
```diff
@@ -1171,7 +1171,7 @@ def render_pulp_sync(self):
 
             source_registry = self.spec.source_registry_uri.value
             perform_delete = (source_registry is None or
-                              source_registry.docker_uri != registry.docker_uri)
```
lcarva: It should be of `RegistryURI` type: https://github.com/projectatomic/osbs-client/blob/master/osbs/build/spec.py#L114

rcerven: yeah I've just find that out ..

PR #732, opened 2018-02-23
remove exit plugin image_image for scratch build

rcerven:


lkolacek:
:+1:

twaugh:
Can we also update tests/build_/test_arrangements.py so that arrangement 5 includes import_image in both the postbuild and exit phases (since valid version-5 arrangements can have it in either)?

That should catch this.

rcerven:
ok will update that test

lcarva:

PR #733, opened 2018-02-27
Add support for BuildUserParams and BuildRequestV2

mlangsdorf:
As part of the move to arrangement 6, create BuildRequestV2 and the
underlying BuildUserParams.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -21,6 +21,7 @@ from contextlib import contextmanager
 from types import GeneratorType
 
 from osbs.build.build_request import BuildRequest
+from osbs.build.build_requestv2 import BuildRequestV2
 from osbs.build.build_response import BuildResponse
 from osbs.build.pod_response import PodResponse
 from osbs.build.config_map_response import ConfigMapResponse
@@ -28,7 +29,8 @@ from osbs.constants import (BUILD_RUNNING_STATES, WORKER_OUTER_TEMPLATE,
                             WORKER_INNER_TEMPLATE, WORKER_CUSTOMIZE_CONF,
                             ORCHESTRATOR_OUTER_TEMPLATE, ORCHESTRATOR_INNER_TEMPLATE,
                             ORCHESTRATOR_CUSTOMIZE_CONF, BUILD_TYPE_WORKER,
-                            BUILD_TYPE_ORCHESTRATOR, BUILD_FINISHED_STATES)
+                            BUILD_TYPE_ORCHESTRATOR, BUILD_FINISHED_STATES,
+                            DEFAULT_ARRANGEMENT_VERSION, REACTOR_CONFIG_ARRANGEMENT_VERSION)
 from osbs.core import Openshift
 from osbs.exceptions import (OsbsException, OsbsValidationException, OsbsResponseException,
                              OsbsOrchestratorNotEnabled)
@@ -170,24 +172,32 @@ class OSBS(object):
 
     @osbsapi
     def get_build_request(self, build_type=None, inner_template=None,
-                          outer_template=None, customize_conf=None):
+                          outer_template=None, customize_conf=None,
+                          arrangement_version=DEFAULT_ARRANGEMENT_VERSION):
         """
-        return instance of BuildRequest
+        return instance of BuildRequest or BuildRequestV2
 
         :param build_type: str, unused
         :param inner_template: str, name of inner template for BuildRequest
         :param outer_template: str, name of outer template for BuildRequest
         :param customize_conf: str, name of customization config for BuildRequest
-        :return: instance of BuildRequest
+        :param arrangement_version: int, value of the arrangement version
+
+        :return: instance of BuildRequest or BuildRequestV2
         """
         if build_type is not None:
             warnings.warn("build types are deprecated, do not use the build_type argument")
 
-        build_request = BuildRequest(
-            build_json_store=self.os_conf.get_build_json_store(),
-            inner_template=inner_template,
-            outer_template=outer_template,
-            customize_conf=customize_conf)
+        if arrangement_version < REACTOR_CONFIG_ARRANGEMENT_VERSION:
+            build_request = BuildRequest(
+                build_json_store=self.os_conf.get_build_json_store(),
+                inner_template=inner_template,
+                outer_template=outer_template,
+                customize_conf=customize_conf)
+        else:
+            build_request = BuildRequestV2(
+                build_json_store=self.os_conf.get_build_json_store(),
+                customize_conf=customize_conf)
 
         # Apply configured resource limits.
         cpu_limit = self.build_conf.get_cpu_limit()
@@ -211,7 +221,7 @@ class OSBS(object):
         :return: instance of build.build_response.BuildResponse
         """
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
-        build = build_request.render()
+        build = build_request.render(self)
         response = self.os.create_build(json.dumps(build))
         build_response = BuildResponse(response.json())
         return build_response
@@ -306,7 +316,7 @@ class OSBS(object):
 
     def _create_build_directly(self, build_request, unique=None):
         logger.debug(build_request)
-        build_json = build_request.render()
+        build_json = build_request.render(self)
         build_json['kind'] = 'Build'
         build_json['spec']['serviceAccount'] = 'builder'
 
@@ -402,7 +412,7 @@ class OSBS(object):
         return existing_bc
 
     def _create_build_config_and_build(self, build_request):
-        build_json = build_request.render()
+        build_json = build_request.render(self)
         api_version = build_json['apiVersion']
         if api_version != self.os_conf.get_openshift_api_version():
             raise OsbsValidationException('BuildConfig template has incorrect apiVersion (%s)' %
--- a/osbs/build/build_requestv2.py
+++ b/osbs/build/build_requestv2.py
@@ -1,0 +1,225 @@
+"""
+Copyright (c) 2018 Red Hat, Inc
+All rights reserved.
+
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+from __future__ import print_function, absolute_import, unicode_literals
+
+import logging
+import os
+
+from osbs.build.build_request import BuildRequest
+from osbs.build.user_params import BuildUserParams
+from osbs.exceptions import OsbsValidationException
+from osbs.constants import SECRETS_PATH
+from osbs.utils import git_repo_humanish_part_from_uri
+
+logger = logging.getLogger(__name__)
+
+_CM_SECRETS = ['required_secrets', 'worker_token_secrets']
+
+
+class BuildRequestV2(BuildRequest):
+    """
+    Wraps logic for creating build inputs
+    """
+
+    def __init__(self, build_json_store, customize_conf=None):
+        """
+        :param build_json_store: str, path to directory with JSON build files
+        :param customize_conf: str, path to customize configuration JSON
+        """
+        super(BuildRequestV2, self).__init__(build_json_store=build_json_store,
+                                             customize_conf=customize_conf)
+        self.spec = None
+        self.user_params = BuildUserParams(build_json_store)
+
+    # Override
+    def set_params(self, **kwargs):
+        """
+        set parameters in the user parameters
+
+        these parameters are accepted:
+        :param git_uri: str, uri of the git repository for the source
+        :param git_ref: str, commit ID of the branch to be pulled
+        :param git_branch: str, branch name of the branch to be pulled
+        :param base_image: str, name of the parent image
+        :param name_label: str, label of the parent image
+        :param user: str, name of the user requesting the build
+        :param component: str, name of the component
+        :param release: str,
+        :param build_image: str,
+        :param build_imagestream: str,
+        :param build_from: str,
+        :param build_type: str, orchestrator or worker
+        :param platforms: list of str, platforms to build on
+        :param platform: str, platform
+        :param koji_target: str, koji tag with packages used to build the image
+        :param koji_task_id: str, koji ID
+        :param koji_parent_build: str,
+        :param flatpak: if we should build a Flatpak OCI Image
+        :param flatpak_base_image: str, name of the Flatpack OCI Image
+        :param reactor_config_map: str, name of the config map containing the reactor environment
+        :param yum_repourls: list of str, uris of the yum repos to pull from
+        :param signing_intent: bool, True to sign the resulting image
+        :param compose_ids: list of str,
+        :param filesystem_koji_task_id: int, Koji Task that created the base filesystem
+        :param platform_node_selector: dict, a nodeselector for a user_paramsific platform
+        :param scratch_build_node_selector: dict, a nodeselector for scratch builds
+        :param explicit_build_node_selector: dict, a nodeselector for explicit builds
+        :param auto_build_node_selector: dict, a nodeselector for auto builds
+        :param isolated_build_node_selector: dict, a nodeselector for isolated builds
+        :param additional_tag_data: dict, str of dir path and filename of additional tags
+                                    and set of str of tags
+        """
+
+        # Here we cater to the koji "scratch" build type, this will disable
+        # all plugins that might cause importing of data to koji
+        self.scratch = kwargs.get('scratch')
+        # When true, it indicates build was automatically started by
+        # OpenShift via a trigger, for instance ImageChangeTrigger
+        self.is_auto = kwargs.pop('is_auto', False)
+        # An isolated build is meant to patch a certain release and not
+        # update transient tags in container registry
+        self.isolated = kwargs.get('isolated')
+
+        self.validate_build_variation()
+
+        self.base_image = kwargs.get('base_image')
+        self.platform_node_selector = kwargs.get('platform_node_selector', {})
+        self.scratch_build_node_selector = kwargs.get('scratch_build_node_selector', {})
+        self.explicit_build_node_selector = kwargs.get('explicit_build_node_selector', {})
+        self.auto_build_node_selector = kwargs.get('auto_build_node_selector', {})
+        self.isolated_build_node_selector = kwargs.get('isolated_build_node_selector', {})
+
+        logger.debug("setting params '%s' for %s", kwargs, self.user_params)
+        self.user_params.set_params(**kwargs)
+
+    # Override
+    @property
+    def inner_template(self):
+        raise RuntimeError('inner_template not supported in BuildRequestV2')
+
+    # Override
+    @property
+    def customize_conf(self):
+        raise RuntimeError('customize_conf not supported in BuildRequestV2')
+
+    # Override
+    @property
+    def dj(self):
+        raise RuntimeError('DockJson not supported in BuildRequestV2')
+
+    def set_reactor_config_map(self, api):
+        if not self.user_params.reactor_config_map.value:
+            return
+        super(BuildRequestV2, self).set_reactor_config_map(self.user_params.reactor_config_map)
+        cm_response = api.get_config_map(self.user_params.reactor_config_map.value)
+        custom = self.template['spec']['strategy']['customStrategy']
+        custom.setdefault('secrets', [])
+        existing = [secret_mount['secretSource']['name'] for secret_mount in custom['secrets']]
+        for key in _CM_SECRETS:
+            secrets = cm_response.get_data_by_key(key)
+            for secret in secrets:
+                if secret in existing:
+                    continue
+                else:
+                    custom['secrets'].append({
+                        'secretSource': {
+                            'name': secret,
+                        },
+                        'mountPath': os.path.join(SECRETS_PATH, secret),
+                    })
+
+    def adjust_for_scratch(self):
+        """
+        Scratch builds must not affect subsequent builds,
+        and should not be imported into Koji.
+        """
+        if self.scratch:
+            self.template['spec'].pop('triggers', None)
+            self.set_label('scratch', 'true')
+
+    def render_node_selectors(self):
+        # for worker builds set nodeselectors
+        if self.user_params.platforms.value is None:
+            # auto or explicit build selector
+            if self.is_auto:
+                self.template['spec']['nodeSelector'] = self.auto_build_node_selector
+            # scratch build nodeselector
+            elif self.scratch:
+                self.template['spec']['nodeSelector'] = self.scratch_build_node_selector
+            # isolated build nodeselector
+            elif self.isolated:
+                self.template['spec']['nodeSelector'] = self.isolated_build_node_selector
+            # explicit build nodeselector
+            else:
+                self.template['spec']['nodeSelector'] = self.explicit_build_node_selector
+
+            # platform nodeselector
+            if self.platform_node_selector:
+                self.template['spec']['nodeSelector'].update(self.platform_node_selector)
+
+    def render(self, api=None, validate=True):
+        # the api is required for BuildRequestV2
+        # can't check that its an OSBS object because of the circular import
+        if not api:
+            raise OsbsValidationException
+
+        # Validate BuildUserParams
+        if validate:
+            self.user_params.validate()
+
+        self.render_name(self.user_params.name, self.user_params.image_tag,
+                         self.user_params.platform)
+        self.render_resource_limits()
+
+        self.template['spec']['source']['git']['uri'] = self.user_params.git_uri.value
+        self.template['spec']['source']['git']['ref'] = self.user_params.git_ref.value
+
+        self.template['spec']['output']['to']['name'] = self.user_params.image_tag.value
+
+        if self.has_ist_trigger():
+            imagechange = self.template['spec']['triggers'][0]['imageChange']
+            imagechange['from']['name'] = self.user_params.trigger_imagestreamtag.value
+
+        custom_strategy = self.template['spec']['strategy']['customStrategy']
+        if self.user_params.build_imagestream.value:
+            custom_strategy['from']['kind'] = 'ImageStreamTag'
+            custom_strategy['from']['name'] = self.user_params.build_imagestream.value
+        else:
+            custom_strategy['from']['name'] = self.user_params.build_image.value
+
+        # Set git-repo-name label
+        # NOTE: Since only the repo name is used, a forked repos will have
+        # the same git-repo-name tag. This is a known limitation. If this
+        # use case must be handled properly, the git URI must be taken into
+        # account.
+        repo_name = git_repo_humanish_part_from_uri(self.user_params.git_uri.value)
+        self.set_label('git-repo-name', repo_name)
+        self.set_label('git-branch', self.user_params.git_branch.value)
+        koji_task_id = self.user_params.koji_task_id.value
+        if koji_task_id is not None:
+            self.set_label('koji-task-id', str(koji_task_id))
+
+        # Set template.spec.strategy.customStrategy.env[] USER_PARAMS
+        # Set required_secrets based on reactor_config
+        # Set worker_token_secrets based on reactor_config, if any
+        self.set_reactor_config_map(api)
+
+        # Adjust triggers for custom base image
+        if self.template['spec'].get('triggers', []) and self.is_custom_base_image():
+            del self.template['spec']['triggers']
+
+        self.adjust_for_repo_info()
+        self.adjust_for_scratch()
+        self.adjust_for_isolated(self.user_params.release)
+        self.render_node_selectors()
+
+        # Log build json
+        # Return build json
+        self.build_json = self.template
+        logger.debug(self.build_json)
+        return self.build_json
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -114,10 +114,53 @@ class SourceRegistryURIParam(BuildParam):
         BuildParam.value.fset(self, RegistryURI(val) if val else None)
 
 
-class BuildSpec(object):
-
+class BuildCommon(object):
     def __init__(self):
+        self.image_tag = BuildParam("image_tag")
+        self.koji_target = BuildParam("koji_target", allow_none=True)
+        self.platform = BuildParam("platform", allow_none=True)
+        self.arrangement_version = BuildParam("arrangement_version", allow_none=True)
+        self.filesystem_koji_task_id = BuildParam("filesystem_koji_task_id", allow_none=True)
+        self.required_params = [
+            self.koji_target,
+        ]
 
+    def _populate_image_tag(self):
+        timestamp = utcnow().strftime('%Y%m%d%H%M%S')
+        # RNG is seeded once its imported, so in cli calls scratch builds would get unique name.
+        # On brew builders we import osbs once - thus RNG is seeded once and `randrange`
+        # returns the same values throughout the life of the builder.
+        # Before each `randrange` call we should be calling `.seed` to prevent this
+        random.seed()
+
+        tag_segments = [
+            self.koji_target.value or 'none',
+            str(random.randrange(10**(RAND_DIGITS - 1), 10**RAND_DIGITS)),
+            timestamp
+        ]
+
+        if self.platform.value and (self.arrangement_version.value or 0) >= 4:
+            tag_segments.append(self.platform.value)
+
+        tag = '-'.join(tag_segments)
+        self.image_tag.value = '{0}/{1}:{2}'.format(self.user.value, self.component.value, tag)
+
+    def validate(self):
+        logger.info("Validating params of %s", self.__class__.__name__)
+        for param in self.required_params:
+            if param.value is None:
+                if param.allow_none:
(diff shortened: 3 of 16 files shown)


osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,395 @@
...
+        self.platform_node_selector = None
+
+    # Override
+    def set_params(self, **kwargs):
+        """
+        set parameters according to specification
+
+        these parameters are accepted:
+
+        :param pulp_secret: str, resource name of pulp secret
```
lcarva: Let's update this comment with the user params/node_selectors we actually expect.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,395 @@
...
+        self.platform_node_selector = kwargs.get('platform_node_selector', {})
+        self.scratch_build_node_selector = kwargs.get('scratch_build_node_selector', {})
+        self.explicit_build_node_selector = kwargs.get('explicit_build_node_selector', {})
+        self.auto_build_node_selector = kwargs.get('auto_build_node_selector', {})
+        self.isolated_build_node_selector = kwargs.get('isolated_build_node_selector', {})
+
+        logger.debug("setting params '%s' for %s", kwargs, self.user_params)
+        self.user_params.set_params(**kwargs)
+
+    def validate_build_variation(self):
```
lcarva: No need to override this method.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,395 @@
...
+        variations = (self.scratch, self.is_auto, self.isolated)
+        if variations.count(True) > 1:
+            raise OsbsValidationException(
+                'Build variations are mutually exclusive. '
+                'Must set either scratch, is_auto, isolated, or none. ')
+
+    def set_resource_limits(self, cpu=None, memory=None, storage=None):
```
lcarva: No need to override this method.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,395 @@
...
+        if cpu is not None:
+            self._resource_limits['cpu'] = cpu
+
+        if memory is not None:
+            self._resource_limits['memory'] = memory
+
+        if storage is not None:
+            self._resource_limits['storage'] = storage
+
+    def set_repo_info(self, repo_info):
```
lcarva: No need to override this method.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,395 @@
...
+        self._repo_info = repo_info
+
+    @property
+    def build_id(self):
```
lcarva: No need to override this method.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,395 @@
...
+        return self.build_json['metadata']['name']
+
+    @property
+    def template(self):
```
lcarva: No need to override this method.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,395 @@
...
+    @property
+    def customize_conf(self):
+        raise RuntimeError('customize_conf not supported in BuildRequestV2')
+
+    # Override
+    @property
+    def dj(self):
+        raise RuntimeError('DockJson not supported in BuildRequestV2')
+
+    def has_ist_trigger(self):
```
lcarva: No need to override this method.

I won't comment on others, but if the method is the same as in parent class we don't need to override it.

osbs/api.py
```diff
@@ -170,24 +171,34 @@ def get_pod_for_build(self, build_id):
...
         """
-        return instance of BuildRequest
+        return instance of BuildRequest or BuildRequestV2
 
         :param build_type: str, unused
         :param inner_template: str, name of inner template for BuildRequest
         :param outer_template: str, name of outer template for BuildRequest
         :param customize_conf: str, name of customization config for BuildRequest
-        :return: instance of BuildRequest
+        :parram arrangement_version: int, value of the arrangement version
```
twaugh: Typo, should be "param".

osbs/api.py
```diff
@@ -170,24 +171,34 @@ def get_pod_for_build(self, build_id):
 
     @osbsapi
     def get_build_request(self, build_type=None, inner_template=None,
-                          outer_template=None, customize_conf=None):
+                          outer_template=None, customize_conf=None,
+                          arrangement_version=5):
```
twaugh: Should this be `DEFAULT_ARRANGEMENT_VERSION`?

osbs/api.py
```diff
@@ -170,24 +171,34 @@ def get_pod_for_build(self, build_id):
...
         """
-        return instance of BuildRequest
+        return instance of BuildRequest or BuildRequestV2
```
twaugh: Technically a `BuildRequestV2` *is* a `BuildRequest` :-) but yes, this probably makes it clearer.

osbs/api.py
```diff
@@ -170,24 +171,34 @@ def get_pod_for_build(self, build_id):
...
+            build_request = BuildRequest(
+                build_json_store=self.os_conf.get_build_json_store(),
+                inner_template=inner_template,
+                outer_template=outer_template,
+                customize_conf=customize_conf)
+        else:
+            build_request = BuildRequestV2(
+                build_json_store=self.os_conf.get_build_json_store(),
+                inner_template=inner_template,
+                outer_template=outer_template,
```
twaugh: Remove `inner_template` and `outer_template` parameters now they're no longer accepted?

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,292 @@
...
+    """
+    Wraps logic for creating build inputs
+    """
+
+    def __init__(self, build_json_store, customize_conf=None):
+        """
+        :param build_json_store: str, path to directory with JSON build files
+        :param customize_conf: str, path to customize configuration JSON
+        """
+        super(BuildRequestV2, self).__init__(build_json_store, customize_conf)
```
twaugh: Does that need to be `customize_conf=customize_conf`? Otherwise it might be taken for `inner_template` due to being passed as a positional argument.

Come to think of it, I don't even know if passing `customize_conf` through the the superclass makes sense. @lcarva?

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,292 @@
...
+                    'name': self.user_params.reactor_config_map.value,
+                    'key': 'config.yaml'
+                }
+            }
+        }
+        custom['env'].append(reactor_config_map)
+        if api:
+            cm_response = api.get_config_map(self.user_params.reactor_config_map.value)
+            for key in _CM_SECRETS:
+                self.template['spec'][key] = cm_response.get_data_by_key(key)
```
twaugh: Something looks wrong here. `'required_secrets'` (for example) is not something an OpenShift v1.BuildRequest knows about. Here is how to add secrets to a v1.BuildRequest: https://github.com/projectatomic/osbs-client/blob/master/osbs/build/build_request.py#L412-L417

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,292 @@
...
+            'name': 'REACTOR_CONFIG',
+            'valueFrom': {
+                'configMapKeyRef': {
+                    'name': self.user_params.reactor_config_map.value,
+                    'key': 'config.yaml'
+                }
+            }
+        }
+        custom['env'].append(reactor_config_map)
+        if api:
```
twaugh: It's a required parameter. Why not just assume it's valid and start to use it? The caller will (correctly) get a traceback if they passed the wrong thing.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,292 @@
...
+            raise OsbsValidationException(
+                'For isolated builds, the release value must be in the format: {0}'
+                .format(ISOLATED_RELEASE_FORMAT.pattern))
+
+        self.set_label('isolated', 'true')
+        self.set_label('isolated-release', self.user_params.release.value)
+
+    def adjust_for_repo_info(self):
+        if not self._repo_info:
+            logger.warning('repo info not set')
```
twaugh: Does this indicate a bug? I feel like a warning isn't severe enough.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,292 @@
...
+            return
+
+        if not self._repo_info.configuration.is_autorebuild_enabled():
```
twaugh: `if not x: ... else: ...` is extra mental work to parse. Can we switch it around so it's `if x: ... else: ...`?

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,292 @@
...
+                self.template['spec']['nodeSelector'] = self.isolated_build_node_selector
+            # explicit build nodeselector
+            else:
+                self.template['spec']['nodeSelector'] = self.explicit_build_node_selector
+
+            # platform nodeselector
+            if self.platform_node_selector:
+                self.template['spec']['nodeSelector'].update(self.platform_node_selector)
+
+    def render(self, api=None, validate=True):
```
twaugh: Should `api` be a required parameter?

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,292 @@
...
+            custom_strategy['from']['name'] = self.user_params.build_image.value
+
+        # Set git-repo-name label
+        # NOTE: Since only the repo name is used, a forked repos will have
+        # the same git-repo-name tag. This is a known limitation. If this
+        # use case must be handled properly, the git URI must be taken into
+        # account.
+        repo_name = git_repo_humanish_part_from_uri(self.user_params.git_uri.value)
+        self.set_label('git-repo-name', repo_name)
+        # Set git-branch label  #DONE#
```
twaugh: Extra bits in the comment

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,292 @@
...
+        self.set_label('git-branch', self.user_params.git_branch.value)
+        # Set koji-task-id label  #DONE#
```
twaugh: Extra bits in the comment

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,292 @@
...
+        self.set_reactor_config_map(api)
+
+        # Adjust for triggers
+        # Adjust for custom base image
+        if self.template['spec'].get('triggers', []) and self.is_custom_base_image():
+            del self.template['spec']['triggers']
+
+        # Adjust for repo info
+        # Adjust for scratch build
+        # Adjust for isolated build
```
twaugh: These could be removed

osbs/build/build_requestv2.py

osbs/build/build_requestv2.py

osbs/build/build_requestv2.py
mlangsdorf: Technically yes, but making it a required parameter messes up the mock'd lambdas in test_api and I would rather not unsort that mess.

osbs/build/build_requestv2.py
mlangsdorf: Fixed

osbs/api.py
mlangsdorf: Fixed

osbs/build/build_requestv2.py
mlangsdorf: Fixed but I'll amend it if lcarva thinks we don't need customize_conf.

osbs/api.py

osbs/api.py
mlangsdorf: Fixed

osbs/build/build_requestv2.py
twaugh: Maybe we need to raise OsbsValidationError first of all if it is None, in that case.

osbs/build/spec.py
```diff
@@ -431,3 +431,154 @@ def validate(self):
...
         return "Spec(%s)" % self.__dict__
+
+
+class BuildUserParams(object):
+    def __init__(self):
+        self.git_uri = BuildParam('git_uri')
+        self.git_ref = BuildParam('git_ref', default=DEFAULT_GIT_REF)
+        self.git_branch = BuildParam('git_branch')
+        self.user = UserParam()
+        self.use_auth = BuildParam("use_auth", allow_none=True)
```
twaugh: `use_auth` doesn't seem like a user parameter. I'd expect this to be an environment setting.

osbs/build/spec.py
```diff
@@ -431,3 +431,154 @@ def validate(self):
...
+        self.platform = BuildParam("platform", allow_none=True)
+        self.component = BuildParam('component')
+        self.name = BuildIDParam()
+        self.build_image = BuildParam('build_image')
+        self.build_imagestream = BuildParam('build_imagestream')
+        self.build_from = BuildParam('build_from')
+        self.proxy = BuildParam("proxy", allow_none=True)
```
twaugh: `proxy` seems like an environment setting.

osbs/build/spec.py
```diff
@@ -431,3 +431,154 @@ def validate(self):
...
+        self.git_ref = BuildParam('git_ref', default=DEFAULT_GIT_REF)
+        self.git_branch = BuildParam('git_branch')
+        self.user = UserParam()
+        self.use_auth = BuildParam("use_auth", allow_none=True)
+        self.platform = BuildParam("platform", allow_none=True)
+        self.component = BuildParam('component')
+        self.name = BuildIDParam()
+        self.build_image = BuildParam('build_image')
+        self.build_imagestream = BuildParam('build_imagestream')
+        self.build_from = BuildParam('build_from')
```
twaugh: I can't work out if these three (build_*) are environment settings or user parameters.

osbs/build/spec.py
```diff
@@ -431,3 +431,154 @@ def validate(self):
...
+        self.proxy = BuildParam("proxy", allow_none=True)
+        self.trigger_imagestreamtag = BuildParam('trigger_imagestreamtag')
+        self.imagestream_name = BuildParam('imagestream_name')
+        self.imagestream_url = BuildParam('imagestream_url')
+        self.imagestream_insecure_registry = BuildParam('imagestream_insecure_registry')
+        self.architecture = BuildParam("architecture")
```
twaugh: Is `architecture` used?

osbs/build/spec.py
```diff
@@ -431,3 +431,154 @@ def validate(self):
...
+        self.koji_target = BuildParam("koji_target", allow_none=True)
+        self.koji_task_id = BuildParam("koji_task_id", allow_none=True)
+        self.flatpak = BuildParam("flatpak", default=False)
+        self.flatpak_base_image = BuildParam("flatpak_base_image", allow_none=True)
+        self.image_tag = BuildParam("image_tag")
+        self.signing_intent = BuildParam("signing_intent", allow_none=True)
+        self.yum_repourls = BuildParam("yum_repourls")
+        self.compose_ids = BuildParam("compose_ids", allow_none=True)
+        self.reactor_config_map = BuildParam("reactor_config_map", allow_none=True)
+        self.reactor_config_secret = BuildParam("reactor_config_secret", allow_none=True)
```
twaugh: `reactor_config_*` seem more like environment parameters.

osbs/build/build_requestv2.py

osbs/build/spec.py
mlangsdorf: Fixed.

osbs/build/spec.py
mlangsdorf: Fixed.

osbs/build/spec.py
mlangsdorf: I hope you're not expecting me to figure that out, then.

osbs/build/spec.py
mlangsdorf: No. Removed.

osbs/build/spec.py
mlangsdorf: reactor_config_map is the name of the reactor configuration map. it has to be a user parameter because it's the key needed to retrieve the environment parameters.

reactor_config_secret removed.

osbs/build/spec.py
twaugh: Looks fine.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,278 @@
...
+        reactor_config_map = {
+            'name': 'REACTOR_CONFIG',
+            'valueFrom': {
+                'configMapKeyRef': {
+                    'name': self.user_params.reactor_config_map.value,
+                    'key': 'config.yaml'
+                }
+            }
+        }
+        custom['env'].append(reactor_config_map)
```
lcarva: Can we call `super` to avoid duplication?

osbs/build/build_requestv2.py
lcarva: This was probably added to avoid massive changes to unit tests...
Maybe a technical debt cleanup task?

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,278 @@
...
+
+        if not ISOLATED_RELEASE_FORMAT.match(self.user_params.release.value):
+            raise OsbsValidationException(
+                'For isolated builds, the release value must be in the format: {0}'
+                .format(ISOLATED_RELEASE_FORMAT.pattern))
+
+        self.set_label('isolated', 'true')
+        self.set_label('isolated-release', self.user_params.release.value)
+
+    def adjust_for_repo_info(self):
```
lcarva: Do we need to override this method? The one from `BuildRequest` looks functionally the same.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,278 @@
...
+                # As expected, release label not set in Dockerfile
+                pass
+            else:
+                raise RuntimeError('when autorebuild is enabled in repo configuration, '
+                                   '"release" label must not be set in Dockerfile')
+        else:
+            logger.info('autorebuild is disabled in repo configuration, removing triggers')
+            self.template['spec'].pop('triggers', None)
+
+    def render_name(self):
```
lcarva: Looks the same as in `BuildRequest`, no need to override?

lcarva: oh `user_params` vs `spec`. Hmm I wonder if there's a cleaner way to do this and avoid duplication.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,278 @@
...
+        :param platform: str, platform
+        :param koji_target: str, koji tag with packages used to build the image
+        :param koji_task_id: str, koji ID
+        :param flatpak: if we should build a Flatpak OCI Image
+        :param flatpak_base_image: str, name of the Flatpack OCI Image
+        :param reactor_config_map: str, name of the config map containing the reactor environment
+        :param yum_repourls: list of str, uris of the yum repos to pull from
+        :param signing_intent: bool, True to sign the resulting image
+        :param compose_ids: list of str,
+        :param filesystem_koji_task_id: int, Koji Task that created the base filesystem
```
twaugh: This parameter seems to be ignored by BuildUserParams. @lcarva? Should it be removed from here, or should BuildUserParams do something with it?

osbs/build/build_requestv2.py
twaugh: Uses `self.user_params` instead of `self.spec`.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,278 @@
...
+                'configMapKeyRef': {
+                    'name': self.user_params.reactor_config_map.value,
+                    'key': 'config.yaml'
+                }
+            }
+        }
+        custom['env'].append(reactor_config_map)
+        cm_response = api.get_config_map(self.user_params.reactor_config_map.value)
+        for key in _CM_SECRETS:
+            self.template['spec'][key] = cm_response.get_data_by_key(key)
```
twaugh: This still looks wrong. See https://github.com/projectatomic/osbs-client/blob/0.45/osbs/build/build_request.py#L409-L420

osbs/build/build_requestv2.py

osbs/build/build_requestv2.py

osbs/build/build_requestv2.py
mlangsdorf: Fixed by paramaterizing the parent function and removing this one.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,207 @@
...
+        :param koji_task_id: str, koji ID
+        :param flatpak: if we should build a Flatpak OCI Image
+        :param flatpak_base_image: str, name of the Flatpack OCI Image
+        :param reactor_config_map: str, name of the config map containing the reactor environment
+        :param yum_repourls: list of str, uris of the yum repos to pull from
+        :param signing_intent: bool, True to sign the resulting image
+        :param compose_ids: list of str,
+        :param filesystem_koji_task_id: int, Koji Task that created the base filesystem
+        :param platform_node_selector: dict, a nodeselector for a user_paramsific platform
+        :param platform_descriptors: dict, platforms and their archiectures and enable_v1 settings
```
twaugh: Seems like platform_descriptors isn't used here or in BuildUserParams.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,207 @@
...
+    def dj(self):
+        raise RuntimeError('DockJson not supported in BuildRequestV2')
+
+    def set_reactor_config_map(self, api):
+        if not self.user_params.reactor_config_map.value:
+            return
+        super(BuildRequestV2, self).set_reactor_config_map(self.user_params.reactor_config_map)
+        cm_response = api.get_config_map(self.user_params.reactor_config_map.value)
+        for key in _CM_SECRETS:
+            self.template['spec'][key] = cm_response.get_data_by_key(key)
```
twaugh: This is still wrong. See set_secret_for_plugin for how to add secrets to a BuildConfig.

osbs/build/spec.py
```diff
@@ -431,3 +431,144 @@ def validate(self):
...
+            timestamp
+        ]
+
+        if self.platform.value:
+            tag_segments.append(self.platform.value)
+
+        tag = '-'.join(tag_segments)
+        self.image_tag.value = '{0}/{1}:{2}'.format(self.user.value, self.component.value, tag)
+
+    def validate(self):
```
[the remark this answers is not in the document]
mlangsdorf: It looks like it would simpler to make BuildSpec inherit from BuildUserParams, but I'm not sure about the conceptual logic.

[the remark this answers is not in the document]
mlangsdorf: Fixed by creating a BuildCommon class.

osbs/build/spec.py
```diff
@@ -431,3 +431,144 @@ def validate(self):
...
+            raise OsbsValidationException("compose_ids must be a list")
+        if not (yum_repourls is None or isinstance(yum_repourls, list)):
+            raise OsbsValidationException("yum_repourls must be a list")
+        self.yum_repourls.value = yum_repourls or []
+        self.signing_intent.value = signing_intent
+        self.compose_ids.value = compose_ids or []
+
+        self._populate_image_tag()
+
+    def _populate_image_tag(self):
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

osbs/build/spec.py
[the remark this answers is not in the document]
mlangsdorf: Fixed by creating a BuildCommon class.

osbs/build/build_requestv2.py

osbs/build/build_requestv2.py
mlangsdorf: Fixed I think.

mlangsdorf:
Coveralls seems to be broken again on Tuesday afternoon; don't know why.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,223 @@
...
+        raise RuntimeError('DockJson not supported in BuildRequestV2')
+
+    def set_reactor_config_map(self, api):
+        if not self.user_params.reactor_config_map.value:
+            return
+        super(BuildRequestV2, self).set_reactor_config_map(self.user_params.reactor_config_map)
+        cm_response = api.get_config_map(self.user_params.reactor_config_map.value)
+        custom = self.template['spec']['strategy']['customStrategy']
+        if 'secrets' not in custom:
+            custom['secrets'] = []
```
twaugh: Minor: can replace these two lines with `custom.setdefault('secrets', [])`

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,223 @@
...
+        for key in _CM_SECRETS:
+            secrets = cm_response.get_data_by_key(key)
+            for secret in secrets:
+                existing = [secret_mount for secret_mount in custom['secrets']
+                            if secret_mount['secretSource']['name'] == secret]
```
twaugh: Minor: could do this outside the loop instead for more efficient look-ups with `in`:
```
existing = [secret_mount['secretSource']['name'] for secret_mount in custom['secrets']]
for key in _CM_SECRETS:
    secrets = ...
    for secret in secrets:
        if secret in existing:
            continue
        ...
```

tests/build_/test_build_requestv2.py
```diff
@@ -0,0 +1,699 @@
...
+        'name_label': 'fedora/resultingimage',
+        'koji_target': 'koji-target',
+        'sources_command': 'make',
+        'platforms': ['x86_64'],
+        'filesystem_koji_task_id': TEST_FILESYSTEM_KOJI_TASK_ID,
+        'build_from': 'image:buildroot:latest',
+    }
+
+
+def get_secret_mountpath_by_name(build_json, name):
```
twaugh: Is this used?

tests/build_/test_build_requestv2.py
```diff
@@ -0,0 +1,699 @@
...
+    secret = named_secrets[0]
+    assert 'mountPath' in secret
+    return secret['mountPath']
+
+
+class TestBuildRequestV2(object):
+    def test_inner_template(self):
+        br = BuildRequestV2('something')
+        with pytest.raises(RuntimeError):
+            assert br.inner_template
```
twaugh: `assert br.inner_template` there is slightly confusing, because at first glance it looks like the test thinks it should be truthy. Maybe just `br.inner_template`? Likewise below.

tests/build_/test_build_requestv2.py
```diff
@@ -0,0 +1,699 @@
...
+                    'key': 'config.yaml'
+                }
+
+            assert 'REACTOR_CONFIG' in envs
+            assert 'configMapKeyRef' in envs['REACTOR_CONFIG']
+            assert envs['REACTOR_CONFIG']['configMapKeyRef'] == configmapkeyref
+            custom_secrets = build_json['spec']['strategy']['customStrategy']['secrets']
+            all_secrets = ['Nothing', 'Nada', 'Secret']
+            for secret in custom_secrets:
+                assert secret['secretSource']['name'] in all_secrets
```
twaugh: Do we also want to check that each of `all_secrets` is set in `custom_secrets`? Could do set comparison for this:
```
all_secrets = set(['Nothing', 'Nada', 'Secret'])  # use py2 syntax for compatibility
configured_secrets = set(secret['secretSource']['name'] for secret in custom_secrets)
assert all_secrets == configured_secrets
```

osbs/build/build_requestv2.py
lcarva: Yes, it should be in `BuildUserParams` because it needs to be stored directly in the `BuildConfig`.
`BuildRequestV2` should forward it along to `add_filesystem` plugin just like `BuildRequest` does.

tests/build_/test_build_requestv2.py
mlangsdorf: Thank you! I wanted to do this check, couldn't figure out how to do it gracefully, and got confused and decided it wasn't necessary.

Fixed.

tests/build_/test_build_requestv2.py
mlangsdorf: No. Removed.

osbs/build/build_requestv2.py

osbs/build/build_requestv2.py

osbs/build/spec.py
```diff
@@ -397,37 +435,116 @@ def set_params(self, git_uri=None, git_ref=None,
...
+                   git_uri=None, git_ref=None, git_branch=None,
+                   base_image=None, name_label=None,
+                   user=None,
+                   component=None, release=None,
+                   build_image=None, build_imagestream=None, build_from=None,
+                   platforms=None, platform=None,
+                   koji_target=None, koji_task_id=None,
+                   flatpak=None, flatpak_base_image=None,
+                   reactor_config_map=None,
+                   yum_repourls=None, signing_intent=None, compose_ids=None,
```
twaugh: Needs to accept `filesystem_koji_task_id` as well and store this is `self.filesystem_koji_task_id.value`, and of course `__init__` needs to create `self.filesystem_koji_task_id = BuildParam(...)` like `BuildSpec.__init__` does. Then `BuildRequestV2` needs to pass that value through in its call to this method.

Also do we have anything testing that the value ends up correctly in the add_filesystem plugin?

tests/build_/test_spec.py
```diff
@@ -172,3 +172,105 @@ def test_compose_ids_and_signing_intent(self, odcs_enabled, signing_intent, comp
                 assert spec.signing_intent.value == signing_intent
             if compose_ids:
                 assert spec.compose_ids.value == compose_ids
+
+
+class TestBuildUserParams(object):
+    def get_minimal_kwargs(self):
+        return {
+            # Params needed to avoid exceptions.
+            # Params needed to avoid exceptions.
```
twaugh: Duplicate comment.

osbs/build/build_requestv2.py
mlangsdorf: BuildRequestV2 doesn't forward anything to plugins. That's the entire point of BuildRequestV2.

osbs/build/build_requestv2.py

tests/build_/test_build_requestv2.py
```diff
@@ -0,0 +1,690 @@
...
+
+            assert 'REACTOR_CONFIG' in envs
+            assert 'configMapKeyRef' in envs['REACTOR_CONFIG']
+            assert envs['REACTOR_CONFIG']['configMapKeyRef'] == configmapkeyref
+            custom_secrets = build_json['spec']['strategy']['customStrategy']['secrets']
+            all_secrets = set(['Nothing', 'Nada', 'Secret'])  # use py2 syntax for compatibility
+            configured_secrets = set(secret['secretSource']['name'] for secret in custom_secrets)
+            assert all_secrets == configured_secrets
+            for secret in custom_secrets:
+                assert secret['secretSource']['name'] in all_secrets
```
twaugh: These last two lines are redundant because we know that `all_secrets == configured_secrets`.

tests/build_/test_build_requestv2.py
mlangsdorf: Fixed.

tests/build_/test_spec.py
mlangsdorf: Fixed.

osbs/build/spec.py
mlangsdorf: Fixed.

tests/build_/test_build_requestv2.py
mlangsdorf: Fixed.

osbs/build/build_requestv2.py
rcerven: in last part of reactor config map story, I will pass self.os, to set_params, and it will be available in BuildRequest as self.os

rcerven: so won't need to pass it to render() but you can just access it

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,223 @@
...
+    # Override
+    @property
+    def dj(self):
+        raise RuntimeError('DockJson not supported in BuildRequestV2')
+
+    def set_reactor_config_map(self, api):
+        if not self.user_params.reactor_config_map.value:
+            return
+        super(BuildRequestV2, self).set_reactor_config_map(self.user_params.reactor_config_map)
+        cm_response = api.get_config_map(self.user_params.reactor_config_map.value)
```
rcerven: this should be api.os.get_config_map

since you are passing self, and not self.os  to render()

osbs/build/build_requestv2.py
mlangsdorf: okay. I'll keep the scaffolding in for now and fix it on the merge.

osbs/build/build_requestv2.py
rcerven: never mind, I've overlooked that api is having method for that :)

osbs/build/build_requestv2.py
lcarva: My bad, `BuildUserParams` should store it so it can be used later when rendering plugins configuration via `PluginsConfigurationRender`.

mlangsdorf:
github is stalling out on refreshing this PR 3 times out of 4; closing it and opening a new PR 
https://github.com/projectatomic/osbs-client/pull/738 in hope that will fix the problem.

PR #734, opened 2018-02-28
import image exit isn't required

rcerven:


mlangsdorf:
added to rn
- and is an optional plugin
to # Improvements @ import_image is now also an ExitPlugin

PR #737, opened 2018-03-08
reactor config env for worker builds

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -535,6 +535,7 @@ class OSBS(object):
                               flatpak=False,
                               signing_intent=None,
                               compose_ids=None,
+                              reactor_config_override=None,
                               **kwargs):
 
         if flatpak:
@@ -617,6 +618,7 @@ class OSBS(object):
             scratch=self.build_conf.get_scratch(scratch),
             reactor_config_secret=self.build_conf.get_reactor_config_secret(),
             reactor_config_map=self.build_conf.get_reactor_config_map(),
+            reactor_config_override=reactor_config_override,
             client_config_secret=self.build_conf.get_client_config_secret(),
             token_secrets=self.build_conf.get_token_secrets(),
             arrangement_version=arrangement_version,
@@ -637,7 +639,8 @@ class OSBS(object):
             isolated=isolated,
             prefer_schema1_digest=self.build_conf.get_prefer_schema1_digest(),
             signing_intent=signing_intent,
-            compose_ids=compose_ids
+            compose_ids=compose_ids,
+            osbs_api=self
         )
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
         build_request.set_repo_info(repo_info)
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -11,6 +11,7 @@ import json
 import logging
 import os
 import re
+import yaml
 from pkg_resources import parse_version
 
 from six.moves import zip_longest
@@ -99,6 +100,7 @@ class BuildRequest(object):
         :param auto_build_node_selector: dict, a nodeselector for auto builds
         :param isolated_build_node_selector: dict, a nodeselector for isolated builds
         :param is_auto: bool, indicates if build is auto build
+        :param reactor_config_override: str, serialized reactor config yaml
         """
 
         # Here we cater to the koji "scratch" build type, this will disable
@@ -123,6 +125,7 @@ class BuildRequest(object):
 
         logger.debug("setting params '%s' for %s", kwargs, self.spec)
         self.spec.set_params(**kwargs)
+        self.osbs_api = kwargs.pop('osbs_api')
 
     def validate_build_variation(self):
         variations = (self.scratch, self.is_auto, self.isolated)
@@ -465,21 +468,73 @@ class BuildRequest(object):
             self.dj.dock_json_set_arg(phase, plugin, 'koji_principal', krb_principal)
             self.dj.dock_json_set_arg(phase, plugin, 'koji_keytab', krb_keytab)
 
-    def set_reactor_config_map(self):
-        if not self.spec.reactor_config_map.value:
+    def set_reactor_config(self):
+        if not self.spec.reactor_config_map.value and not self.spec.reactor_config_override.value:
             return
         custom = self.template['spec']['strategy']['customStrategy']
 
-        reactor_config_map = {
-            'name': 'REACTOR_CONFIG',
-            'valueFrom': {
-                'configMapKeyRef': {
-                    'name': self.spec.reactor_config_map.value,
-                    'key': 'config.yaml'
+        if self.spec.reactor_config_override.value:
+            reactor_config = {
+                'name': 'REACTOR_CONFIG',
+                'value': yaml.dump(self.spec.reactor_config_override.value)
+            }
+        elif self.spec.reactor_config_map.value:
+            reactor_config = {
+                'name': 'REACTOR_CONFIG',
+                'valueFrom': {
+                    'configMapKeyRef': {
+                        'name': self.spec.reactor_config_map.value,
+                        'key': 'config.yaml'
+                    }
                 }
             }
-        }
-        custom['env'].append(reactor_config_map)
+
+        custom['env'].append(reactor_config)
+
+    def set_required_secrets(self):
+        """
+        Sets required secrets
+        """
+        if not self.spec.reactor_config_map.value and not self.spec.reactor_config_override.value:
+            return
+
+        req_secrets_key = 'required_secrets'
+        token_secrets_key = 'worker_token_secrets'
+        required_secrets = []
+        token_secrets = []
+        if self.spec.reactor_config_override.value:
+            data = self.spec.reactor_config_override.value
+            required_secrets = data.get(req_secrets_key, [])
+            token_secrets = data.get(token_secrets_key, [])
+        elif self.spec.reactor_config_map.value:
+            config_map = self.osbs_api.get_config_map(self.spec.reactor_config_map.value)
+            required_secrets = config_map.get_data_by_key('config.yaml').get(req_secrets_key, [])
+            token_secrets = config_map.get_data_by_key('config.yaml').get(token_secrets_key, [])
+
+        if self.spec.platforms.value is not None:
+            required_secrets += token_secrets
+
+        if not required_secrets:
+            return
+
+        secrets = self.template['spec']['strategy']['customStrategy'].setdefault('secrets', [])
+        existing = set(secret_mount['secretSource']['name'] for secret_mount in secrets)
+        required_secrets = set(required_secrets)
+
+        already_set = required_secrets.intersection(existing)
+        if already_set:
+            logger.debug("secrets %s are already set", already_set)
+
+        for secret in required_secrets - existing:
+            secret_path = os.path.join(SECRETS_PATH, secret)
+            logger.info("Configuring %s secret at %s", secret, secret_path)
+
+            secrets.append({
+                'secretSource': {
+                    'name': secret,
+                },
+                'mountPath': secret_path,
+            })
 
     @staticmethod
     def remove_tag_and_push_registries(tag_and_push_registries, version):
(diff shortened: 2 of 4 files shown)


osbs/api.py
```diff
@@ -617,6 +617,7 @@ def _do_create_prod_build(self, git_uri, git_ref,
             scratch=self.build_conf.get_scratch(scratch),
             reactor_config_secret=self.build_conf.get_reactor_config_secret(),
             reactor_config_map=self.build_conf.get_reactor_config_map(),
+            reactor_config_yaml=self.build_conf.get_reactor_config_yaml(),
```
lcarva: This shouldn't be in `build_conf`, it should be a parameter like `koji_task_id`, for example.

osbs/build/build_request.py
```diff
@@ -481,6 +481,21 @@ def set_reactor_config_map(self):
...
+        if self.spec.platforms.value is not None:
+            return
+
+        if not self.spec.reactor_config_yaml.value:
+            return
+        custom = self.template['spec']['strategy']['customStrategy']
+
+        reactor_config_yaml = {
+            'name': 'REACTOR_CONFIG',
+            'value': self.spec.reactor_config_yaml.value
```
lcarva: This is the point that `reactor_config` needs to be serialized, if not already done.

osbs/build/build_request.py
```diff
@@ -1676,6 +1691,7 @@ def render(self, validate=True):
         self.render_version()
         self.render_node_selectors()
         self.set_reactor_config_map()
+        self.set_reactor_config_yaml()
```
lcarva: This method should be combined with `set_reactor_config_map` to avoid setting `REACTOR_CONFIG` twice.

```python
if self.spec.reactor_config_yaml.value:
  # set REACTOR_CONFIG with value
elif self.spec.reactor_config_map.value:
  # set REACTOR_CONFIG with valueFrom
```

I don't think we necessarily need to restrict `reactor_config_yaml` to just worker. If it's there use it. `reactor_config_yaml` takes precedence over `reactor_config_map`.

osbs/api.py

osbs/build/build_request.py
rcerven: ok

osbs/build/build_request.py
rcerven: ok

osbs/build/build_request.py
```diff
@@ -466,20 +468,62 @@ def set_kerberos_auth(self, plugins):
             self.dj.dock_json_set_arg(phase, plugin, 'koji_keytab', krb_keytab)
 
     def set_reactor_config_map(self):
```
lcarva: Rename this to `set_reactor_config` ?

osbs/build/build_request.py
```diff
@@ -466,20 +468,62 @@ def set_kerberos_auth(self, plugins):
...
+                'valueFrom': {
+                    'configMapKeyRef': {
+                        'name': self.spec.reactor_config_map.value,
+                        'key': 'config.yaml'
+                    }
                 }
             }
-        }
-        custom['env'].append(reactor_config_map)
+        elif self.spec.reactor_config_override.value:
```
lcarva: Override should be processed first as it takes precedence over `reactor_config_map`

osbs/build/build_request.py
rcerven: ok

osbs/build/build_request.py
rcerven: right

osbs/build/build_request.py
```diff
@@ -465,21 +467,63 @@ def set_kerberos_auth(self, plugins):
...
+
+        data =self.spec.reactor_config_override.value
+        required_secrets = data.get('required_secrets')
+        if not required_secrets :
+            return
+
+        if 'secrets' not in self.template['spec']['strategy']['customStrategy']:
+            self.template['spec']['strategy']['customStrategy']['secrets'] = []
+
+        custom = self.template['spec']['strategy']['customStrategy']
```
mlangsdorf: Nit: assign the custom first, then replace the if secrets block with:
custom.setdefault('secrets', [])

osbs/build/build_request.py
```diff
@@ -465,21 +467,63 @@ def set_kerberos_auth(self, plugins):
...
+        custom['env'].append(reactor_config)
+
+    def set_required_secrets(self):
+        """
+        Sets required secrets for worker build
+        """
+        if self.spec.platforms.value is not None:
+            return
+        if not self.spec.reactor_config_override.value:
+            return
```
mlangsdorf: Should the order of checks be reversed, and a validation exception raised if there's a config_override but no platforms?

osbs/build/build_request.py
```diff
@@ -465,21 +467,63 @@ def set_kerberos_auth(self, plugins):
...
-        }
-        custom['env'].append(reactor_config_map)
+
+        custom['env'].append(reactor_config)
+
+    def set_required_secrets(self):
+        """
+        Sets required secrets for worker build
+        """
+        if self.spec.platforms.value is not None:
```
lcarva: Actually `required_secrets` are meant to be used by both orchestrator and worker builds. This check should be removed.

osbs/build/build_request.py
lcarva: We should use either `reactor_config_override` or `reactor_config_map` to figure out secrets.

osbs/build/build_request.py
lcarva: `setdefault` also returns the value after it operates on it, so you could:

```python
secrets = self.template['spec']['strategy']['customStrategy'].setdefault('secrets')
```

That will return a reference to the `secrets` list which you can manipulate.

osbs/build/build_request.py
```diff
@@ -465,21 +467,63 @@ def set_kerberos_auth(self, plugins):
...
+        for secret in required_secrets:
+            secret_path = os.path.join(SECRETS_PATH, secret)
+
+            logger.info("Configuring %s secret at %s", secret, secret_path)
+            existing = [secret_mount for secret_mount in custom['secrets']
+                        if secret_mount['secretSource']['name'] == secret]
+
+            if existing:
+                logger.debug("secret %s already set", secret)
+            else:
```
lcarva: Could use `continue` on the if-block above to avoid nested code.

lcarva:
There's also `worker_token_secrets` that behaves similarly to `required_secrets` but it's only done when creating an orchestrator build.

osbs/build/build_request.py
rcerven: ahh ok

osbs/build/build_request.py
rcerven: ok

osbs/build/build_request.py
rcerven: ok

osbs/build/build_request.py
rcerven: ok

osbs/build/build_request.py
```diff
@@ -465,21 +468,67 @@ def set_kerberos_auth(self, plugins):
...
+            return
+
+        req_secrets_key = 'required_secrets'
+        required_secrets = []
+        if self.spec.reactor_config_override.value:
+            data = self.spec.reactor_config_override.value
+            required_secrets = data.get(req_secrets_key)
+        elif self.spec.reactor_config_map.value:
+            config_map = self.osbs_api.get_config_map(self.spec.reactor_config_map.value)
+            required_secrets = config_map.get_data_by_key(req_secrets_key)
```
lcarva: You probably want this:

```python
required_secrets = config_map.get_data_by_key('config.yaml')[req_secrets_key]
```

osbs/build/build_request.py
```diff
@@ -465,21 +468,67 @@ def set_kerberos_auth(self, plugins):
...
+        if not required_secrets:
+            return
+
+        secrets = self.template['spec']['strategy']['customStrategy'].setdefault('secrets', [])
+
+        for secret in required_secrets:
+            secret_path = os.path.join(SECRETS_PATH, secret)
+
+            logger.info("Configuring %s secret at %s", secret, secret_path)
+            existing = [secret_mount for secret_mount in secrets
```
lcarva: Might be more straightforward to use `set` for `existing`:

```python
existing = set(secret_mount['secretSource']['name'] for secret_mount in secrets)
for secret in required_secrets:
  ...
  if secret in existing:
    ...
    continue
  ...
  existing.add(secret)
```

osbs/build/build_request.py
rcerven: ok

osbs/build/build_request.py
rcerven: ok

rcerven:
added handling for : worker_token_secrets

osbs/build/build_request.py
```diff
@@ -465,21 +468,72 @@ def set_kerberos_auth(self, plugins):
...
+        config_map = None
+        if self.spec.reactor_config_override.value:
+            data = self.spec.reactor_config_override.value
+            required_secrets = data.get(req_secrets_key)
+        elif self.spec.reactor_config_map.value:
+            config_map = self.osbs_api.get_config_map(self.spec.reactor_config_map.value)
+            required_secrets = config_map.get_data_by_key('config.yaml').get(req_secrets_key, [])
+
+        if self.spec.platforms.value is not None and self.spec.reactor_config_map.value:
+            if not config_map:
```
lcarva: Can you just fetch `worker_token_secrets` on the if-elif blocks above? Then only Can you just fetch `worker_token_secrets` on the if-elif blocks above? Then combine with `required_secrets` in this if-block (orchestrator build)

osbs/build/build_request.py
```diff
@@ -465,21 +468,72 @@ def set_kerberos_auth(self, plugins):
...
+        """
+        if not self.spec.reactor_config_map.value and not self.spec.reactor_config_override.value:
+            return
+
+        req_secrets_key = 'required_secrets'
+        required_secrets = []
+        config_map = None
+        if self.spec.reactor_config_override.value:
+            data = self.spec.reactor_config_override.value
+            required_secrets = data.get(req_secrets_key)
```
lcarva: Fallback should be empty list: `get(req_secrets_key, [])`

osbs/build/build_request.py
```diff
@@ -465,21 +468,72 @@ def set_kerberos_auth(self, plugins):
...
+            required_secrets +=\
+                config_map.get_data_by_key('config.yaml').get('worker_token_secrets', [])
+
+        if not required_secrets:
+            return
+
+        secrets = self.template['spec']['strategy']['customStrategy'].setdefault('secrets', [])
+        existing = set(secret_mount['secretSource']['name'] for secret_mount in secrets)
+
+        for secret in set(required_secrets):
```
lcarva: Could exclude `existing` before iterating:
```python
for secret in set(required_secrets) - existing:
  ...
```

osbs/build/build_request.py
rcerven: I can't do that, because if override will be defined, the elif with fetching config won't run

so I can either just fetch config every time, or leave it as it is

osbs/build/build_request.py
rcerven: adding default empty list to spec

osbs/build/build_request.py
rcerven: so we don't want to report debug that secret is already set?

lcarva: I don't see a lot of value in reporting that, but you could log `existing` before entering the loop.

osbs/build/build_request.py
rcerven: nvm, adding default here

osbs/build/build_request.py
rcerven: ok

osbs/build/build_request.py
rcerven: ok I'll report intersection then

lcarva:
Updated release notes draft:
> - reactor_config_override parameter

PR #738, opened 2018-03-12
Arrangement 6: Add support for BuildUserParams, BuildRequestV2, and PluginsConfiguration

mlangsdorf:
As part of the move to arrangement 6, create PluginsConfiguration, BuildRequestV2 and the
underlying BuildUserParams.

(move from PR733 because github keeps stalling out on me).

Currently the templates are *_inner:7.json to avoid collisions with the DEFAULT_ARRANGEMENT_VERSION tests.

Please review.

Signed-off-by: Mark Langsdorf mlangsdo@redhat.com

### DIFF
--- a/inputs/orchestrator_inner:7.json
+++ b/inputs/orchestrator_inner:7.json
@@ -1,0 +1,154 @@
+{
+  "client_version": "{{VERSION}}",
+  "prebuild_plugins": [
+    {
+      "name": "reactor_config"
+    },
+    {
+      "name": "resolve_module_compose"
+    },
+    {
+      "name": "flatpak_create_dockerfile",
+      "args": {
+        "base_image": "{{BASE_IMAGE}}"
+      }
+    },
+    {
+      "name": "add_filesystem",
+      "args": {
+        "repos": []
+      }
+    },
+    {
+      "name": "inject_parent_image",
+      "args": {
+        "koji_parent_build": "{{KOJI_PARENT_BUILD}}"
+      }
+    },
+    {
+      "name": "pull_base_image",
+      "args": {
+        "parent_registry_insecure": true
+      }
+    },
+    {
+      "name": "bump_release"
+    },
+    {
+      "name": "add_labels_in_dockerfile",
+      "args": {
+        "labels": "{{IMPLICIT_LABELS}}"
+      }
+    },
+    {
+      "name": "koji_parent",
+      "required": false
+    },
+    {
+      "name": "check_and_set_rebuild",
+      "args": {
+        "label_key": "is_autorebuild",
+        "label_value": "true",
+        "verify_ssl": false
+      }
+    },
+    {
+      "name": "resolve_composes",
+      "required": false
+    }
+  ],
+  "buildstep_plugins": [
+    {
+      "name": "orchestrate_build",
+      "args": {
+        "platforms": "{{PLATFORMS}}",
+        "build_kwargs": "{{BUILD_KWARGS}}",
+        "osbs_client_config": "{{OSBS_CLIENT_CONFIG}}"
+      }
+    }
+  ],
+  "postbuild_plugins": [
+    {
+      "name": "fetch_worker_metadata"
+    },
+    {
+      "name": "compare_components"
+    },
+    {
+      "name": "tag_from_config",
+      "args": {
+        "tag_suffixes": "{{TAG_SUFFIXES}}"
+      }
+    },
+    {
+      "name": "group_manifests"
+    },
+    {
+      "name": "pulp_tag"
+    },
+    {
+      "name": "pulp_sync",
+      "args": {
+        "dockpulp_loglevel": "INFO",
+        "publish": false
+      }
+    }
+  ],
+  "prepublish_plugins": [],
+  "exit_plugins": [
+    {
+      "name": "pulp_publish",
+      "args": {
+        "dockpulp_loglevel": "INFO"
+      }
+    },
+    {
+      "name": "pulp_pull",
+      "args": {
+        "insecure": true
+      }
+    },
+    {
+      "name": "import_image",
+      "args": {
+        "imagestream": "{{IMAGESTREAM}}",
+        "verify_ssl": false,
+        "build_json_dir": "{{PATH}}"
+      },
+      "required": false
+    },
+    {
+      "name": "delete_from_registry"
+    },
+    {
+      "name": "koji_import",
+      "args": {
+        "verify_ssl": false,
+        "koji_keytab": false
+      }
+    },
+    {
+      "name": "koji_tag_build",
+      "args": {
+        "koji_keytab": false,
+        "koji_principal": false
+      }
+    },
+    {
+      "name": "store_metadata_in_osv3",
+      "args": {
+        "verify_ssl": false
+      }
+    },
+    {
+      "name": "sendmail"
+    },
+    {
+      "name": "remove_built_image"
+    },
+    {
+      "name": "remove_worker_metadata",
+      "required": false
+    }
+  ]
+}
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -21,6 +21,7 @@ from contextlib import contextmanager
 from types import GeneratorType
 
 from osbs.build.build_request import BuildRequest
+from osbs.build.build_requestv2 import BuildRequestV2
 from osbs.build.build_response import BuildResponse
 from osbs.build.pod_response import PodResponse
 from osbs.build.config_map_response import ConfigMapResponse
@@ -28,7 +29,8 @@ from osbs.constants import (BUILD_RUNNING_STATES, WORKER_OUTER_TEMPLATE,
                             WORKER_INNER_TEMPLATE, WORKER_CUSTOMIZE_CONF,
                             ORCHESTRATOR_OUTER_TEMPLATE, ORCHESTRATOR_INNER_TEMPLATE,
                             ORCHESTRATOR_CUSTOMIZE_CONF, BUILD_TYPE_WORKER,
-                            BUILD_TYPE_ORCHESTRATOR, BUILD_FINISHED_STATES)
+                            BUILD_TYPE_ORCHESTRATOR, BUILD_FINISHED_STATES,
+                            DEFAULT_ARRANGEMENT_VERSION, REACTOR_CONFIG_ARRANGEMENT_VERSION)
 from osbs.core import Openshift
 from osbs.exceptions import (OsbsException, OsbsValidationException, OsbsResponseException,
                              OsbsOrchestratorNotEnabled)
@@ -170,24 +172,32 @@ class OSBS(object):
 
     @osbsapi
     def get_build_request(self, build_type=None, inner_template=None,
-                          outer_template=None, customize_conf=None):
+                          outer_template=None, customize_conf=None,
+                          arrangement_version=DEFAULT_ARRANGEMENT_VERSION):
         """
-        return instance of BuildRequest
+        return instance of BuildRequest or BuildRequestV2
 
         :param build_type: str, unused
         :param inner_template: str, name of inner template for BuildRequest
         :param outer_template: str, name of outer template for BuildRequest
         :param customize_conf: str, name of customization config for BuildRequest
-        :return: instance of BuildRequest
+        :param arrangement_version: int, value of the arrangement version
+
+        :return: instance of BuildRequest or BuildRequestV2
         """
         if build_type is not None:
             warnings.warn("build types are deprecated, do not use the build_type argument")
 
-        build_request = BuildRequest(
-            build_json_store=self.os_conf.get_build_json_store(),
-            inner_template=inner_template,
-            outer_template=outer_template,
-            customize_conf=customize_conf)
+        if arrangement_version < REACTOR_CONFIG_ARRANGEMENT_VERSION:
+            build_request = BuildRequest(
+                build_json_store=self.os_conf.get_build_json_store(),
+                inner_template=inner_template,
+                outer_template=outer_template,
+                customize_conf=customize_conf)
+        else:
+            build_request = BuildRequestV2(
+                build_json_store=self.os_conf.get_build_json_store(),
+                customize_conf=customize_conf)
 
         # Apply configured resource limits.
         cpu_limit = self.build_conf.get_cpu_limit()
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -22,6 +22,8 @@ from types import GeneratorType
 
 from osbs.build.build_request import BuildRequest
 from osbs.build.build_requestv2 import BuildRequestV2
+from osbs.build.user_params import BuildUserParams
+from osbs.build.plugins_configuration import PluginsConfiguration
 from osbs.build.build_response import BuildResponse
 from osbs.build.pod_response import PodResponse
 from osbs.build.config_map_response import ConfigMapResponse
@@ -1218,3 +1220,9 @@ class OSBS(object):
         self.os.retries_enabled = False
         yield
         self.os.retries_enabled = True
+
+    @osbsapi
+    def render_plugins_configuration(self, user_params_json):
+        user_params = BuildUserParams.from_json(user_params_json)
+
+        return PluginsConfiguration(user_params).render()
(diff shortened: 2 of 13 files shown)


osbs/api.py
```diff
@@ -21,14 +21,18 @@
 from types import GeneratorType
 
 from osbs.build.build_request import BuildRequest
+from osbs.build.build_requestv2 import BuildRequestV2
+from osbs.build.user_params import BuildUserParams
+from osbs.build.plugins_configuration import PluginsConfiguration
```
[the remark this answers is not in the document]
mlangsdorf: Fixed

inputs/orchestrator_inner:7.json
```diff
@@ -0,0 +1,154 @@
+{
```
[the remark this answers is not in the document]
mlangsdorf: quoting from the top of the PR "Currently the templates are *_inner:7.json to avoid collisions with the DEFAULT_ARRANGEMENT_VERSION tests."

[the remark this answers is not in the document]
lcarva: ah just temporarily, ack.

osbs/build/build_request.py
```diff
@@ -465,16 +465,16 @@ def set_kerberos_auth(self, plugins):
             self.dj.dock_json_set_arg(phase, plugin, 'koji_principal', krb_principal)
             self.dj.dock_json_set_arg(phase, plugin, 'koji_keytab', krb_keytab)
 
-    def set_reactor_config_map(self):
-        if not self.spec.reactor_config_map.value:
+    def set_reactor_config_map(self, reactor_config_map):
```
lcarva: Let's rebase with master as this method has changed.

osbs/build/build_request.py
```diff
@@ -1464,12 +1464,12 @@ def render_node_selectors(self):
...
                 self.template['spec']['nodeSelector'].update(self.platform_node_selector)
 
-    def render(self, validate=True):
+    def render(self, api=None, validate=True):
         if validate:
             self.spec.validate()
 
         self.render_customizations()
-        self.render_name()
+        self.render_name(self.spec.name, self.spec.image_tag, self.spec.platform)
```
lcarva: Would it be simpler to follow if these render methods took the actual value of the `BuildParams`?
```python
self.render_name(self.spec.name.value, self.spec.image_tag.value, self.spec.platform.value)
```

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,225 @@
...
+    @property
+    def customize_conf(self):
+        raise RuntimeError('customize_conf not supported in BuildRequestV2')
+
+    # Override
+    @property
+    def dj(self):
+        raise RuntimeError('DockJson not supported in BuildRequestV2')
+
+    def set_reactor_config_map(self, api):
```
lcarva: No need to override. See parent's `set_reactor_config`.

Come to think of it, should that method move to this class? Usage of `reactor_config_map/override` is really only meant to be supported in arrangement 6.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,225 @@
...
+        Scratch builds must not affect subsequent builds,
+        and should not be imported into Koji.
+        """
+        if self.scratch:
+            self.template['spec'].pop('triggers', None)
+            self.set_label('scratch', 'true')
+
+    def render_node_selectors(self):
+        # for worker builds set nodeselectors
+        if self.user_params.platforms.value is None:
```
lcarva: Change parent class to take `platforms` as parameter to avoid duplication?

osbs/build/plugins_configuration.py
```diff
@@ -0,0 +1,583 @@
...
+                                    (path, repr(ex)))
+        return self._template
+
+    def remove_plugin(self, plugin_type, plugin_name):
+        """
+        if config contains plugin, remove it
+        """
+        for p in self.template[plugin_type]:
+            if p.get('name') == plugin_name:
+                self.template[plugin_type].remove(p)
```
lcarva: Just an idea, but `remove_plugin`, `add_plugin` and all `template_*` methods could be grouped together in a new class to better isolate concerns.

osbs/build/plugins_configuration.py
```diff
@@ -0,0 +1,583 @@
...
+        if not self.template_has_plugin_conf(phase, plugin):
+            return
+
+        if self.user_params.platforms.value is None:
+            logger.debug('removing %s plugin: no platforms', plugin)
+            self.remove_plugin(phase, plugin)
+            return
+
+        # Parameters to be used in call to create_worker_build
+        build_kwargs = {
```
lcarva: Should this be using a variation of `self.user_params.to_json()` that returns a `dict` instead?

osbs/build/plugins_configuration.py
```diff
@@ -0,0 +1,583 @@
...
+        phase = 'prepublish_plugins'
+        plugin = 'squash'
+
+        if self.user_params.flatpak.value:
+            # We'll extract the filesystem anyways for a Flatpak instead of exporting
+            # the docker image directly, so squash just slows things down.
+            self.remove_plugin(phase, plugin)
+            return
+
+    def render_tag_from_config(self):
```
lcarva: FYI: there's been some changes to parent's version of this method.

osbs/build/plugins_configuration.py
```diff
@@ -0,0 +1,583 @@
...
+            elif self.user_params.isolated.value:
+                tag_suffixes['primary'].extend(['{version}-{release}'])
+            else:
+                tag_suffixes['primary'].extend(['latest', '{version}', '{version}-{release}'])
+                if self.user_params.additional_tags:
+                    tag_suffixes['primary'].extend(self.user_params.additional_tags.tags)
+
+        self.template_set_arg(phase, plugin, 'tag_suffixes', tag_suffixes)
+
+    def render(self):
```
lcarva: Parent's version calls `self.spec.validate()`, do the same here for `user_params` ?

osbs/build/build_requestv2.py
mlangsdorf: Sure, fixed.

osbs/build/build_requestv2.py
mlangsdorf: set_reactor_config_map() existed in BuildRequest before this latest round of changes. I'm not sure about removing it.

I'll just add the parameters to BuildRequest.set_reactor_config() and BuildRequest.set_required_secrets() and see how that goes.

osbs/build/plugins_configuration.py
mlangsdorf: It's identical to the code from BuildRequest, so I think we only want that subset of build parameters.

If you think we should pass all the user parameters, it would be easy if not efficient to do build_kwargs = json.loads(self.user_params.to_json())

osbs/build/plugins_configuration.py
mlangsdorf: There have been?  It doesn't look substantially different.

osbs/build/plugins_configuration.py
mlangsdorf: Fixed.

osbs/build/build_request.py
mlangsdorf: Fixed.

osbs/api.py
```diff
@@ -211,7 +223,7 @@ def create_build_from_buildrequest(self, build_request):
         :return: instance of build.build_response.BuildResponse
         """
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
-        build = build_request.render()
+        build = build_request.render(self)
```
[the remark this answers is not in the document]
mlangsdorf: BuildReqestV2.set_params is an override of BuildRequest.set_params, but I'll add it in.

osbs/constants.py
```diff
@@ -23,6 +23,7 @@
 WORKER_INNER_TEMPLATE = "worker_inner:{arrangement_version}.json"
 ORCHESTRATOR_INNER_TEMPLATE = "orchestrator_inner:{arrangement_version}.json"
 DEFAULT_ARRANGEMENT_VERSION = 5  # this should be the highest-numbered version
+REACTOR_CONFIG_ARRANGEMENT_VERSION = DEFAULT_ARRANGEMENT_VERSION + 2
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

osbs/build/user_params.py
```diff
@@ -0,0 +1,207 @@
...
+        self.arrangement_version.value = REACTOR_CONFIG_ARRANGEMENT_VERSION
+        self.required_params = [
+            self.build_json_dir,
+            self.build_type,
+            self.git_ref,
+            self.git_uri,
+            self.koji_target,
+            self.user,
+        ]
+        self.convert_dict = {
```
lcarva: Can this be a tuple of strings instead? Less room to make a typo.
You can use `setattr` and `getattr` to modify/access attributes dynamically.

`convert_dict`/`convert_keys` should probably be a class attribute instead.

osbs/build/user_params.py
```diff
@@ -0,0 +1,207 @@
...
+        self.scratch = BuildParam("release", allow_none=True)
+        self.koji_task_id = BuildParam("koji_task_id", allow_none=True)
+        self.koji_parent_build = BuildParam("koji_parent_build", allow_none=True)
+        self.flatpak = BuildParam("flatpak", default=False)
+        self.flatpak_base_image = BuildParam("flatpak_base_image", allow_none=True)
+        self.signing_intent = BuildParam("signing_intent", allow_none=True)
+        self.yum_repourls = BuildParam("yum_repourls")
+        self.compose_ids = BuildParam("compose_ids", allow_none=True)
+        self.reactor_config_map = BuildParam("reactor_config_map", allow_none=True)
+        self.additional_tags = None
```
lcarva: Why is `additional_tags` different?

osbs/build/build_request.py
mlangsdorf: Sure, why not.

Fixed.

osbs/build/user_params.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def __repr__(self):
+        return "UserParams(%s)" % self.__dict__
+
+    def from_json(self, user_params_json):
+        if not user_params_json:
+            return
+        json_dict = json.loads(user_params_json)
+        logger.debug("CALLED FROM JSON HERE with dict %s", json_dict)
+        for key, value in json_dict.items():
```
lcarva: Can we just call `self.set_params(**json_dict)` ?

osbs/build/user_params.py
mlangsdorf: it's an AdditionalTagsConfig structure out of repo_utils. It's a little weird in all kinds of ways.

osbs/build/user_params.py
```diff
@@ -0,0 +1,207 @@
...
+            logger.debug("read additional tags %s from json", json_dict['additional_tags'])
+            self.additional_tags = AdditionalTagsConfig(tags=json_dict['additional_tags'])
+
+    def set_if_exists(self, json_dict, param):
+        if self.convert_dict[param].value:
+            json_dict[param] = self.convert_dict[param].value
+
+    def to_json(self):
+        json_dict = {}
+        for key in self.convert_dict:
```
lcarva: What if we just iterate through object attributes and only operate on `BuildParam` types?
```python
for x, param in b.__dict__.items():
    if isinstance(param, BuildParam):
        print x, param.value
```

osbs/build/plugins_configuration.py
lcarva: You could add `to_dict()` helper method in `BuildUserParams`, which can also be used by `to_json`.

osbs/build/plugins_configuration.py
lcarva: See https://github.com/projectatomic/osbs-client/pull/728/files

osbs/build/user_params.py
mlangsdorf: base_image isn't in the json, so that would crash non-flatpak builds.

lcarva: It might be easier to pluck out parameters that need special handling rather than maintaining a list of everything that needs to be serialized though.

osbs/build/user_params.py
mlangsdorf: Fixed I think.

osbs/build/user_params.py
mlangsdorf: I've used your suggestion below to create convert_dict during __init__ from the BuildParams, so we don't have to maintain the list.

osbs/build/user_params.py
mlangsdorf: I'm not at all sure what you mean here.

lcarva: Since you're populating `convert_dict` dynamically, my suggestion is no longer relevant.

What I had meant is that you don't really need a dict that maps to each BuildParam, all you need a list/tuple of attribute names.

```python
keys = ('koji_target', 'user')

# For reading attributes
for key in keys:
  getattr(self, key)

# For setting attributes
for key in keys:
  setattr(self, key, value)
```

osbs/build/plugins_configuration.py
mlangsdorf: Fixed.

osbs/build/plugins_configuration.py
mlangsdorf: Fixed.

osbs/build/build_requestv2.py
```diff
@@ -0,0 +1,184 @@
...
+        :param platform: str, platform
+        :param koji_target: str, koji tag with packages used to build the image
+        :param koji_task_id: str, koji ID
+        :param koji_parent_build: str,
+        :param flatpak: if we should build a Flatpak OCI Image
+        :param flatpak_base_image: str, name of the Flatpack OCI Image
+        :param reactor_config_map: str, name of the config map containing the reactor environment
+        :param yum_repourls: list of str, uris of the yum repos to pull from
+        :param signing_intent: bool, True to sign the resulting image
+        :param compose_ids: list of str,
```
twaugh: :param compose_ids, list of int, ODCS composes to use instead of generating them

osbs/build/plugins_configuration.py
```diff
@@ -0,0 +1,591 @@
...
+        # sure that the build is unique in Koji
+        if self.user_params.flatpak.value:
+            self.pt.set_plugin_arg(phase, plugin, 'append', True)
+
+    def render_import_image(self, use_auth=None):
+        """
+        Configure the import_image plugin
+        """
+        # import_image is a multi-phase plugin
+        phases = ('postbuild_plugins', 'exit_plugins')
```
twaugh: We only have to handle it being an exit plugin here, so we can simplify this method a little: `phase = 'exit_plugins'`, and no need to loop.

osbs/build/user_params.py
```diff
@@ -0,0 +1,183 @@
...
+                raise OsbsValidationException("name_label must be provided")
+            self.imagestream_name.value = name_label.replace('/', '-')
+
+        if signing_intent and compose_ids:
+            raise OsbsValidationException(
+                'Please only define signing_intent -OR- compose_ids, not both')
+        if compose_ids and yum_repourls:
+            raise OsbsValidationException(
+                'Please only define yum_repourls -OR- compose_ids, not both')
+        if not (compose_ids is None or isinstance(compose_ids, list)):
```
twaugh: Minor: can do this more flexibly with:
```
import collections
[...]
if isinstance(x, collections.Sequence):
```

(then it works for any sequence, e.g. tuples as well.

mlangsdorf: strings are collections and so this fails an existing test case. Leaving it as it was.

osbs/build/plugins_configuration.py
mlangsdorf: Fixed.

osbs/build/build_requestv2.py
mlangsdorf: Fixed, thanks.

lcarva:
The message for third commit hints there's a question that needs to be answered to unblock this MR.
I think this has been resolved. I'm waiting on a response before doing another review.

mlangsdorf:
Custom templates are supported.
Plugins are disabled appropriately, either here or in atomic_reactor. I think this is ready to go.

inputs/orchestrator_inner:7.json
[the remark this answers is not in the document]
lcarva: What's the expectation here? Will this be changed to arrangement 6 before or after PR is merged?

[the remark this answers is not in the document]
mlangsdorf: I'll change them to arrangement 6 as part of the next PR when I actually create arrangement 6.

mlangsdorf:
added to release notes
- plugin configuration can be rendered from reactor config and user parameters
under # API Changes

PR #739, opened 2018-03-14
add info about buildroot image in build annotations, for direct build…

rcerven:
…s without buildConfig

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -311,6 +311,9 @@ class OSBS(object):
         build_json['spec']['serviceAccount'] = 'builder'
 
         builder_img = build_json['spec']['strategy']['customStrategy']['from']
+        build_name = builder_img['name']
+        build_kind = builder_img['kind']
+
         kind = builder_img['kind']
         if kind == 'ImageStreamTag':
             # Only BuildConfigs get to specify an ImageStreamTag. When
@@ -320,6 +323,11 @@ class OSBS(object):
             ref = response.json()['image']['dockerImageReference']
             builder_img['kind'] = 'DockerImage'
             builder_img['name'] = ref
+
+        build_json['metadata'].setdefault('annotations', {})
+        build_json['metadata']['annotations']['from'] = json.dumps({
+            'kind': build_kind,
+            'name': build_name})
 
         if unique:
             unique_labels = {}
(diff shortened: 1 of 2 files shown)


osbs/api.py
```diff
@@ -311,6 +311,9 @@ def _create_build_directly(self, build_request, unique=None):
         build_json['spec']['serviceAccount'] = 'builder'
 
         builder_img = build_json['spec']['strategy']['customStrategy']['from']
+        build_name = builder_img['name']
+        build_kind = builder_img['kind']
```
mlangsdorf: Can't we use the existing kind assignment here instead?

osbs/api.py
```diff
@@ -321,6 +324,14 @@ def _create_build_directly(self, build_request, unique=None):
             builder_img['kind'] = 'DockerImage'
             builder_img['name'] = ref
 
+        if 'annotations' not in build_json['metadata']:
+            build_json['metadata']['annotations'] = {}
```
mlangsdorf: build_json['metadata'].setdefault('annotations', {})

osbs/api.py
rcerven: if you mean what was there before, no because if there was imagestreamtag it will get changed int dockerimage, and I need the original

osbs/api.py

osbs/api.py
mlangsdorf: Right. Ignore this one, then.

rcerven:
also updated unit test

rcerven:
using 'kind:name' in from instead of dict, as annotations can have only strings

osbs/api.py
```diff
@@ -321,6 +324,9 @@ def _create_build_directly(self, build_request, unique=None):
             builder_img['kind'] = 'DockerImage'
             builder_img['name'] = ref
 
+        build_json['metadata'].setdefault('annotations', {})
+        build_json['metadata']['annotations']['from'] = ','.join([build_kind, build_name])
```
twaugh: IMHO we should just use `json.dumps(builder_img)` here.

mlangsdorf: Don't we run into the issue that kind and name in builder_image get changed if it's an ImageStreamTag?

twaugh: Well, right, it has to be done before that. Just commenting on the format of the annotation, that's all.

rcerven: ok I can do json.dumps

lcarva:
Updated release notes draft:
> - Annotation added to direct builds (scratch/isolated) for tracking actual requested buildroot image

PR #742, opened 2018-03-16
don't remove reactor_config plugin if config_map is provided

rcerven:

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -228,7 +228,9 @@ class BuildRequest(object):
         self.template['metadata']['labels'][name] = value
 
     def render_reactor_config(self):
-        if self.spec.reactor_config_secret.value is None:
+        if self.spec.reactor_config_secret.value is None and \
+                self.spec.reactor_config_map.value is None and \
+                self.spec.reactor_config_override.value is None:
             logger.debug("removing reactor_config plugin: no secret")
             self.dj.remove_plugin('prebuild_plugins', 'reactor_config')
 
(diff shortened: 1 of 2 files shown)


osbs/build/build_request.py
```diff
@@ -228,7 +228,9 @@ def set_label(self, name, value):
         self.template['metadata']['labels'][name] = value
 
     def render_reactor_config(self):
-        if self.spec.reactor_config_secret.value is None:
+        if self.spec.reactor_config_secret.value is None and \
+                not self.spec.reactor_config_map.value and \
```
lcarva: It's a little confusing that for some values we're comparing to None while others verifying they're falsy. Can we make it the same? `is None` seems like a reasonable check for all.

lcarva:
Needs unit tests.

osbs/build/build_request.py
rcerven: ok

rcerven:
unit test added

tests/build_/test_build_request.py
```diff
@@ -3091,3 +3091,50 @@ def test_set_required_secrets(self, platforms, reactor_config_map,
...
+            'reactor_config_secret': reactor_config_secret,
+            'osbs_api': MockOSBSApi(),
+        }
+        build_request.set_params(**kwargs)
+        build_json = build_request.render()
+
+        plugins = get_plugins_from_build_json(build_json)
+
+        if reactor_config_secret is None and reactor_config_override is None and \
+                reactor_config_map is None:
```
twaugh: This is repeating the logic inside the test, but I don't see an easier way of writing it without introducing a truth table with all 8 possibilities.

twaugh:
Typo in commit message ("conig")

rcerven:
fixed typo in commit msg

lcarva:

PR #743, opened 2018-03-17
Handle new behavior for etcd data v3

lcarva:


mlangsdorf:
Added to release notes
- avoid 410 errors while watching imagestreams using the v3 data format
under # Bug Fixes

PR #745, opened 2018-03-20
arrangement6: enable arrangement 6

mlangsdorf:
Add arrangement 6 to osbs-client. Make all test cases pass for it and add some
test cases to test_arrangements.

Depends on https://github.com/projectatomic/atomic-reactor/pull/952


inputs/orchestrator_inner:6.json
```diff
@@ -8,37 +7,22 @@
...
     },
     {
-      "name": "add_labels_in_dockerfile",
-      "args": {
-        "labels": "{{IMPLICIT_LABELS}}"
-      }
+      "name": "add_labels_in_dockerfile"
     },
     {
       "name": "koji_parent",
```
lcarva: I think it's safe to remove the `required` key for this plugin now.

inputs/orchestrator_inner:6.json
```diff
@@ -8,41 +7,25 @@
...
-      }
+      "name": "add_labels_in_dockerfile"
     },
     {
-      "name": "koji_parent",
-      "required": false
+      "name": "koji_parent"
     },
     {
       "name": "check_and_set_rebuild",
```
lcarva: Let's remove `"verify_ssl"` too since that's an environment config.

inputs/orchestrator_inner:6.json
```diff
@@ -59,12 +42,7 @@
   ],
```
lcarva: Let's make "resolve_composes" required by removing the ` "required": false` property.

inputs/orchestrator_inner:6.json
```diff
@@ -59,12 +42,7 @@
...
-      "name": "orchestrate_build",
-      "args": {
-        "platforms": "{{PLATFORMS}}",
-        "build_kwargs": "{{BUILD_KWARGS}}",
-        "osbs_client_config": "{{OSBS_CLIENT_CONFIG}}"
-      }
+      "name": "orchestrate_build"
     }
   ],
   "postbuild_plugins": [
```
lcarva: I can't comment further down, but let's also remove:

- [ ] `dockpulp_loglevel` from `pulp_sync` and `pulp_publish`
- [ ]  `"koji_keytab"` and  `"koji_principal"` from `koji_tag_build`
- [ ] `"verify_ssl"` and "build_json_dir" from `import_image`
- [ ] `"verify_ssl"` from `store_metadata_in_osv3`

I think it's also safe to make `remove_worker_metadata` required by remove `"required": false` property.

lcarva: I'm not sure what to make out of `pulp_pull`, maybe that can be addressed later.

inputs/worker_inner:6.json
```diff
@@ -111,10 +95,7 @@
   ],
```
lcarva: Let's remove `"verify_ssl"`, `"koji_keytab"`, and `"koji_principal"` from `koji_upload`

inputs/orchestrator_inner:6.json
mlangsdorf: Fixed I think.

inputs/worker_inner:6.json
mlangsdorf: Fixed I think.

inputs/orchestrator_inner:6.json
mlangsdorf: Fixed.

inputs/orchestrator_inner:6.json
```diff
@@ -0,0 +1,112 @@
...
+      "name": "import_image",
+      "required": false
+    },
+    {
+      "name": "delete_from_registry"
+    },
+    {
+      "name": "koji_import",
+      "args": {
+        "koji_keytab": false
```
lcarva: Let's remove this.

inputs/worker_inner:6.json
```diff
@@ -102,19 +86,13 @@
     {
```
lcarva: `pulp_push` needs `"publish": false` to avoid double publishing content

osbs/build/build_requestv2.py
```diff
@@ -59,6 +59,8 @@ def set_params(self, **kwargs):
         :param flatpak: if we should build a Flatpak OCI Image
         :param flatpak_base_image: str, name of the Flatpack OCI Image
         :param reactor_config_map: str, name of the config map containing the reactor environment
+        :param reactor_config_override: str, name of the config map containing the
```
lcarva: This is not correct. `build_request.py` says:
```
:param reactor_config_override: str, serialized reactor config yaml
```
Which is also not correct.

It should be something like this instead:
```
:param reactor_config_override: dict, data structure for reactor config to be injected as an environment variable into build; when used, reactor_config_map is ignored.
```

osbs/build/build_requestv2.py
```diff
@@ -177,6 +182,15 @@ def render(self, validate=True):
...
 
+        # Set our environment variables
+        custom_strategy['env'].append({
+            'name': 'USER_PARAMS',
+            'value': self.user_params.to_json(),
+        })
+        # delete the ATOMIC_REACTOR_PLUGINS placeholder
+        for env in custom_strategy['env']:
+            if env['name'] == 'ATOMIC_REACTOR_PLUGINS':
+                del custom_strategy['env'][custom_strategy['env'].index(env)]
```
lcarva: Should probably `break` here to avoid some oddities with modification while iterating.

Also, you could use `enumerate` to capture the index:

```python
for index, env in custom_strategy['env']:
  if env['name'] ==  'ATOMIC_REACTOR_PLUGINS':
    del custom_strategy['env'][index]
    break
```

osbs/build/build_requestv2.py
mlangsdorf: Fixed, thanks.

osbs/build/build_requestv2.py

inputs/worker_inner:6.json
mlangsdorf: Fixed.

inputs/orchestrator_inner:6.json
mlangsdorf: Fixed, along with its test code.

inputs/orchestrator_inner:6.json
```diff
@@ -0,0 +1,112 @@
...
+  ],
+  "prepublish_plugins": [],
+  "exit_plugins": [
+    {
+      "name": "pulp_publish"
+    },
+    {
+      "name": "pulp_pull",
+      "args": {
+        "insecure": true
```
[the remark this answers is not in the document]
mlangsdorf: Leaving alone for now.

inputs/orchestrator_inner:6.json
```diff
@@ -0,0 +1,112 @@
...
+      }
+    },
+    {
+      "name": "import_image",
+      "required": false
```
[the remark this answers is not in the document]
mlangsdorf: Leaving alone for now.

tests/build_/test_arrangements.py
```diff
@@ -113,7 +130,29 @@ def get_build_request(self, build_type, osbs,  # noqa:F811
 
         params.update(additional_params or {})
         params['arrangement_version'] = self.ARRANGEMENT_VERSION
-        return params, fn(**params).json
+        if self.ARRANGEMENT_VERSION < REACTOR_CONFIG_ARRANGEMENT_VERSION:
```
lcarva: I'm wondering if `TestArrangementV6` should extend this method:

```python
def get_build_request(self, *args, **kwargs):
  params, build_json = super(TestArrangementV6).get_build_request(*args, **kwargs)
  ...
```

mlangsdorf: I made a similar adjustment to test_running_order

osbs/api.py
```diff
@@ -190,7 +190,7 @@ def get_build_request(self, build_type=None, inner_template=None,
         if build_type is not None:
             warnings.warn("build types are deprecated, do not use the build_type argument")
 
-        if arrangement_version < REACTOR_CONFIG_ARRANGEMENT_VERSION:
+        if not arrangement_version or arrangement_version < REACTOR_CONFIG_ARRANGEMENT_VERSION:
```
twaugh: I'd prefer if we either
* don't supply that keyword argument in `_do_create_prod_build` unless we really have a value for it (i.e. pass a **kwargs), or
* have the default value as `None` and then set it to the default as needed in this method body, e.g.
```
if arrangement_version is None:
    arrangement_version = DEFAULT_ARRANGEMENT_VERSION
```

mlangsdorf: I wrote it this way because there are dozens of existing tests that don't pass an arrangement_version and will fail if arrangement_version == REACTOR_CONFIG_ARRANGEMENT_VERSION because they don't pass a build_type in the kwargs to build_request.set_params(), but BuildRequestV2 requires one.

I'd rather not rewrite the tests. And I can't edit the kwargs that will be used inside of build_request.set_params from inside get_build_request, so I can't go back and add a build_type if I'm returning a BuildRequestV2.

If there's a better way to do this, I'm interested, but final_version = arrangement_version or DEFAULT_ARRANGEMENT_VERSION is not right unless I rewrite all those tests.

osbs/build/plugins_configuration.py
```diff
@@ -422,7 +414,8 @@ def render_orchestrate_build(self):
         worker_params = [
             'component', 'git_branch', 'git_ref', 'git_uri', 'koji_task_id',
             'filesystem_koji_task_id', 'scratch', 'koji_target', 'user', 'yum_repourls',
-            'arrangement_version', 'koji_parent_build', 'isolated',
+            'arrangement_version', 'koji_parent_build', 'isolated', 'reactor_config_map',
+            'reactor_config_override'
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

mlangsdorf:
added the following text to the release notes:

Default arrangement version is 6.

## API changes
- get_build_request takes an optional arrangement_version argument, and returns a BuildRequestV2 object if arrangement_version is greater than or equal to 6.
- BuildRequestV2 objects ignore most arguments passed in via set_params, as those arguments are retrieved from the REACTOR_CONFIG config map at runtime. the remaining arguments are stored in the USER_PARAMS field of the the build_json['spec']['strategy']['custom_strategy']['env']
- new API call render_plugins_configuration takes the json from USER_PARAMS and returns the json of the atomic-reactor plugins configuration, much like BuildRequest.render() did in previous arrangements

## Build JSON changes
- ['spec']['strategy']['custom_strategy']['env']['USER_PARAMS'] field added
- ['spec']['strategy']['custom_strategy']['env']['ATOMIC_REACTOR_PLUGINS'] field removed. atomic-reactor generates the plugin configuration from the USER_PARAMS and REACTOR_CONFIG config map at run time.
- most plugin arguments are retrieved at runtime from the REACTOR_CONFIG config map at run time.

PR #746, opened 2018-04-03
Set architecture parameter for add_filesystem

lcarva:
This parameter is used by add_filesystem plugin to select the arch
specific filesystem tarball.


twaugh:
Updated draft release notes

PR #748, opened 2018-04-09
fixes for arrangement 6

mlangsdorf:
A handful of fixes for arrangement 6
* add koji_upload_dir to BuildUserParams and koji_upload's rendered args
* remove build_json_dir from the args of some plugins
* change the way that labels are passed to add_labels_in_df

works with https://github.com/projectatomic/atomic-reactor/pull/964

### DIFF
--- a/tests/build_/test_plugins_configuration.py
+++ b/tests/build_/test_plugins_configuration.py
@@ -115,7 +115,6 @@ class TestPluginsConfiguration(object):
     def test_render_koji_upload(self, build_type):
         user_params = get_sample_user_params({'koji_upload_dir': 'test'},
                                              build_type=build_type)
-        user_params = get_sample_user_params(build_type=build_type)
         build_json = PluginsConfiguration(user_params).render()
         plugins = get_plugins_from_build_json(build_json)
         if build_type == BUILD_TYPE_WORKER:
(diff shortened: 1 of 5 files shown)


tests/build_/test_plugins_configuration.py
```diff
@@ -104,6 +104,7 @@ def assert_koji_upload_plugin(self, plugins):
 
         assert get_plugin(plugins, phase, name)
         plugin_args = plugin_value_get(plugins, phase, name, 'args')
+        assert plugin_args.get('koji_upload_dir')
```
lcarva: Maybe verify that it's actually the expected value?

lcarva: Maybe verify it's set to the expected value?

tests/build_/test_plugins_configuration.py
```diff
@@ -104,7 +104,7 @@ def assert_koji_upload_plugin(self, plugins):
 
         assert get_plugin(plugins, phase, name)
         plugin_args = plugin_value_get(plugins, phase, name, 'args')
-        assert plugin_args.get('build_json_dir')
+        assert plugin_args.get('koji_upload_dir')
```
lcarva: Maybe verify it's set to the expected value?

* Duplicating this comment because the previous one got hidden along the way...

mlangsdorf: Fixed.

twaugh:

PR #750, opened 2018-04-11
more platform 6 fixes

mlangsdorf:
koji_import gets rendered properly
platforms is an optional argument in arrangement6 and create_orchestrate_build shouldn't raise an error if it is omitted.

### DIFF
--- a/osbs/build/plugins_configuration.py
+++ b/osbs/build/plugins_configuration.py
@@ -394,10 +394,6 @@ class PluginsConfiguration(object):
         if not self.pt.has_plugin_conf(phase, plugin):
             return
 
-        if self.user_params.platforms.value is None:
-            self.pt.remove_plugin(phase, plugin, 'no platforms in user parameters')
-            return
-
         # Parameters to be used in call to create_worker_build
         worker_params = [
             'component', 'git_branch', 'git_ref', 'git_uri', 'koji_task_id',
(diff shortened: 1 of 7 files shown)


osbs/build/plugins_configuration.py
```diff
@@ -363,6 +363,15 @@ def render_inject_parent_image(self):
 
         self.pt.set_plugin_arg(phase, plugin, 'koji_parent_build', koji_parent_build)
 
+    def render_koji_import(self, use_auth=None):
+        if not self.pt.has_plugin_conf('exit_plugins', 'koji_import'):
+            return
+
+        koji_target = self.user_params.koji_target.value
+        if not self.pt.set_plugin_arg_valid('exit_plugins', 'koji_import', 'target', koji_target):
```
lcarva: Looking at `koji_import` plugin, I don't think `target` parameter is used at all.

tests/test_api.py
```diff
@@ -530,8 +530,8 @@ def test_create_worker_build_ioerror(self, osbs):  # noqa
         )
     )
     @pytest.mark.parametrize(('platforms', 'raises_exception'), (
```
lcarva: Any reason to keep `raises_exception` parameter if it's always `False`?

osbs/build/plugins_configuration.py
mlangsdorf: Fixed, or at least, I'm not doing anything to render koji_import anymore.

tests/test_api.py
mlangsdorf: Fixed.

lcarva:

PR #754, opened 2018-04-16
Don't backup Build objects

lcarva:
The information required by OSBS is fully stored in BuildConfig and
ImageStream objects. Additionally, the Build objects are only restored
if they don't have a BuildConfig associated with them. Those are scratch
and isolated builds which we care very little about backing up in OSBS.


lcarva:
Release notes draft updated
> - Backup builder command no longer backs up `Build` OpenShift objects

PR #757, opened 2018-04-20
BuildRequestV2 should get outer template as param

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -199,6 +199,7 @@ class OSBS(object):
         else:
             build_request = BuildRequestV2(
                 build_json_store=self.os_conf.get_build_json_store(),
+                outer_template=outer_template,
                 customize_conf=customize_conf)
 
         # Apply configured resource limits.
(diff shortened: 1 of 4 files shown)


osbs/api.py
```diff
@@ -362,7 +363,7 @@ def _get_image_stream_info_for_build_request(self, build_request):
         image_stream_tag_name = None
 
         if build_request.has_ist_trigger():
-            image_stream_tag_id = build_request.spec.trigger_imagestreamtag.value
+            image_stream_tag_id = build_json['spec']['triggers'][0]['imageChange']['from']['name']
```
lcarva: Why this change?

rcerven: buildrequest2 has empty spec

lcarva: Good call! Maybe this should be a helper method in BuildRequest? BuildRequestV2 inherits from it so it should be available there too.

osbs/build/build_requestv2.py
```diff
@@ -22,12 +22,13 @@ class BuildRequestV2(BuildRequest):
     Wraps logic for creating build inputs
     """
 
-    def __init__(self, build_json_store, customize_conf=None):
+    def __init__(self, build_json_store, outer_template=None, customize_conf=None):
         """
         :param build_json_store: str, path to directory with JSON build files
```
sosiouxme: Can there be a matching `:param` comment?

        :param outer_template: str, path to outer template JSON

rcerven: will add

lcarva:
Updated release notes draft:
> - `BuildRequestV2` sets correct outer template for orchestrator and worker builds

PR #758, opened 2018-04-23
ignore reactor_config_map and override for arrangement 5 and lower

rcerven:


osbs/build/build_requestv2.py
```diff
@@ -130,6 +133,74 @@ def adjust_for_scratch(self):
             self.template['spec'].pop('triggers', None)
             self.set_label('scratch', 'true')
 
+    def set_reactor_config(self, reactor_config_map=None, reactor_config_override=None):
```
lcarva: This method and `set_required_secrets` take these parameters because they used to handle values from either `UserParams` or `BuildSpec`. Since this is no longer the case, can we remove them and simply reach into `self.user_params` directly?

rcerven: sure

lcarva:
Updated release notes draft:
> - `reactor_config_map` and `reactor_config_override` are ignored for arrangement 5 and lower

PR #762, opened 2018-05-11
core: don't loop forever waiting for a response

mlangsdorf:
core: don't loop forever waiting for a response
    
watch_resource() has an infinite loop until success or a bad http response.
If a valid http response is returned, but the response returns an empty
iterator, watch_resource will loop forever.
    
Change it to be a bounded loop that will return to wait() in 5 minutes. wait()
will then raise an error that is caught by wait_for_build_to_finish() or
wait_for_build_to_get_scheduled().
    
The current total timeout is 12 hours - 144 retries in wait_for_build_*, each
of which takes 5 minutes.
    
### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -35,6 +35,13 @@ from .http import HttpSession
 logger = logging.getLogger(__name__)
 
 
+# Retry each connection attempt after 30 seconds, for a maximum of 10 times
+WATCH_RETRY_SECS = 30
+WATCH_RETRY = 10
+# Give up after 12 hours
+WAIT_RETRY_HOURS = 12
+WAIT_RETRY = int(WAIT_RETRY_HOURS * 3600 / (WATCH_RETRY_SECS * WATCH_RETRY))
+
 def check_response(response, log_level=logging.ERROR):
     if response.status_code not in (http_client.OK, http_client.CREATED):
         if hasattr(response, 'content'):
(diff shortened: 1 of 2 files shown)


osbs/core.py
```diff
@@ -563,31 +563,32 @@ def watch_resource(self, resource_type, resource_name=None, **request_args):
             path += "%s/" % resource_name
         url = self._build_url(path, _prepend_namespace=False, **request_args)
 
-        while True:
-            with self._get(url, stream=True, headers={'Connection': 'close'}) as response:
-                check_response(response)
-                encoding = None
-                for line in response.iter_lines():
-                    logger.debug(line)
+        for retry in range(1, 10):
```
twaugh: Probably should be `range(10)`: len(range(1,10)) == 9 but len(range(10)) == 10.

osbs/core.py
```diff
@@ -630,24 +631,29 @@ def wait(self, build_id, states):
...
         #   2. our object was not found and we keep waiting (in the loop)
         # Therefore, let's raise here
-        logger.error("build '%s' was not found during wait", build_id)
+        logger.warning("build '%s' was not found during wait", build_id)
         raise OsbsWatchBuildNotFound("build '%s' was not found and response stream ended" %
                                      build_id)
 
     def wait_for_build_to_finish(self, build_id):
-        for retry in range(1, 10):
+        for retry in range(1, 144):
```
twaugh: Probably should be `range(144)` (pre-existing off-by-one).

osbs/core.py
```diff
@@ -630,24 +631,29 @@ def wait(self, build_id, states):
...
                 # this is woraround for https://github.com/openshift/origin/issues/2348
-                logger.error("I'm going to wait again. Retry #%d.", retry)
+                logger.warning("I'm going to wait again. Retry #%d.", retry)
                 continue
         raise OsbsException("Failed to wait for a build: %s" % build_id)
 
     def wait_for_build_to_get_scheduled(self, build_id):
-        build_response = self.wait(build_id, BUILD_FINISHED_STATES + BUILD_RUNNING_STATES)
-        return build_response
+        for retry in range(1, 144):
```
twaugh: Probably should be `range(144)`.

osbs/core.py
mlangsdorf: Fixed.

osbs/core.py
mlangsdorf: Fixed.

osbs/core.py
mlangsdorf: Fixed.

osbs/core.py
```diff
@@ -563,31 +563,32 @@ def watch_resource(self, resource_type, resource_name=None, **request_args):
...
 
-        while True:
-            with self._get(url, stream=True, headers={'Connection': 'close'}) as response:
-                check_response(response)
-                encoding = None
-                for line in response.iter_lines():
-                    logger.debug(line)
+        for retry in range(10):
+            response = self._get(url, stream=True, headers={'Connection': 'close'})
+            # with self._get(url, stream=True, headers={'Connection': 'close'}) as response:
```
twaugh: No longer need this line.

osbs/core.py
```diff
@@ -630,24 +631,29 @@ def wait(self, build_id, states):
...
     def wait_for_build_to_get_scheduled(self, build_id):
-        build_response = self.wait(build_id, BUILD_FINISHED_STATES + BUILD_RUNNING_STATES)
-        return build_response
+        for retry in range(144):
+            try:
+                build_response = self.wait(build_id, BUILD_FINISHED_STATES + BUILD_RUNNING_STATES)
+                return build_response
+            except OsbsWatchBuildNotFound:
+                continue
+            raise OsbsException("Failed to schedule a build in 144 attempts: %s" % build_id)
```
twaugh: Worth putting 144 in a variable and using it both in the range() and in this exception message?

tests/test_core.py
```diff
@@ -168,6 +169,32 @@ def test_get_user(self, openshift):  # noqa
         l = openshift.get_user()
         assert l.json() is not None
 
+    def test_watch_resource_and_wait_to_build_timeouts(self, caplog, openshift):  # noqa
```
twaugh: Is this `# noqa:F811`? Although a lot of places say '# noqa', it would be nicer to explicitly silent the particular diagnostic we know about.

osbs/core.py
lcarva: Why 144 btw?

mlangsdorf: watch resources will sleep for 30 seconds in each of 10 loops, so each invocation of watch_resource takes 5 minutes. looping 12 times can take an hour. looping 144 times (12 * 12) is 12 hours.

I guessed that most builds would be scheduled or finished within 12 hours.

twaugh:
Updated draft release notes:
- When watching a resource on the server, do not wait forever

PR #763, opened 2018-05-11
HttpResponse: better error message when json is corrupt

mlangsdorf:
Currently, if the json string received from a _get is corrupt and
can't be converted, the json package raises a cryptic message about
extra data, but doesn't provide the bad string.
    
Catch that exception and raise an error with more detail about what
went wrong and the full string that was being loaded.
    
This doesn't fix osbs-5424, but will hopefully make it easier to
determine what went wrong the next time someone sees a similar failure.
    
### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -12,8 +12,8 @@ abstraction on top of http api calls
 from __future__ import print_function, absolute_import, unicode_literals
 
 import sys
-import json
 import logging
+import json
 from six.moves import http_client
 
 
@@ -212,4 +212,10 @@ class HttpResponse(object):
         if check and self.status_code not in (0, requests.codes.OK, requests.codes.CREATED):
             raise OsbsResponseException(text, self.status_code)
 
-        return json.loads(text)
+        try:
+            return json.loads(text)
+        except ValueError:
+            msg = '{0}Headers {1}\nContent {2}'.format('HtttpResponse has corrupt json:\n',
+                                                       self.headers, self.content)
+            logger.exception(msg)
+            raise OsbsResponseException(msg, self.status_code)
--- a/tests/test_http.py
+++ b/tests/test_http.py
@@ -13,7 +13,7 @@ import pytest
 import requests
 
 from requests.packages.urllib3.util import Retry
-from osbs.http import HttpSession, HttpStream, http_client
+from osbs.http import HttpSession, HttpStream, http_client, HttpResponse
 from osbs.exceptions import OsbsNetworkException, OsbsException, OsbsResponseException
 from osbs.constants import HTTP_RETRIES_STATUS_FORCELIST, HTTP_REQUEST_TIMEOUT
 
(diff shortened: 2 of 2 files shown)


tests/test_http.py
```diff
@@ -215,3 +216,18 @@ def test_set_timeout(self):
             .once())
 
         HttpStream(url, method, verify_ssl=False, **kwargs)
+
+
+class TestHttpResponse(object):
+    def test_simple_response(self):
+        content = 'this is content'
```
twaugh: I think the content should be a 'bytes' instance, e.g. `content = b'this is content'`. This is the difference between requests.Response.content and requests.Response.text -- 'text' has been decoded according to the media type header. (Note that for json we have to decode 'content' separately using guess_json_utf() because the media type header application/json gives no clue about encoding.)

osbs/http.py
```diff
@@ -212,4 +212,10 @@ def json(self, check=True):
         if check and self.status_code not in (0, requests.codes.OK, requests.codes.CREATED):
             raise OsbsResponseException(text, self.status_code)
 
-        return json.loads(text)
+        try:
+            return json.loads(text)
+        except ValueError:
+            msg = '{0}Headers {1}\nContent {2}'.format('HtttpResponse has corrupt json:\n',
```
lcarva: Seems odd to have the literal string being formatted to msg. Can this be changed to:
```python
msg = 'HtttpResponse has corrupt json:\nHeaders {0}\nContent {1}'.format(self.headers, self.content)
```

osbs/http.py
```diff
@@ -212,4 +212,10 @@ def json(self, check=True):
...
+                                                       self.headers, self.content)
+            logger.error(msg)
```
lcarva: How about changing `logger.error` to `logger.exception` so the originating stack trace is not lost?

tests/test_http.py
lcarva: This is likely the reason why tests failed in python 3.

osbs/http.py
mlangsdorf: It could be, but then the line break is at column 92 and self.headers, self.content are formatted ugly.

osbs/http.py
mlangsdorf: Great idea. Fixed.

tests/test_http.py
```diff
@@ -215,3 +215,19 @@ def test_set_timeout(self):
...
+        content_str = "this is content"
+        response = HttpResponse(status_code=http_client.OK, headers={}, content=content_json)
+        assert content_str == response.json()
+
+    def test_bad_coding_guess(self):
+        content_str = "this is content"
+        bad_json = b'[\"cat\", \"dog\"][\"cat\", \"dog\"]'
+        response = HttpResponse(status_code=http_client.OK, headers={}, content=bad_json)
+        with pytest.raises(OsbsResponseException) as exc_info:
+            content_str == response.json()
```
twaugh: This looks wrong. Do we need content_str at all?

twaugh:
Updated draft release notes:
- Report a better error message when JSON received in an HTTP response is corrupt

PR #767, opened 2018-05-22
HttpStream: don't catch exceptions in iter_lines

mlangsdorf:
Remove the try/except sequence from iter_lines(), so that any exceptions
generated by the req.iter() will be propagatetd unchanged up the stack.

Add a test case to prove that an exception in HttpStream.iter_lines()
will move up the stack.


lcarva:
Updated release notes draft:
> - Allow exceptions from `HttpStream.iter_lines` to propagate

PR #768, opened 2018-05-22
remove OpenShift 0.5.4 support

mlangsdorf:
There's not much openShift 0.5.4 support left, but the double set of mock_json responses means that tests take twice as long to run and are slightly harder to support.


twaugh:
Not updating release notes because this is a change to the unit tests only.

PR #771, opened 2018-05-29
distinguish similar git repos in different locations

mlangsdorf:
fixes OSBS-5592

currently, host.com/stage/repo.git and host.com/prod/repo.git resolve to
the same BuildConfig, and this breaks some people's workflows. Add an
optional git-dir-name label to BuildConfigs. If it doesn't exist, operate
as before. If it does exist, consider two BuildConfigs to be the same if
they have the same name and the same git-dir-name, git-repo-name, and
git-branch labels.


osbs/api.py
```diff
@@ -274,28 +274,52 @@ def _get_existing_build_config(self, build_config):
...
+                git_labels.append((key, build_config['metadata']['labels'][key]))
+            # skip missing labels
+            except KeyError:
+                if key == 'git-dir-name':
+                    ignore_dirname = True
+                continue
         name = build_config['metadata']['name']
 
         queries = (
             (self.os.get_build_config_by_labels, git_labels),
```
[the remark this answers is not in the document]
lcarva: I see. How about if `get_build_config_by_labels` was modified to perform this check based on the BuildConfig parameter at `spec.source.git.uri`? For instance, get all build configs that match repo and branch labels, then check `spec.source.git.uri` to filter out based on expected git dir name.

We may also have to consider a different algo for naming build configs.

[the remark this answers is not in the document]
mlangsdorf: arrangement6 doesn't use spec and doesn't have a spec.source.git.uri.

but filtering in get_build_config_by_labels is more elegant. let me take a stab at that.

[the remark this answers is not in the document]
lcarva: `spec.source.git.uri` is definitely set for arrangement 6 as well: https://github.com/projectatomic/osbs-client/blob/master/osbs/build/build_requestv2.py#L221-L222

Note that `spec` in this case is not osbs-client's `BuildSpec` object.

osbs/api.py
```diff
@@ -274,8 +274,15 @@ def _get_existing_build_config(self, build_config):
         - metadata.labels.git-repo-name AND metadata.labels.git-branch are equal
```
twaugh: Can we update the docstring for this method?

osbs/api.py
```diff
@@ -274,8 +274,15 @@ def _get_existing_build_config(self, build_config):
...
         """
 
-        git_labels = [(key, build_config['metadata']['labels'][key])
-                      for key in self._GIT_LABEL_KEYS]
+        git_labels = {
+            'repo_labels': [(key, build_config['metadata']['labels'][key])
+                            for key in self._GIT_LABEL_KEYS]
+        }
+        try:
+            git_labels['git-dir-name'] = build_config['metadata']['labels']['git-dir-name']
```
twaugh: Are we happy with 'git-dir-name'? Not, say, 'git-namespace'? (I have no preference)

Also, do we want to put it in osbs/constants.py instead of having multiple string literals in the code for it?

osbs/api.py
```diff
@@ -274,8 +274,15 @@ def _get_existing_build_config(self, build_config):
...
+        except KeyError:
+            git_labels['git-dir-name'] = None
```
twaugh: Can just use `git_labels[...] = build_config['metadata']['labels'].get(...)` in that case, no need for try:/except:.

osbs/core.py
```diff
@@ -320,12 +321,13 @@ def get_build_config(self, build_config_id):
         build_config = response.json()
         return build_config
 
-    def get_build_config_by_labels(self, label_selectors):
+    def get_build_config_by_labels(self, label_data):
         """
         Returns a build config matching the given label
         selectors. This method will raise OsbsException
```
twaugh: Can we update the docstring for this method?

mlangsdorf:
since we're not changing the actual name, I think old configs will continue to be found by name after failing the labels test.

sosiouxme:
On Fri, Jun 1, 2018 at 7:31 AM, Mark Langsdorf <notifications@github.com>
wrote:

> since we're not changing the actual name, I think old configs will
> continue to be found by name after failing the labels test.
>

Again - no need to go deep if I'm just not getting it yet - but, if old
build configs continue to be found, what prompts osbs-client to create a
new one? Say I already have a BC for stage/foorepo, this change goes into
effect, now I build prod/foorepo -- does it build in the old BC or create a
new one?

osbs/build/build_request.py
```diff
@@ -1506,12 +1507,17 @@ def render(self, api=None, validate=True):
...
+        # Attempt to handle multiple instances of repos with the same name
+        # by tracking the directory they are in.
         # NOTE: Since only the repo name is used, a forked repos will have
         # the same git-repo-name tag. This is a known limitation. If this
         # use case must be handled properly, the git URI must be taken into
         # account.
         self.set_label('git-repo-name', repo_name)
         self.set_label('git-branch', self.spec.git_branch.value)
+        if repo_dirname:
+            self.set_label('git-dir-name', repo_dirname)
```
lcarva: To avoid ambiguity, can we fallback `git-dir-name` to an empty string if a `repo_dirname` is not set?

lcarva:
> Again - no need to go deep if I'm just not getting it yet - but, if old
> build configs continue to be found, what prompts osbs-client to create a
> new one? Say I already have a BC for stage/foorepo, this change goes into
> effect, now I build prod/foorepo -- does it build in the old BC or create a
> new one?

A new BuildConfig is only created if not found by the query methods. Otherwise,
they are updated.

It would be ideal to perform the search as this:
1. Find build config by all 3 labels
2. If not found, find build config by repo name and branch labels, and use `spec.source.git.uri` to match on dir name. If dir name doesn't match, it also reports as not found and the next step is tried.
3. If not found, match by name. This is a legacy method of finding build configs, nothing should still require it. We can probably remove this.
4. If still not found, then it doesn't exist.

Then during update/create of build config all three labels are always set. This will cause the first method above to find the Build Config next time. In the future, we can also retired the second method of finding build config.

To make the search steps work more smoothly, I suggest always adding a dir name label even when there isn't a dir (use empty string, or some other marker). 

Thoughts?

mlangsdorf:
So currrently the code is doing:
1. Match on git-repo-name and git-branch labels
2. Filter against spec.source.git.uri (essentially) as another label, if more than one match is found
3. if no match on the labels, search by name.
4. if not found, then no existing config and we can build a new one.

I might need to add another commit to always update the filter label.

lcarva:
> Filter against spec.source.git.uri (essentially) as another label, if more than one match is found

I think that needs to be done even if a single one is found.

mlangsdorf:
so if we have two repos, each on the master branch
http://example.com/stage/oops.git, commit 12345678
http://example.com/prod/oops.git commit 98765432
the current (prior to this patch release) existing config from built from prod/oops.git
a build is launched for stage/oops.git after this patch is release:
First check - do all labels match? no, existing json doesn't have a full-repo-name label
Second check - do the old labels match AND is spec.source.git.uri a match? no, the old labels match but the source.git.uri is different (I think?)
Third check - do the names match? yes, master-oops is the same as master-oops
Update the existing config arguments, so now the existing config gets updated to match stage/oops.
another build is launched, this time for prod/oops.
First check - do all labels match? no, both have full-repo-names labels that are different.
Second check - do old labels match AND is spec.source.git.uri a match? no, the old labels match but the source.git.uri is different (I think?)
Third check - do the names match? yes, master-oops is the same as master-oops.

am I missing something here?

also, why are matching on the symbolic branch name which is non-unique across forks, instead of on the commit hash, which is nearly unique? Or I am misunderstanding the logic again?

lcarva:
> the current (prior to this patch release) existing config from built from prod/oops.git
> a build is launched for stage/oops.git after this patch is release

No, it just prevents them both from running at the same time. Although they currently map to the same BuildConfig, before each one runs, the BuildConfig is fully updated to point all data points to the requested repository.


> am I missing something here?

We need to take dir-name into account when generating build config names. Otherwise, multiple BuildConfigs for the same repo-name and branch but different dir-name still won't be allowed.

I think it's fine to change the code that generates the name of the BuildConfig. It's only used for querying as a last resort (which we should eventually remove btw). If a BuildConfig is found, its name is not updated, so existing BuildConfigs won't change their names unexpectedly and they won't be duplicated by accident.



> why are matching on the symbolic branch name which is non-unique across forks, instead of on the commit hash, which is nearly unique? Or I am misunderstanding the logic again?

The point of a BuildConfig is to group builds for a given branch. Using the commit hash in this case won't give us the desired result. We currently have the limitation that the same git server is used for all builds. It seems beyond the scope of this change.

osbs/utils.py
```diff
@@ -259,7 +260,7 @@ def git_repo_humanish_part_from_uri(git_uri):
     elif git_uri.endswith(".git"):
         git_uri = git_uri[:-4]
 
-    return os.path.basename(git_uri)
+    return urlparse(git_uri)[2].lstrip('/')
```
lcarva: `urlparse` returns a named tuple, maybe use `.path` instead of `[2]`?

But! Should we actually do this? The length of build config names are quite short, this will cause a bunch of BuildConfigs to start with either rpms/... or containers/... which seems wasteful and not very intuitive.

I'm thinking that this shouldn't change at all. The `git_repo_humanish_part_from_uri` is used to determine the value for label `git-repo-name`. If we change processing of this value, it means that on the next code release, osbs-client won't be able to find **any** existing BuildConfigs.

Instead, we should change `make_name_from_git` to take the full repo name into account. Maybe just for generating the hash?

mlangsdorf: You were the one who said to change the names. And didn't we just have a discussion about the fact that we get existing configs by labels, not names?
It's easy enough to change the hash code if that's what we want to do.
```
-     full = repo + branch
+    full = urlparse(repo).path.lstrip('/') + branch
```

osbs/build/build_requestv2.py
```diff
@@ -234,14 +234,14 @@ def render(self, validate=True):
...
-        # the same git-repo-name tag. This is a known limitation. If this
-        # use case must be handled properly, the git URI must be taken into
-        # account.
+        # Set git-repo-name and git-dir-name labels
         repo_name = git_repo_humanish_part_from_uri(self.user_params.git_uri.value)
+        # Use the repo name to differentiate different repos, but include the full url as an
+        # optional filter.
         self.set_label('git-repo-name', repo_name)
         self.set_label('git-branch', self.user_params.git_branch.value)
+        self.set_label('git-full-repo', self.user_params.git_uri.value)
```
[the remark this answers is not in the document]
mlangsdorf: it's matching the value in spec.source.git.uri right now. I think that's fine since we're not using it to build the name.

osbs/utils.py
lcarva: I suggested changing the generated BuildConfig name, not the `git-repo-name` label.

osbs/core.py
```diff
@@ -342,6 +343,35 @@ def get_build_config_by_labels(self, label_selectors):
...
         return items[0]
 
+    def get_build_config_by_labels_filtered(self, label_data):
+        """
+        Returns a build config matching the given label
+        selectors. This method will raise OsbsException
+        if not exactly one build config is found.
+        """
+        label_selectors = label_data['search_labels']
+        labels = ['%s=%s' % (field, value) for field, value in label_selectors]
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

osbs/utils.py

osbs/api.py
```diff
@@ -269,17 +270,27 @@ def _get_existing_build_config(self, build_config):
...
+          metadata.spec.source.git.uri are equal
         OR
-        - metadata.labels.git-repo-name AND metadata.labels.git-branch are equal
+        - metadata.name are equal
         """
 
-        git_labels = [(key, build_config['metadata']['labels'][key])
-                      for key in self._GIT_LABEL_KEYS]
+        bc_labels = build_config['metadata']['labels']
+        git_labels = [(key, bc_labels[key]) for key in self._OLD_LABEL_KEYS]
```
lcarva: You want `_GIT_LABEL_KEYS` here, right?

rcerven:
I think we can just use basically 2 queries
1: label:branch & spec.git.uri (which includes that dir)
2: build config name

3 labels query doesn't give us any benefit

as all we need is just repo name with dir (spec.git.uri) & branch

osbs/api.py

osbs/constants.py
```diff
@@ -105,3 +105,6 @@
 
 ANNOTATION_SOURCE_REPO = 'osbs/source_repo'
 ANNOTATION_INSECURE_REPO = 'openshift.io/image.insecureRepository'
+
+# optional key path for filtering existing build config results
+FILTER_KEY = 'spec.source.git.ref'
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

osbs/core.py
```diff
@@ -320,21 +321,54 @@ def get_build_config(self, build_config_id):
...
-        labels = ','.join(labels)
-        url = self._build_url("buildconfigs/", labelSelector=labels)
-        items = self._get(url).json()['items']
+        items = self.get_all_build_configs_by_labels(label_selectors)
+
+        if not items:
+            raise OsbsException(
+                "Build config not found for labels: %r" %
+                (label_selectors, ))
+        if len(items) > 1:
```
rcerven: this is missing coverage

osbs/core.py
```diff
@@ -320,21 +321,54 @@ def get_build_config(self, build_config_id):
...
         if not items:
             raise OsbsException(
                 "Build config not found for labels: %r" %
                 (label_selectors, ))
+        if match_filter is not None:
+            for build_config in items:
+                filter_value = graceful_chain_get(build_config, *FILTER_KEY.split('.'))
+                if filter_value == match_filter:
+                    return build_config
         if len(items) > 1:
```
rcerven: this is missing coverage, also 364, and 377

lcarva: Let's make sure this is taken care of before merging.

osbs/utils.py
```diff
@@ -361,7 +362,7 @@ def make_name_from_git(repo, branch, limit=53, separator='-', hash_size=5):
 
     repo = git_repo_humanish_part_from_uri(repo)
     branch = branch or 'unknown'
-    full = repo + branch
+    full = urlparse(repo).path.lstrip('/') + branch
```
twaugh: This looks odd. At this point, `repo` came from `git_repo_humanish_part_from_uri()`, which returns the result from a call to `os.path.basename()`. Why would there be '/' characters at the beginning of that?

tests/constants.py
```diff
@@ -24,6 +24,7 @@
 TEST_GIT_URI = "git://hostname/path"
 TEST_GIT_URI_HUMAN_NAME = "path"
 TEST_GIT_REF = "0123456789012345678901234567890123456789"
+TEST_GIT_URI_HTML = "git%3A%2F%2Fhostname%2Fpath"
```
twaugh: Or, could use this:
```
from six.moves.urllib.parse import quote as urlquote
urlquote(TEST_GIT_URI, safe='')
```

tests/test_core.py
```diff
@@ -254,6 +254,34 @@ def test_get_build_config_by_labels(self, openshift):  # noqa
         response = openshift.get_build_config_by_labels(label_selectors)
         assert response['spam'] == 'maps'
 
+    def test_get_build_config_by_labels_filtered(self, openshift):  # noqa
```
twaugh: Could specify the particular reason we're turning off diagnostics: `# noqa:F811`

osbs/api.py
```diff
@@ -269,17 +270,27 @@ def _get_existing_build_config(self, build_config):
...
-        - metadata.labels.git-repo-name AND metadata.labels.git-branch are equal
+        - metadata.name are equal
         """
 
-        git_labels = [(key, build_config['metadata']['labels'][key])
-                      for key in self._GIT_LABEL_KEYS]
+        bc_labels = build_config['metadata']['labels']
+        git_labels = [(key, bc_labels[key]) for key in self._GIT_LABEL_KEYS]
+        old_labels = {
+            'search_labels': [(key, bc_labels[key]) for key in self._OLD_LABEL_KEYS],
```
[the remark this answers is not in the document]
mlangsdorf: Thanks for the suggestion. Fixed.

osbs/utils.py

tests/constants.py
mlangsdorf: Fixed.

tests/test_core.py
mlangsdorf: Fixed.

osbs/core.py
mlangsdorf: Test coverage added.

tests/constants.py
twaugh: Seems like that causes problems on CentOS 6 with Python 2.6, not sure why. I'm fine with putting it back how it was, although it might be better if it was the line next to the definition of TEST_GIT_URI.

mlangsdorf: Fixed again, along with a comment as to why the code isn't using the urlparse lib

mlangsdorf:
added to RN under API changes:
- builds with the same git name but different directory paths will receive different labels and hashes, allowing them to be distinguished. existing build configs should continue to work.

sosiouxme:
I suspect this is what broke https://osbs-stage.upshift.engineering.redhat.com/console/project/osbs-stage/browse/builds/osbs-test-sandwich-docker-private-balkov-test-c47f6/osbs-test-sandwich-docker-private-balkov-test-c47f6-68?tab=logs

mlangsdorf:
Labels have to be sanitized, the same as BuildConfigIDs. makes sense.
I will open another PR shortly to fix this.

mlangsdorf:
#777

PR #772, opened 2018-06-01
conf: quieter deprecation warnings

mlangsdorf:
only print deprecation warnings for cli args and file args, not for kwargs.

### DIFF
--- a/osbs/conf.py
+++ b/osbs/conf.py
@@ -60,7 +60,8 @@ class Configuration(object):
         self.args = cli_args
         self.kwargs = kwargs
 
-    def _get_value(self, args_key, conf_section, conf_key, default=None, is_bool_val=False):
+    def _get_value(self, args_key, conf_section, conf_key, default=None, is_bool_val=False,
+                   deprecated=False):
         # FIXME: this is too bloated: split it into separate classes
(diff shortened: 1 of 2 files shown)


lcarva:
What's the motivation for not printing warning for kwargs; that means no warnings when build is started via koji-containerbuild which is our main use case.

I do think the change to only emit warning if there's a value set is a very good improvement.

mlangsdorf:
from looking at the test cases, it looked like kwargs were mainly internal/default values, while cli-args and file arguments where things that people actually set and thus needed a deprecation warning.

If the actual use case is to grab the args in another program and pass them as kwargs, I can change it to always print a message for found values.

lcarva:
> If the actual use case is to grab the args in another program and pass them as kwargs, I can change it to always print a message for found values.

Looking at the k-cb code, I realize now that the kwargs args used there are mostly user parameters that are not deprecated. Everything else comes from `/etc/osbs.conf`. My bad, your code changes make a lot of sense to me now. Let's keep it as it is.

https://github.com/release-engineering/koji-containerbuild/blob/master/koji_containerbuild/plugins/builder_containerbuild.py#L517-L556

PR #774, opened 2018-06-05
build response: better error messages

mlangsdorf:
add a new function to BuildReponse that returns the error reason: either the error message
from the build metadata if that exists, or from the pod's failure reason if that exists.

this new function will also be called by atomic reactor's build_orcestrate_build plugin, to
replace the current code for getting failures from a worker build.

change get_error_message() to use this new function also.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -135,7 +135,7 @@ class OSBS(object):
         serialized_response = response.json()
         build_list = []
         for build in serialized_response["items"]:
-            build_list.append(BuildResponse(build))
+            build_list.append(BuildResponse(build, self))
 
         return build_list
 
@@ -150,13 +150,13 @@ class OSBS(object):
     @osbsapi
     def get_build(self, build_id):
         response = self.os.get_build(build_id)
-        build_response = BuildResponse(response.json())
+        build_response = BuildResponse(response.json(), self)
         return build_response
 
     @osbsapi
     def cancel_build(self, build_id):
         response = self.os.cancel_build(build_id)
-        build_response = BuildResponse(response.json())
+        build_response = BuildResponse(response.json(), self)
         return build_response
 
     @osbsapi
@@ -228,14 +228,14 @@ class OSBS(object):
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
         build = build_request.render()
         response = self.os.create_build(json.dumps(build))
-        build_response = BuildResponse(response.json())
+        build_response = BuildResponse(response.json(), self)
         return build_response
 
     def _get_running_builds_for_build_config(self, build_config_id):
         all_builds_for_bc = self.os.list_builds(build_config_id=build_config_id).json()['items']
         running = []
         for b in all_builds_for_bc:
-            br = BuildResponse(b)
+            br = BuildResponse(b, self)
             if br.is_pending() or br.is_running():
                 running.append(br)
         return running
@@ -367,7 +367,7 @@ class OSBS(object):
                 raise RuntimeError('Matching build(s) already running: {0}'
                                    .format(', '.join(x.get_build_name() for x in running_builds)))
 
-        return BuildResponse(self.os.create_build(build_json).json())
+        return BuildResponse(self.os.create_build(build_json).json(), self)
 
     def _get_image_stream_info_for_build_request(self, build_request):
         """Return ImageStream, and ImageStreamTag name for base_image of build_request
@@ -480,10 +480,10 @@ class OSBS(object):
             prev_version = existing_bc['status']['lastVersion']
             build_id = self.os.wait_for_new_build_config_instance(
                 build_config_name, prev_version)
-            build = BuildResponse(self.os.get_build(build_id).json())
+            build = BuildResponse(self.os.get_build(build_id).json(), self)
         else:
             response = self.os.start_build(build_config_name)
-            build = BuildResponse(response.json())
+            build = BuildResponse(response.json(), self)
 
         return build
 
@@ -909,13 +909,13 @@ class OSBS(object):
     @osbsapi
     def wait_for_build_to_finish(self, build_id):
         response = self.os.wait_for_build_to_finish(build_id)
-        build_response = BuildResponse(response)
+        build_response = BuildResponse(response, self)
         return build_response
 
     @osbsapi
     def wait_for_build_to_get_scheduled(self, build_id):
         response = self.os.wait_for_build_to_get_scheduled(build_id)
-        build_response = BuildResponse(response)
+        build_response = BuildResponse(response, self)
         return build_response
 
     @osbsapi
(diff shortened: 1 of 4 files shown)


osbs/build/build_response.py
```diff
@@ -106,17 +107,39 @@ def get_error_message(self):
...
+            str_metadata = graceful_chain_get(self.get_annotations_or_labels(), "plugins-metadata")
+            metadata_dict = json.loads(str_metadata)
+            plugin, error_message = list(metadata_dict['errors'].items())[0]
+            return {'plugin': [plugin, error_message]}
+        except (ValueError, AttributeError):
+            try:
+                from osbs.api import OSBS
+                from osbs.conf import Configuration
+
+                osbs_conf = Configuration(self.get_build_json_dir())
```
lcarva: Creating new `Configuration` and `OSBS` objects in `BuildResponse` seems to blur the layers of abstraction. Does it makes sense to move get-error-reason-for-build logic to `api.py`?

mlangsdorf: I agree about breaking the abstraction, but moving it up to api.py means a change to osbs-client to add the new API call, and to koji_containerbuild and atomic-reactor to use the new API call.

While breaking the abstraction means that only osbs-client needs to change - though atomic-reactor can be changed to take advantage of the new interface.

mlangsdorf: The only reason I'm creating an OSBS object is get a pod_response. Would it be cleaner to pass the pod into the BuildResponse at creation?

mlangsdorf: Conceptually cleaner; a lot uglier in the test cases. Oh, well.

osbs/build/build_response.py
```diff
@@ -106,17 +107,36 @@ def get_error_message(self):
...
+        try:
+            str_metadata = graceful_chain_get(self.get_annotations_or_labels(), "plugins-metadata")
+            metadata_dict = json.loads(str_metadata)
+            plugin, error_message = list(metadata_dict['errors'].items())[0]
+            return {'plugin': [plugin, error_message]}
+        except (ValueError, AttributeError):
+            if self.pod:
+                try:
+                    return {'pod': self.pod.get_failure_reason()}
+                except OsbsException:
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

osbs/build/build_response.py
lcarva: I think it makes sense.

osbs/api.py
```diff
@@ -465,10 +474,11 @@ def _create_build_config_and_build(self, build_request):
             prev_version = existing_bc['status']['lastVersion']
             build_id = self.os.wait_for_new_build_config_instance(
                 build_config_name, prev_version)
-            build = BuildResponse(self.os.get_build(build_id).json())
+            build = BuildResponse(self.os.get_build(build_id).json(),
+                                  self.get_pod_for_build(build_id))
         else:
             response = self.os.start_build(build_config_name)
-            build = BuildResponse(response.json())
+            build = BuildResponse(response.json(), self.get_pod_for_build_json(response.json()))
```
twaugh: We've only just started the build here, so the pod might very well not yet exist. Won't that raise an exception in that case?

twaugh:
This adds an awful lot of work in some common paths, e.g. list builds. Previously listing builds was a single request; now it's N more requests, scaling with the number of builds!

If we only care about the pod when calling `get_error_message()`, can we not provide a means for that method to fetch the pod on demand? e.g. we could pass it an `OSBS` instance.

osbs/api.py
mlangsdorf: Since its calling get_pod_for_build_json which is a wrapper for get_pod_for_build, it's easy enough to add a try: except: block and return None (which is valid) if the pod doesn't exist.

mlangsdorf:
lucarval didn't want an OSBS object created in the BuildResponse because it broke the abstraction; twaugh would like an OSBS object passed into the BuildResponse to simplify things. Before I go and change the test cases again, can I get a definite statement that passing in an OSBS object is an acceptable solution?

twaugh:
Whatever solution we use, let's make sure we don't hyperinflate the number of requests we send.

osbs/api.py
lcarva: You're right. Do we need a new API method for this?

mlangsdorf: a new API method means that people calling BuildResponse.get_error_message have to pass that error and decide if they also need to call another function. Or if BuildResponse does it, then BuildResponse needs access to an OSBS instance to call the API. 

Based on twaugh's objections, I favor stashing the OSBS instance in the BuildResponse, which can then retrieve the PodResponse iff needed.

twaugh: It could even just be an optional parameter to the `get_error_message` method.

mlangsdorf: I would prefer to not change get_error_message, because that means changing programs that call it. Changing the __init__ function is something that is mostly self-contained within osbs-client.

mlangsdorf: fixed this so that BuildReponse takes an OSBS instance and only retrieves the PodResponse if needed. Code is much cleaner.

osbs/build/build_response.py
```diff
@@ -109,17 +109,34 @@ def get_error_message(self):
...
+            return {'plugin': [plugin, error_message]}
+        except (ValueError, AttributeError):
+            if self.osbs:
+                try:
+                    pod = self.osbs.get_pod_for_build(self.get_build_name())
+                    return {'pod': pod.get_failure_reason()}
+                except OsbsException:
+                    return None
+            else:
+                return None
```
twaugh: Minor: cascading exits can be hard to read. Perhaps restructure like this?:
```
if not self.osbs:
    return None

try:
    ....
except OsbsException:
    return None
```

osbs/build/build_response.py
```diff
@@ -106,17 +109,34 @@ def get_error_message(self):
...
+            metadata_dict = json.loads(str_metadata)
+            plugin, error_message = list(metadata_dict['errors'].items())[0]
+            return {'plugin': [plugin, error_message]}
+        except (ValueError, AttributeError):
+            if self.osbs:
+                try:
+                    pod = self.osbs.get_pod_for_build(self.get_build_name())
+                    return {'pod': pod.get_failure_reason()}
+                except OsbsException:
+                    return None
```
twaugh: This line doesn't get executed during unit testing. `return None` is obviously fine, but without triggering the exception it's not clear we have the name of the exception correctly in scope.

osbs/build/build_response.py
mlangsdorf: Fixed. Also added a message if osbs is unavailable.

osbs/build/build_response.py
mlangsdorf: Fixed.

lcarva:
Updated release notes draft:
> - more accurate error messages for build failures

PR #775, opened 2018-06-06
raise error if release label contains forbidden characters

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -33,7 +33,8 @@ from osbs.constants import (BUILD_RUNNING_STATES, WORKER_OUTER_TEMPLATE,
                             ORCHESTRATOR_CUSTOMIZE_CONF, BUILD_TYPE_WORKER,
                             BUILD_TYPE_ORCHESTRATOR, BUILD_FINISHED_STATES,
                             DEFAULT_ARRANGEMENT_VERSION, REACTOR_CONFIG_ARRANGEMENT_VERSION,
-                            ANNOTATION_SOURCE_REPO, ANNOTATION_INSECURE_REPO, FILTER_KEY)
+                            ANNOTATION_SOURCE_REPO, ANNOTATION_INSECURE_REPO, FILTER_KEY,
+                            RELEASE_LABEL_FORMAT)
 from osbs.core import Openshift
 from osbs.exceptions import (OsbsException, OsbsValidationException, OsbsResponseException,
                              OsbsOrchestratorNotEnabled)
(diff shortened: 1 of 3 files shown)


rcerven:
still no unit tests

twaugh:
Would be nicer to match against a regular express describing a valid release label. For example " " and "-" should also be disallowed, just off the top of my head, and there are going to be plenty of others.

rcerven:
added unit test, and matching characters against regular expression

twaugh:
My preference would be for the regular expression to describe *valid* release labels. Something like: `^[0-9]([._]?[a-zA-Z0-9]+)*$`. Even better if it uses a verbose regexp to describe it.

rcerven:
regex is now matching valid values

osbs/constants.py
```diff
@@ -102,6 +102,9 @@
 BUILD_TYPE_WORKER = "worker"
 
 ISOLATED_RELEASE_FORMAT = re.compile(r'^\d+\.\d+(\..+)?$')
+RELEASE_LABEL_FORMAT = re.compile(r"""^\d+             # First character must be a digit
+                                      ([._]?          # allow separators between groups
+                                      [a-zA-Z0-9]+)*$  # last characters must be alphanumeric""", re.X)
```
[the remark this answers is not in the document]
twaugh: Could shorten line lengths:
```
RELEASE_LABEL_FORMAT = re.compile(r"""
    ^\d+ ...
""", re.X)

File: osbs/api.py [hunk 2]
```diff
@@ -491,6 +492,14 @@ def _check_labels(self, repo_info):
                 logger.error("required label missing from Dockerfile : %s",
                              labels.get_name(label))
 
+        try:
+            _, release_value = labels.get_name_and_value(utils.Labels.LABEL_TYPE_RELEASE)
+            if not RELEASE_LABEL_FORMAT.match(release_value):
+                logger.error("release label contains forbidden characters : %s", release_value)
+                raise OsbsValidationException("release label contains '/'")
```
lcarva: This error message is misleading. If the release label is `'abc'` user will get this message and be quite confused. Can we instead output the expected regex? `RELEASE_LABEL_FORMAT.pattern` might give you what you want.

osbs/constants.py
```diff
@@ -102,6 +102,10 @@
 BUILD_TYPE_WORKER = "worker"
 
 ISOLATED_RELEASE_FORMAT = re.compile(r'^\d+\.\d+(\..+)?$')
+RELEASE_LABEL_FORMAT = re.compile(r"""^\d+             # First character must be a digit
+                                      ([._]?           # allow separators between groups
+                                      [a-zA-Z0-9]+)*$  # last characters must be alphanumeric""",
```
lcarva: I really enjoy this way of wring regexes, much easier to read!

A minor improvement would be to move the closing `""",` to a new line.

osbs/api.py
twaugh: Yes... once you remove whitespace and comments. :-)
Perhaps `re.sub('\s', '', re.sub(r' #.*(\n|$)', '', RELEASE_LABEL_FORMAT.pattern))`

osbs/constants.py
rcerven: sure

lcarva:
@mlangsdorf, I think your requested changes have been addressed. PTAL

rcerven:
rn updated: raise error if release label contains forbidden characters

PR #776, opened 2018-06-08
import image imports only specified tags, if provided

rcerven:

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -814,7 +814,7 @@ class Openshift(object):
         return response
 
     @retry_on_conflict
-    def import_image(self, name, stream_import):
+    def import_image(self, name, stream_import, tags=None):
         """
         Import image tags from a Docker registry into an ImageStream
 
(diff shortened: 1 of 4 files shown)


osbs/core.py
```diff
@@ -837,13 +837,14 @@ def import_image(self, name, stream_import):
...
         stream_import['metadata']['name'] = name
         for tag in imagestream_json.get('spec', {}).get('tags', []):
-            image_import = {
-                'from': tag['from'],
-                'to': {'name': tag['name']},
-                'importPolicy': tag.get('importPolicy'),
-                'referencePolicy': tag.get('referencePolicy'),
-            }
-            stream_import['spec']['images'].append(image_import)
+            if not tags or tag['name'] in tags:
```
lcarva: Minor suggestion is to reverse the check to `continue` the loop instead of nesting the main piece of code in the `if-block`. For example:

```python
if tags and tag['name'] not in tags:
  continue

# Do the work here
```

osbs/core.py
```diff
@@ -837,13 +837,14 @@ def import_image(self, name, stream_import):
 
         stream_import['metadata']['name'] = name
         for tag in imagestream_json.get('spec', {}).get('tags', []):
```
lcarva: Maybe convert `tags` to a set to make it a bit more efficient and to be clear of how it's going to be used?

osbs/core.py
rcerven: will do

osbs/core.py
rcerven: ok

PR #777, opened 2018-06-11
BuildRequest: sanitize most labels

mlangsdorf:
OpenShift requires labels be 64 characters or less and only contain "clean"
characters. Pull the sanitation code from make_name_from_git and use it for
setting labels.

Isolated labels can contain periods. Don't ask me why.

Fixes a bug introduced in PR-771.


tests/test_utils.py
```diff
@@ -173,6 +189,11 @@ def test_get_time_from_rfc3339_valid(rfc3339, seconds, tz):
     ('1_2_3_4_', '_f_', 10, '-', '1234-f'),
     ('longer-name-than', 'this', 22, '-', 'longer-name-than-this'),
     ('longer--name--than', 'this', 22, '-', 'longer--name--tha-this'),
+    ('https://github.com/blah/my_very_very_very_long_and_broken_a$$_repo.git',
+     'bacon', 65, '-', 'myveryveryverylongandbrokenarepo-bacon'),
+    ('https://github.com/blah/my_very_very_very_long_and_broken_a$$_repo.git', '',
+      65, '-', 'myveryveryverylongandbrokenarepo-unknown'),
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

twaugh:
The [actual rules for labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) are:
* keys are:
  * [optional] _prefix_, DNS subdomain, up to 253 characters, followed by a slash
  * _name_, up to 63 characters, beginning and ending with alphanumeric, `-`/`_`/`.` allowed between
* values follow the same rule as _name_ above

I can't seem to locate the rules for BuildConfig names.

osbs/utils.py
```diff
@@ -333,6 +333,40 @@ def utcnow():
 VALID_BUILD_CONFIG_NAME_CHARS = re.compile('[-a-z0-9]')
 
 
+def sanitize_strings_for_openshift(str1, str2='', limit=64, separator='-'):
```
twaugh: I think `limit` should default to 63.

osbs/utils.py
```diff
@@ -333,6 +333,40 @@ def utcnow():
...
+    """
+    OpenShift requires labels and build names to be no more than 64 characters and
+    forbids any characters other than alphanumerics and -.
```
twaugh: This isn't strictly true -- dots and underscores are also allowed, but the names cannot begin or end with punctuation.

osbs/utils.py
```diff
@@ -333,6 +333,40 @@ def utcnow():
...
+    """
+    str1_san = ''.join(filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(str1)))
+    str2_san = ''.join(filter(VALID_BUILD_CONFIG_NAME_CHARS.match, list(str2)))
+
+    str1_chars = []
+    str2_chars = []
+    groups = ((str1_san, str1_chars), (str2_san, str2_chars))
+
+    size = len(separator)
+    limit = min(limit, 64)
```
twaugh: Should be 63.

osbs/utils.py
```diff
@@ -333,6 +333,40 @@ def utcnow():
...
+                group_chars.append(group[i])
+                size += 1
+                if size >= limit:
+                    break
+        else:
+            continue
+        break
+
+    final_str1 = ''.join(str1_chars).strip('-')
+    final_str2 = ''.join(str2_chars).strip('-')
```
twaugh: Wonder if these ought to be `strip(separator)`.

osbs/utils.py

osbs/utils.py

osbs/utils.py

osbs/utils.py
mlangsdorf: Error in the original code, but fixed.

osbs/utils.py
```diff
@@ -331,6 +331,44 @@ def utcnow():
 
 
 VALID_BUILD_CONFIG_NAME_CHARS = re.compile('[-a-z0-9]')
+VALID_LABEL_CHARS = re.compile('[-a-z0-9\/\.]')
```
twaugh: We probably don't want to include `/` because label keys can only include that once, and anything before the `/` has to be valid for a DNS subdomain.

sosiouxme:
release note changes from 771 should suffice

PR #779, opened 2018-06-15
Do not load the flatpak_create_plugin if we are not building a flatpak

cverna:
Without this we are trying to run the pre_flatpak_create_dockerfile plugin when trying to build normal containers.

Maybe @mlangsdorf could review this and make sure this is a sane change.

I ll be happy to work on some tests if I could get some pointers on which test to modify.

cc'ing @owtaylor 

Signed-off-by: Clement Verna <cverna@tutanota.com>

### DIFF
--- a/tests/build_/test_plugins_configuration.py
+++ b/tests/build_/test_plugins_configuration.py
@@ -244,16 +244,23 @@ class TestPluginsConfiguration(object):
 
     @pytest.mark.parametrize('build_type', (BUILD_TYPE_ORCHESTRATOR, BUILD_TYPE_WORKER))
     @pytest.mark.parametrize('compose_ids', (None, [], [42], [42, 2]))
-    def test_render_flatpak(self, compose_ids, build_type):
+    @pytest.mark.parametrize('flatpak_base_image', (TEST_FLATPAK_BASE_IMAGE, None))
+    def test_render_flatpak(self, compose_ids, build_type, flatpak_base_image):
         extra_args = {
             'flatpak': True,
             'compose_ids': compose_ids,
-            'flatpak_base_image': TEST_FLATPAK_BASE_IMAGE,
+            'flatpak_base_image': flatpak_base_image,
             'base_image': TEST_FLATPAK_BASE_IMAGE,
             'build_type': build_type,
         }
 
-        user_params = get_sample_user_params(extra_args)
+        if flatpak_base_image is None:
+            with pytest.raises(OsbsValidationException):
+                user_params = get_sample_user_params(extra_args)
+            return
+        else:
+            user_params = get_sample_user_params(extra_args)
+
         self.mock_repo_info()
         build_json = PluginsConfiguration(user_params).render()
 
@@ -269,11 +276,12 @@ class TestPluginsConfiguration(object):
         else:
             assert args['compose_ids'] == compose_ids
 
-        plugin = get_plugin(plugins, "prebuild_plugins", "flatpak_create_dockerfile")
-        assert plugin
+        if flatpak_base_image is not None:
+            plugin = get_plugin(plugins, "prebuild_plugins", "flatpak_create_dockerfile")
+            assert plugin
 
-        args = plugin['args']
-        assert args['base_image'] == TEST_FLATPAK_BASE_IMAGE
+            args = plugin['args']
+            assert args['base_image'] == TEST_FLATPAK_BASE_IMAGE
 
         if build_type == BUILD_TYPE_ORCHESTRATOR:
             plugin = get_plugin(plugins, "prebuild_plugins", "bump_release")
@@ -294,9 +302,11 @@ class TestPluginsConfiguration(object):
             assert get_plugin(plugins, "postbuild_plugins", "import_image")
 
     @pytest.mark.parametrize('build_type', (BUILD_TYPE_ORCHESTRATOR, BUILD_TYPE_WORKER))
-    def test_render_prod_not_flatpak(self, build_type):
+    @pytest.mark.parametrize('flatpak_base_image', (TEST_FLATPAK_BASE_IMAGE, None))
+    def test_render_prod_not_flatpak(self, build_type, flatpak_base_image):
         extra_args = {
             'flatpak': False,
+            'flatpak_base_image': flatpak_base_image,
             'build_type': build_type,
         }
         user_params = get_sample_user_params(extra_args)
@@ -329,6 +339,7 @@ class TestPluginsConfiguration(object):
             'release': release,
             'flatpak': flatpak,
             'build_type': BUILD_TYPE_ORCHESTRATOR,
+            'flatpak_base_image': TEST_FLATPAK_BASE_IMAGE,
         }
 
         user_params = get_sample_user_params(extra_args)
--- a/osbs/build/plugins_configuration.py
+++ b/osbs/build/plugins_configuration.py
@@ -280,6 +280,11 @@ class PluginsConfiguration(object):
         plugin = 'flatpak_create_dockerfile'
 
         if self.pt.has_plugin_conf(phase, plugin):
+
+            if not self.user_params.flatpak.value:
+                self.pt.remove_plugin(phase, plugin)
+                return
+
             if not self.pt.set_plugin_arg_valid(phase, plugin, 'base_image',
                                                 self.user_params.flatpak_base_image.value):
                 self.pt.remove_plugin(phase, plugin, 'unable to set flatpak base image')
--- a/osbs/build/user_params.py
+++ b/osbs/build/user_params.py
@@ -130,7 +130,10 @@ class BuildUserParams(BuildCommon):
         self.isolated.value = isolated
         self.scratch.value = scratch
 
-        if not flatpak:
+        if flatpak:
+            if not flatpak_base_image:
+                raise OsbsValidationException("faltpak_base_image must be provided")
+        else:
             if not base_image:
                 raise OsbsValidationException("base_image must be provided")
             self.trigger_imagestreamtag.value = get_imagestreamtag_from_image(base_image)


cverna:
Thanks for the pointers, I ll work on modifying the tests.

Following is the log of the failing build if you are interested https://kojipkgs.stg.fedoraproject.org//work/tasks/1014/90001014/orchestrator.log

owtaylor:
On Fri, Jun 15, 2018, 7:26 AM Mark Langsdorf <notifications@github.com>
wrote:

> I'm curious why your buildconfig is passing a flatpak_base_image value
> when not building a flatpak, but this is still a bug.
>
In Fedora a single Koji instance and hence osbs.conf is used to build both
docker containers and flatpaks.

>

cverna:
> Could you add a test for this?
Yes working on it :monkey_face:

tests/build_/test_plugins_configuration.py
```diff
@@ -294,9 +294,12 @@ def test_render_flatpak(self, compose_ids, build_type):
             assert get_plugin(plugins, "postbuild_plugins", "import_image")
 
     @pytest.mark.parametrize('build_type', (BUILD_TYPE_ORCHESTRATOR, BUILD_TYPE_WORKER))
-    def test_render_prod_not_flatpak(self, build_type):
+    @pytest.mark.parametrize('flatpak', (False, True, False))
```
lcarva: It's unnecessary to list `False` twice. This will cause identical tests to run twice.
Same for `flatpak_base_image` parameter below.

tests/build_/test_plugins_configuration.py
```diff
@@ -305,18 +308,19 @@ def test_render_prod_not_flatpak(self, build_type):
...
         plugins = get_plugins_from_build_json(build_json)
 
-        with pytest.raises(NoSuchPluginException):
-            get_plugin(plugins, "prebuild_plugins", "resolve_module_compose")
-        with pytest.raises(NoSuchPluginException):
-            get_plugin(plugins, "prebuild_plugins", "flatpak_create_dockerfile")
-        with pytest.raises(NoSuchPluginException):
-            get_plugin(plugins, "prepublish_plugins", "flatpak_create_oci")
-        if build_type == BUILD_TYPE_ORCHESTRATOR:
+        if not flatpak:
```
lcarva: What's the point of this test when `flatpak` is true? It doesn't seem to do much and there are no assertions in such case.

tests/build_/test_plugins_configuration.py
cverna: oh yes good catch :+1:

tests/build_/test_plugins_configuration.py
mlangsdorf: I think it needs to be replaced with if flatpak and flatpak_base_image and some code to assert that the flatpak stuff is there, and then the else code that has this stuff in it.

It would be best to just squash both test_render_flatpak and test_render_prod_not_flatpak into one test.

cverna: So I have added the assert for the `flatpak and flatpak_base_image` case. I ll look at `test_render_flatpak`

cverna:
@mlangsdorf currently `flatpak = True and flatpak_base_image = Flase` is loading the `resolve_module_compose`.

I am not 100% if this is correct or not, based on the `flatpak_create_dockerfile` plugin should we check if we have a `flatpak_base_image` before loading the plugin.

tests/build_/test_plugins_configuration.py
```diff
@@ -294,9 +294,12 @@ def test_render_flatpak(self, compose_ids, build_type):
             assert get_plugin(plugins, "postbuild_plugins", "import_image")
 
     @pytest.mark.parametrize('build_type', (BUILD_TYPE_ORCHESTRATOR, BUILD_TYPE_WORKER))
-    def test_render_prod_not_flatpak(self, build_type):
+    @pytest.mark.parametrize('flatpak', (False, True))
+    @pytest.mark.parametrize('flatpak_base_image', (TEST_FLATPAK_BASE_IMAGE, False))
```
owtaylor: Would expect None not False for flatpak_base_image

tests/build_/test_plugins_configuration.py
```diff
@@ -294,9 +294,12 @@ def test_render_flatpak(self, compose_ids, build_type):
...
+    def test_render_prod_not_flatpak(self, build_type, flatpak, flatpak_base_image):
```
owtaylor: It doesn't make sense to me to add flatpak True/False to a test that is called _not_flatpak - the point of this test was to check what happens when Flatpak is not set. Parameterization here of flatpak_base_image here would make sense - and should have no effect on the resulting set of plugins. If you wanted to check what happens for flatpak = True, flatpak_base_image = False then that would belong in test_render_flatpak.

Right now, I think you get a plugin configuration that will make atomic-reactor fail -with resolve_module_compose and flatpak_create_oci plugins, but no flatpak_create_dockerfile plugin. You could assert that you get that, but it would certainly be better to fail in osbs-client with an informative error message about the missing config key, rather than fail in atomic-reactor with '"flatpak_create_dockerfile must be run before flatpak_create_oci" which doesn't help the poor admin at all!

[ As far as I can see, flatpak_base_image should have moved to the reactor config and out of osbs.conf with other site-static configuration. It was probably left over because it sounds like a "base image" - that you might want different for different builds. But actually, the base image should have little on the final contents of the generated Flatpak OCI - it's just the image that provides rpm/dnf/etc. to assemble the contents. So we should fix that up at some point, but let's not do it in this PR. There's not much harm in having it left over since we're not doing openshift-triggered layered image rebuilds for flatpaks. ]

lcarva: Filed https://github.com/projectatomic/osbs-client/issues/784 for tracking `flatpak_base_image` move to reactor config map.

mlangsdorf:
I think it doesn't make sense to enable resolve_composes unless we've got a valid flatpak_base_image and we're doing a flatpak build.

Would you please remove the test_render_not_flatpak and move all the logic into test_render_flatpak? I'd like to see one test function, something like:
`
@pytest.mark.parametrize('build_type', (BUILD_TYPE_ORCHESTRATOR, BUILD_TYPE_WORKER))
@pytest.mark.parametrize('compose_ids', (None, [], [42], [42, 2]))
@pytest.mark.parametrize('flatpak_base_image', (TEST_FLATPAK_BASE_IMAGE, None))
@pytest.mark.parametrize('flatpak', (True, False))
def test_render_flatpak(self, compose_ids, build_type, flatpak, flatpak_base_image):
`
It confirms the plugins are set up if and only if both flatpak and flatpak_base_image are true/have values, and confirms that the plugins are disabled if either flatpak is false or flatpak_base_image is None.

It should be pretty easy - all the logic is there.

owtaylor:
If flatpak is specified, then it doesn't make any sense to proceed if
flatpak_base_image is not configured. You can't fall back to a docker build
if you just remove the plugins and submit the build: that's just asking
atomic reactor to haul in a confusing way. osbs-client needs to fail with
that combination.

cverna:
@mlangsdorf, I think I kind of agree with @owtaylor here, `if flatpak = True and flatpak_base_image is None` should raise an exception in osbs-client for `flatpak_create_dockerfile`, `flatpak_create_oci` and `resolve_module_compose`.

I ll modify this PR in that direction if this is ok with you.

owtaylor:
> @mlangsdorf, I think I kind of agree with @owtaylor here, if flatpak = True and flatpak_base_image is None should raise an exception in osbs-client for flatpak_create_dockerfile, flatpak_create_oci and resolve_module_compose.

I think it's a bit better to check up front - you are going to have trouble getting complete test coverage if you do it in the render_<plugin> because the first plugin will fail and you never will get to the second. Perhaps it could be done in user_params.py - though if you put it in set_params(), that doesn't look like it gets run in actual use through the api - only from the test code? the api seems to only use from_json which doesn't validate anything?

mlangsdorf:
set_params is called in osbs/api.py line 610, as part of _do_create_prod_build, after retrieving either a BuildRequest or BuildRequestV2 depending on the arrangement version. So user_params.set_params is the right place do raise the error - there's currently an if not flatpack: test around line 133, which should be replaced by 
```
if flatpak:
    if not flatpak_base_image:
        raise OsbsValidationException("flatpak_base_image must be provided)
else:
    # existing checks go here
```

owtaylor:
I think I'm failing to follow some code path - the code that is being patched here - and that was apparently triggering a problem in Fedora staging - is the code in plugin_configurations.py. But outside of testing/ PluginsConfiguration is referenced only by osbs.render_plugins_configuration() which as far as I can tell is not used in osbs-client, koji-containerbuild, or atomic_reactor. So clearly I'm missing something :-)

mlangsdorf:
The way it works is:
someone calls _do_create_prod_build in osbs_client.api
_do_create_prod_build gets a BuildRequestV2 object for arrangement 6 and later via get_build_request
_do_create_prod_build sets the user params via build_request.set_params()
_do_create_prod_build spawns atomic_reactor, passing the json of the build configuration via the USER_PARAMS environment variable
atomic reactor calls the input_osv3 plugin
the input_osv3 plugin retrieves the build json from USER_PARAMS
the input_osv3 plugin creates an an osbs instance and calls osbs.render_plugins_configuration
the input_osv3 plugin finishes removing plugins based on the reactor config - including removing resolve_composes if there isn't an odcs value in the reactor_config

at some point, I'd like to re-engineer some of this to be a little less confusing.

owtaylor:
But doesn't all that mean that BuildUserParams.set_params() is never called?

mlangsdorf:
in step 2oc, build_request is a BuildRequestV2, so build_request.set_params() is BuildRequestV2.set_params(), which calls self.user_params.set_params().

twaugh:
I restarted the failed Travis CI test runs and they succeeded.

lcarva:
Updated release notes draft:
> - `flatpak_create_dockerfile` is only enabled for flatpak builds

PR #780, opened 2018-06-18
Allow import of specific image tags from CLI

lcarva:


lcarva:
Updated release notes draft:
> - `import-image` CLI command allows import of specific image tags

PR #781, opened 2018-06-18
when deciding whether it is orchestrator or worker build, use build_t…

rcerven:
…ype, instead of checking platforms


lcarva:
Updated release notes draft:
> - Use build_type instead of heuristics to distinguish between worker and orchestrator builds

PR #782, opened 2018-06-18
Check import-image response and retry if needed

lcarva:
In some cases, the container registry used may return a 500 error
response when image is fetched by OpenShift via ImageStream. This has
been seen when using pulp+crane container registry.

This change will verify that each requested imported tag was successful.
If not successful, and container registry has returned a 500 error, the
import image operation is retried.

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -21,10 +21,12 @@ from osbs.constants import (DEFAULT_NAMESPACE, BUILD_FINISHED_STATES, BUILD_RUNN
                             SERVICEACCOUNT_CACRT, ANNOTATION_SOURCE_REPO,
                             ANNOTATION_INSECURE_REPO)
 from osbs.exceptions import (OsbsResponseException, OsbsException,
-                             OsbsWatchBuildNotFound, OsbsAuthException)
-from osbs.utils import graceful_chain_get, retry_on_conflict
+                             OsbsWatchBuildNotFound, OsbsAuthException,
+                             ImportImageFailed, ImportImageFailedServerError)
+from osbs.utils import graceful_chain_get, retry_on_conflict, retry_on_exception
 from requests.exceptions import ConnectionError
 from requests.utils import guess_json_utf
+from requests import codes
 
 from six.moves import http_client
 from six.moves.urllib.parse import urljoin, urlencode, urlparse, parse_qs
@@ -844,6 +846,7 @@ class Openshift(object):
         return response
 
     @retry_on_conflict
+    @retry_on_exception(ImportImageFailedServerError)
     def import_image(self, name, stream_import, tags=None):
         """
         Import image tags from a Docker registry into an ImageStream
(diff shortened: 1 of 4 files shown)


tests/test_core.py
```diff
@@ -700,3 +701,49 @@ def test_import_image(self, openshift, tags, imagestream_name, expect_update, ex
...
+        json_path = os.path.join(this_dir, "mock_jsons", openshift._con.version,
+                                 'imagestreamimport.json')
+        with open(json_path) as f:
+            content_json = json.load(f)
+
+        # Assume mocked data contains a good response
+        good_resp = HttpResponse(200, {}, content=json.dumps(content_json).encode('utf-8'))
+
+        # Create a bad response
+        for image in content_json['status']['images']:
```
twaugh: This assumes .status.images has at least one element, and it also sets the 'bad' status for all of those images.
Instead, how about this?:
```
# Only update first image status.
# A single server error should cause retry.
image = content_json['status']['images'][0]
```

osbs/core.py
```diff
@@ -895,6 +899,28 @@ def import_image(self, name, stream_import, tags=None):
...
         return True
 
+    def _check_import_image_response(self, import_response):
+        check_response(import_response)
+
+        failed_images = []
+        failed_images_server_error = False
+
+        for image in import_response.json()['status']['images']:
+            if image['status']['status'] == 'Success':
```
twaugh: Do we care about verifying that the status descriptions are for the specific set of tags we care about? Is that something the API ensures?

lcarva: [ImageStreamImportStatus](https://docs.openshift.com/container-platform/3.6/rest_api/openshift_v1.html#v1-imagestreamimportstatus) is only populated with the images that were actually requested to be imported based on the [ImageStreamImportSpec](https://docs.openshift.com/container-platform/3.6/rest_api/openshift_v1.html#v1-imagestreamimportspec).

tests/test_core.py
lcarva: Good call! Changed it.

tests/test_core.py
```diff
@@ -700,3 +701,48 @@ def test_import_image(self, openshift, tags, imagestream_name, expect_update, ex
...
+                                 'imagestreamimport.json')
+        with open(json_path) as f:
+            content_json = json.load(f)
+
+        # Assume mocked data contains a good response
+        good_resp = HttpResponse(200, {}, content=json.dumps(content_json).encode('utf-8'))
+
+        # Create a bad response by marking the first image as failed
+        for key in ('status', 'code', 'reason'):
+            content_json['status']['images'][0]['status'][key] = image_status[key]
```
twaugh: Now that I look at this, I guess it could be:
```
content_json['status']['images'][0]['status'].update(image_status)
```
but that's just being picky now.

lcarva: That's true, but I wanted to be more explicit in case someone added a new use case in this unit test that had a typo in the keys used.

lcarva:
Updated release notes draft:
> - Check response when importing images and retry if needed

PR #785, opened 2018-06-22
plugins_configuration: disable check_and_set_platforms for scratch bu…

mlangsdorf:
…ilds

scratch builds may run on a subset of available architectures. don't run
check_and_set_platforms when running scratch builds, so that any arches
specified by the user are used instead of taking arches from koji.


mlangsdorf:
added to release notes under build json changes:
- check_and_set_platforms is removed in scratch builds so that user specified platforms will be used in atomic_reactor

PR #786, opened 2018-06-25
plugins_configuration.py: Don't exclude the 'koji' plugin for Flatpaks

owtaylor:
This was implemented for build_request.py in 3c18d4bb44, but the change
never made its way into plugins_configuration.py.


lcarva:
Updated release notes draft:
> - `koji` plugin not excluded for Flatpaks

PR #787, opened 2018-06-25
Don't include 'resolve_composes' for Flatpak builds

owtaylor:
For Flatpak builds, handling of composes is done in
resolve_module_compose, so the 'resolve_composes' plugin just duplicates
this and causes things to go wrong.

### DIFF
--- a/osbs/build/plugins_configuration.py
+++ b/osbs/build/plugins_configuration.py
@@ -435,6 +435,10 @@ class PluginsConfiguration(object):
         if not self.pt.has_plugin_conf(phase, plugin):
             return
 
+        if self.user_params.flatpak.value:
+            self.pt.remove_plugin(phase, plugin, 'flatpak build')
+            return
+
         if self.user_params.yum_repourls.value:
             self.pt.remove_plugin(phase, plugin, 'yum repourls specified in user parameters')
             return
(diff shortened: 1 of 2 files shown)


mlangsdorf:
when is resolve_composes used, then? When we have an ODCS but not yum_repourls and we're not building a flatpak build? what kind of build does that describe?

I'm not saying this is wrong, but it's a substantial change from the BuildRequest (arrangement 5) version of this code, which as far as I can tell did not disable resolve_composes for flatpak builds.

owtaylor:
resolve_composes is used: for non-flatpak builds with a compose: section in container.yaml - this provides a flexible way of specifying a package set for a container to install frm.

resolve_module_compose is used: for flatpak builds with a compose: section in container.yaml - that is, all flatpak builds.

[ I originally wrote resolve_module_compose intending it to be general beyond flatpaks, but it is tied to the idea of a compose of a single module, and when general ODCS support was added, something more general was wanted - but the more general case doesn't doesn't work for Flatpak  ]

owtaylor:
Oh - in terms of build_request.py - yes, the same change would logically be needed there. I didn't do it since we're no longer testing that code - I'm not fully sure why the exact problem I was hitting that inspired this change didn't come up in the testing I did in February with the old code.

lcarva:
Updated release notes draft:
> - `resolve_composes` plugin excluded for Flatpak builds

PR #789, opened 2018-06-28
Fix yaml exception wrapping

twaugh:
Previously the traceback could look like this:
```
[...]
OsbsException: ScannerError()

Original traceback (most recent call last):
[...]
  File "/usr/lib64/python2.6/site-packages/yaml/scanner.py", line 289, in stale_possible_simple_keys
    "could not found expected ':'", self.get_mark())
ScannerError()
```

This is because repr() for a yaml exception only reveals 'ScannerError()'
whereas str() reveals the line and column.

Use str() on the underlying exception when str() is called on the
wrapping exception.


twaugh:
... python 3 uses repr() anyway, so there's no point in doing this.

PR #792, opened 2018-07-17
core: add retry on timeout to get_image_stream_tag

mlangsdorf:
image pulls seem to have a high number of gateway timeouts, even when
the gateway server should be configured correctly. add a retry to
get_image_stream_tag that checks for 504 and related errors.

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -23,7 +23,7 @@ from osbs.constants import (DEFAULT_NAMESPACE, BUILD_FINISHED_STATES, BUILD_RUNN
 from osbs.exceptions import (OsbsResponseException, OsbsException,
                              OsbsWatchBuildNotFound, OsbsAuthException,
                              ImportImageFailed, ImportImageFailedServerError)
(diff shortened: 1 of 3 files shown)


twaugh:
Sometimes a 504 error is a result of the server spending too long trying to put the response together. I worry a little that if we just retry when seeing these errors we'll make things worse by increasing the load on the server even further.

For the Travis CI test failure, see #793.

osbs/core.py
```diff
@@ -765,6 +765,7 @@ def set_annotations_on_build(self, build_id, annotations):
                                                 'annotations', annotations,
                                                 self._replace_metadata_things)
 
+    @retry_on_timeout
     def get_image_stream_tag(self, tag_id):
```
lcarva: This is not the api where we're seeing errors:
```
HTTPSConnectionPool(host='xxxxx', port=443): Max retries exceeded with url: /oapi/v1/namespaces/osbs-prod/imagestreamimports/ (Caused by ResponseError('too many 504 error responses',)
```

mlangsdorf:
Wrong fix to the wrong problem.

PR #794, opened 2018-07-20
 Always include check_and_set_platforms for orchestrated builds

owtaylor:
There was a lot of (barely used) complexity in atomic reactor to handle the check_and_set_platforms plugin being only run for builds with koji_target. Including it for all orchestrated builds allows us to simplify that logic. (This does introduce incompatibility with older atomic-reactor for the rarer case of builds missing koji_target.)

See https://github.com/projectatomic/atomic-reactor/pull/1046 for the atomic-reactor piece.

This PR also contains a patch to fix the parameterization of the check_and_set_platforms test.

### DIFF
--- a/osbs/build/plugins_configuration.py
+++ b/osbs/build/plugins_configuration.py
@@ -337,9 +337,9 @@ class PluginsConfiguration(object):
         if not self.pt.has_plugin_conf(phase, plugin):
             return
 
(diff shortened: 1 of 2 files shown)


lcarva:
@twaugh, I'd like your thoughts on this before merging.

lcarva:
@twaugh ping

lcarva:
Updated release notes draft:
> - `check_and_set_platforms` is always included for orchestrated builds

PR #796, opened 2018-07-23
Increase max time-out and retry backoff factor

mlangsdorf:
Review of the server logs showed that roughly 15% of HTTP requests are
completing after 600 seconds, and almost all the rest are completing
within a few seconds. Our current backoff and retry strategy of starting
retries after 4 seconds is inefficient, and the max time-out of 600 seconds
means that some tasks that complete on the server side are failing on the
osbs-client side because osbs-client is timing out too early.

Increase the back-off factor to 60 seconds, reduce the number of retries
to 60, and increase the max time-out to 1900 seconds so that all the
retries have time to complete. Include documentation explaining this
change.

### DIFF
--- a/osbs/constants.py
+++ b/osbs/constants.py
@@ -77,20 +77,26 @@ REPO_CONFIG_FILE = '.osbs-repo-config'
 ADDITIONAL_TAGS_FILE = 'additional-tags'
 REPO_CONTAINER_CONFIG = 'container.yaml'
 
+# Analysis of server logs shows that most requests either complete almost immediately or take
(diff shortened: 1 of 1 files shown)


sosiouxme:
I would guess that most/all of the requests that are longer than a minute are logs streaming. Those aren't the kind that need a retry, so I wouldn't factor them in. This seems like it would slow down legitimate retries for little benefit?

lcarva:
I think the scope of this change should be just for the `imagestreamimport` api call.

mlangsdorf:
Bad idea.

PR #800, opened 2018-08-29
Flatpak config cleanups

owtaylor:
Couple of tiny cleanups as a result of https://github.com/projectatomic/atomic-reactor/pull/1064 and https://github.com/projectatomic/atomic-reactor/pull/1065.


mlangsdorf:
Looks good to me, but I'm not sure if the Travis failure is bogus or not.

owtaylor:
Travis failure is https://pypi.org/project/py releasing a new version (1.6.0) a couple of days ago that drops python-2.6 support. test.sh could probably force downgrade that to the previous release with the right pip incantations - if centos 6 is still needed in the test matrix.

lcarva:
Please rebase to pick up unit test fix.

lcarva:
Updated release notes draft:
> - Flatpak base image can be set from reactor config map

PR #801, opened 2018-08-31
test.sh: Force py == 1.5.4 on Centos 6

owtaylor:
Version 1.6.0 of the py module is incompatible with Python-2.6.

This should get the Travis build working again. Trying to test against Python-2.6 may not be sustainable long term :-)


lcarva:
Not updating release notes draft because this is a change solely to unit tests.

PR #802, opened 2018-09-03
use build for koji id, if already exists

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -241,6 +241,18 @@ class OSBS(object):
                 running.append(br)
         return running
 
+    def _get_not_cancelled_builds_for_koji_task(self, koji_task_id):
+        all_builds_for_task = self.os.list_builds(koji_task_id=koji_task_id).json()['items']
+        not_cancelled = []
+
+        for b in all_builds_for_task:
+            br = BuildResponse(b)
+            build_labels = br.get_labels()
+            if not br.is_cancelled() and build_labels['is_autorebuild'] == "false":
+                not_cancelled.append(br)
+
+        return not_cancelled
+
     def _verify_labels_match(self, new_build_config, existing_build_config):
         new_labels = new_build_config['metadata']['labels']
         existing_labels = existing_build_config['metadata']['labels']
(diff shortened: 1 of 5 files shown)


osbs/api.py
```diff
@@ -241,6 +241,17 @@ def _get_running_builds_for_build_config(self, build_config_id):
                 running.append(br)
         return running
 
+    def _get_not_cancelled_builds_for_koji_task(self, koji_task_id):
+        all_builds_for_task = self.os.list_builds(koji_task_id=koji_task_id).json()['items']
+        not_cancelled = []
+
+        for b in all_builds_for_task:
```
alcm-b: Looks like there's a shorter version:
~~~~
responses = (BuildResponse(b) for b in all_builds_for_task)
not_cancelled = [br for br in responses if not br.is_cancelled()]
~~~~

osbs/api.py
```diff
@@ -680,7 +691,17 @@ def _do_create_prod_build(self, git_uri, git_ref,
         )
         build_request.set_openshift_required_version(self.os_conf.get_openshift_required_version())
         build_request.set_repo_info(repo_info)
-        if build_request.scratch:
+
+        builds_for_koji_task = []
+        if koji_task_id and not is_auto:
```
lcarva: I think we want to do this just for orchestrator builds. If we do that, there's no need to check `is_auto` since orchestrator autorebuilds will not go through this code path.

osbs/api.py
```diff
@@ -680,7 +691,17 @@ def _do_create_prod_build(self, git_uri, git_ref,
...
+            # try to find build for koji_task which isn't canceled and use that one
+            builds_for_koji_task = self._get_not_cancelled_builds_for_koji_task(koji_task_id)
+
+        if len(builds_for_koji_task) == 1:
```
lcarva: If more than 1, we should probably raise an exception as that's definitely unexpected.

osbs/api.py
```diff
@@ -680,7 +691,17 @@ def _do_create_prod_build(self, git_uri, git_ref,
...
+            response =\
```
lcarva: Let's log what's happening here to help us debug things in the future.

osbs/api.py
```diff
@@ -241,6 +241,17 @@ def _get_running_builds_for_build_config(self, build_config_id):
                 running.append(br)
         return running
 
+    def _get_not_cancelled_builds_for_koji_task(self, koji_task_id):
+        all_builds_for_task = self.os.list_builds(koji_task_id=koji_task_id).json()['items']
+        not_cancelled = []
+
+        for b in all_builds_for_task:
+            br = BuildResponse(b)
+            if not br.is_cancelled():
```
lcarva: Also filter out autorebuilds?

osbs/api.py

osbs/api.py

osbs/api.py

osbs/api.py

osbs/api.py
lcarva: I don't think substituting the for-loop with two list comprehensions makes things clearer in this case.

lcarva:
Updated release notes draft:
> - Use existing build when one already exists for given koji task ID

PR #804, opened 2018-10-01
add verify_media_types to orchestrator_inner:6.json

mlangsdorf:
Resolving which of verify_media_types or pulp_pull will be used in
a particular build is handled by atomic_reactor during input_osv3,
when it has access to the reactor config and scratch configuration.
It is unconditionally added in osbs-client.


inputs/orchestrator_inner:6.json
```diff
@@ -88,6 +88,9 @@
         "insecure": true
       }
     },
+    {
+      "name": "verify_media_types"
```
lcarva: Let's add `"required": false` so it doesn't break older versions of atomic-reactor.

lcarva:
Updated release notes draft:
> - `verify_media_types` added to orchestrator_inner:6.json

PR #805, opened 2018-10-03
verify media types: fix an incorrect name in json, add tests to avoid that mistake in the future

mlangsdorf:
I didn't remember the key name for exit_verify_media_types.py correctly, and put the wrong value in
orchestrator_inner:6.json and test_arrangements.py.  Because I used the wrong value in 
test_arrangements, it didn't detect that I'd put the wrong value in orchestrator_inner:6.json.

Correct that mistake.

Also, use the key values in test_arrangements.py as a hedge against this mistake in the future.

### DIFF
--- a/inputs/orchestrator_inner:6.json
+++ b/inputs/orchestrator_inner:6.json
@@ -89,7 +89,7 @@
       }
     },
     {
-      "name": "verify_media_types",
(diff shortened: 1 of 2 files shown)


lcarva:

PR #807, opened 2018-10-16
Set koji_target for add_filesystem plugin

athos-ribeiro:
See https://github.com/projectatomic/atomic-reactor/pull/1093 for reference


lcarva:
Updated release notes draft:
> - koji_target parameter is set for add_filesystem plugin

PR #808, opened 2018-10-19
remove flatpak_base_image

rcerven:

### DIFF
--- a/osbs/build/spec.py
+++ b/osbs/build/spec.py
@@ -189,7 +189,6 @@ class BuildSpec(BuildCommon):
         self.koji_kerberos_principal = BuildParam("koji_kerberos_principal", allow_none=True)
         self.koji_kerberos_keytab = BuildParam("koji_kerberos_keytab", allow_none=True)
         self.flatpak = BuildParam("flatpak", default=False)
-        self.flatpak_base_image = BuildParam("flatpak_base_image", allow_none=True)
         self.odcs_url = BuildParam("odcs_url", allow_none=True)
         self.odcs_insecure = BuildParam("odcs_insecure", allow_none=True)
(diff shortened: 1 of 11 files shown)


osbs/build/user_params.py
```diff
@@ -126,14 +124,10 @@ def set_params(self,
         self.koji_parent_build.value = koji_parent_build
         self.koji_upload_dir.value = koji_upload_dir
         self.flatpak.value = flatpak
-        self.flatpak_base_image.value = flatpak_base_image
```
mlangsdorf: You'll need to change atomic_reactor/schemas/user_params.json in atomic-reactor

osbs/build/spec.py
```diff
@@ -189,7 +189,6 @@ def __init__(self):
         self.koji_kerberos_principal = BuildParam("koji_kerberos_principal", allow_none=True)
         self.koji_kerberos_keytab = BuildParam("koji_kerberos_keytab", allow_none=True)
         self.flatpak = BuildParam("flatpak", default=False)
-        self.flatpak_base_image = BuildParam("flatpak_base_image", allow_none=True)
```
mlangsdorf: I don't think you want to change build_request v1 here - that changes the behavior of Arrangement 5- systems.

twaugh: I think we're fine with making changes that only affect flatpak builds, as they are not working prior to arrangement 6 in any case.

osbs/build/user_params.py
rcerven: ok will make that change

MartinBasti:
Release notes:

> - flatpak base image is now properly loaded from a reactor-config-map

PR #810, opened 2018-10-23
PluginsConfiguration: allow signing_intent for Flatpaks

owtaylor:
atomic_reactor will be enhanced to allow signing_intent to be passed
for Flatpak builds as well. If a signing intent is specified, pass that
as a parameter to the resolve_module_compose plugin. (Compatiblity
with old atomic-reactor versions is retained in the case where signing
intent is not specified.)


rcerven:
rn updated:
allow signing_intent for Flatpaks

PR #811, opened 2018-10-24
Travis: add install section

MartinBasti:
Without install section, travis do default `pip install -r
requirements.txt` and we don't need it. It only slowdowns builds.


MartinBasti:
The same as: https://github.com/projectatomic/atomic-reactor/pull/1100

MartinBasti:
No release notes needed, change in travis ci only

PR #813, opened 2018-10-31
Disable delete_from_registry plugin

athos-ribeiro:
Deleting content from the registry after pulp finishes syncing is
convenient so the next pulp sync will not take longer (pulp iterates
over the repository tags fetching all referenced content). However,
removing content may confuse the docker distribution registry, which
leads to failing builds.

### DIFF
--- a/inputs/orchestrator_inner:6.json
+++ b/inputs/orchestrator_inner:6.json
@@ -97,9 +97,6 @@
       "required": false
     },
     {
-      "name": "delete_from_registry"
-    },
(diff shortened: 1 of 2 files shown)


athos-ribeiro:
For now, we should perform manual garbage collection.

athos-ribeiro:
@mlangsdorf, any thoughts?

athos-ribeiro:
Added entry to realease notes to inform the plugin was disabled

PR #815, opened 2018-10-31
Remove pulp plugins for Flatpak builds

owtaylor:
Flatpaks are OCI images, and currently Pulp doesn't have OCI image support.
To avoid failures trying to upload to Pulp, simply omit all pulp-related
plugins for Flatpak builds. (The resulting Flatpaks will still be
uploaded to Koji, and to any configured Docker registries with OCI support.)

### DIFF
--- a/osbs/build/plugins_configuration.py
+++ b/osbs/build/plugins_configuration.py
@@ -219,6 +219,29 @@ class PluginsConfiguration(object):
         for when, which in plugins:
             self.pt.remove_plugin(when, which, msg)
 
+    def adjust_for_flatpak(self):
+        """
+        Remove plugins that don't work when building Flatpaks
+        """
+        if self.user_params.flatpak.value:
+            remove_plugins = [
+                ("prebuild_plugins", "resolve_composes"),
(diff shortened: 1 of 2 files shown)


osbs/build/plugins_configuration.py
```diff
@@ -219,6 +219,29 @@ def adjust_for_custom_base_image(self):
         for when, which in plugins:
             self.pt.remove_plugin(when, which, msg)
 
+    def adjust_for_flatpak(self):
+        """
+        Remove plugins that don't work when building Flatpaks
+        """
+        if self.user_params.flatpak.value:
+            remove_plugins = [
+                ("prebuild_plugins", "resolve_composes"),
```
MartinBasti: In other plugins I've seen `atomic_reactor.constants` used instead of hardcoded names. Would be better to keep constants IMO

rcerven: that would mean dependency on atomic_reactor

MartinBasti: Ahh, right, I forgot I'm in client

rcerven:
rn updated:
- Remove pulp plugins for Flatpak builds

PR #816, opened 2018-10-31
Update mock_jsons/

mlangsdorf:
Add a new tests/mock_jsons/3.9.41/ directory with mock JSON data taken from recent runs of osbs-client.


twaugh:
1.0.6? As in OpenShift 1.0.6?
Is that the version you ran these queries against, or was it newer?

mlangsdorf:
Okay, this is now 3.9.41.

mlangsdorf:
Removed all instances of redhat.com and replaced them with an example.com

tests/fake_api.py
```diff
@@ -498,7 +498,7 @@ def delete(self, url, *args, **kwargs):
         return self.request(url, "delete", *args, **kwargs)
 
 
-@pytest.fixture(params=["1.0.4"])
+@pytest.fixture(params=["1.0.4", "3.9.41"])
```
MartinBasti: Do we still care about version `1.0.4`?

mlangsdorf:
Not updating release notes because this is a change to our internal test infrastructure.

I'd like to wait a week or 3 for these values to go through test before removing 1.0.4.

MartinBasti:
Ok, thanks

PR #818, opened 2018-11-05
Remove delete_from_registry plugin checks

athos-ribeiro:
This plugin has been deactivated in 9cb8b97 (PR #813), which was merged without a rebase for the PR #815 changes, causing the related test to fail.

This removes the deactivated plugin from the expected plugins list for non flatpack builds, which fixes the issue.


athos-ribeiro:
No entries to release notes added: Test related change

PR #822, opened 2018-11-12
docs: Update OpenShift references

### LINKED ISSUE #569: docs reference 'oc' without noting it means origin-clients
docs/backups.md refers to a command 'oc'. Naive users are going to assume that is a variant of osbs-client but it's actually in the origin-clients package. origin-clients is not mentioned in the documentation or requirements so this is not an intuitive package to install.

athos-ribeiro:
Explicitly reference the oc client as the OpenShift Client with up to
date Fedora package references for both the `oc` binary and the
OpenShift Origin packages.

* Closes #569

### DIFF
--- a/docs/backups.md
+++ b/docs/backups.md
@@ -26,6 +26,8 @@ It is recommended to perform restore on freshly installed OpenShift with no data
 
 ## Manually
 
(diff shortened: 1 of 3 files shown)


athos-ribeiro:
Docs change. No release notes added.

PR #823, opened 2018-11-12
core.py: tolerate bad response in watch_resources

mlangsdorf:
Since watch_resources operates in a loop, a single bad response
from the _get call shouldn't cancel operation.  Instead, catch
the exception from check_response() and increment a counter.
Only raise the exception to the next level if there are multiple
bad responses in a single watch_resources call.

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -40,6 +40,7 @@ logger = logging.getLogger(__name__)
 # Retry each connection attempt after 30 seconds, for a maximum of 10 times
 WATCH_RETRY_SECS = 30
 WATCH_RETRY = 10
+MAX_BAD_RESPONSES = int(WATCH_RETRY / 3)
 # Give up after 12 hours
 WAIT_RETRY_HOURS = 12
 WAIT_RETRY = int(WAIT_RETRY_HOURS * 3600 / (WATCH_RETRY_SECS * WATCH_RETRY))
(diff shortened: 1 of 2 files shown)


osbs/core.py
```diff
@@ -601,14 +602,31 @@ def delete_resource_quota(self, name):
...
         for retry in range(WATCH_RETRY):
-            response = self._get(url, stream=True, headers={'Connection': 'close'})
-            check_response(response)
+            try:
+                response = self._get(url, stream=True, headers={'Connection': 'close'})
+                check_response(response)
+            # we're already retrying, so there's no need to panic just because of a bad response
+            except OsbsResponseException as exc:
+                bad_responses += 1
+                if bad_responses > MAX_BAD_RESPONSES:
```
MartinBasti: I really like https://pypi.org/project/retry/ which does exactly the same.

Unfortunately it is not packaged in Fedora/Centos. But I really like the way with decorators.

I see there is `retry_on_exception` in `osbs-client/osbs/utils.py`. Could be used that decorator here?

mlangsdorf: retry_on_exception will retry the entire function.  That's overkill.  I just want to catch the exception and continue.

MartinBasti: ok

MartinBasti:
No release notes, internal change only

PR #824, opened 2018-11-15
Validate modules NSV format

athos-ribeiro:
Validate module NSV format to comply to NAME:STREAM[:VERSION] for quick
feedback.


osbs/repo_utils.py
```diff
@@ -59,6 +61,26 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
...
+        modules = self.container.get('compose', {}).get('modules', [])
+        for module in modules:
+            pieces = module.split(':')
+            if not 1 < len(pieces) < 4:
+                raise OsbsValidationException(
+                    'Module specification should be NAME:STREAM or NAME:STREAM:VERSION')
+
+    def get_container_module_specs(self):
+        ModuleSpec = namedtuple('ModuleSpec', 'name, stream, version')
+        ModuleSpec.__new__.__defaults__ = (None,) * len(ModuleSpec._fields)
```
twaugh: Don't we only want a single default here? `(None,)`, not `(None,) * len(ModuleSpec._fields)`? The name and stream are required I think.

Could add a comment here like `# For compatibility with Python < 3.7` so we can tidy this up to use the nicer 'defaults' parameter in future.

tests/build_/test_arrangements.py
```diff
@@ -93,6 +94,10 @@ class MockConfiguration(object):
             def is_autorebuild_enabled(self):
                 return False
 
+            def get_container_module_specs(self):
+                ModuleSpec = namedtuple('ModuleSpec', 'name, stream, version')
```
twaugh: Maybe this definition should come from the module under test.

tests/test_api.py
```diff
@@ -1837,6 +1838,16 @@ class MockConfiguration(object):
             def is_autorebuild_enabled(self):
                 return False
 
+            def get_container_module_specs(self):
+                ModuleSpec = namedtuple('ModuleSpec', 'name, stream, version')
```
twaugh: Would prefer that this came from some place in the code rather than being redefined in multiple tests.

MartinBasti: Hmm thinking about this, probably it can be just real class with a classmethod `.from_text` that will do validation.

```python
class ModuleSpec(object):
    __slots__ = ('name', 'stream', 'version')

    ...init...

    @classmethod 
    def from_text(cls, text):
       .. validation ..
       return cls(*values)
```

osbs/repo_utils.py
```diff
@@ -60,10 +60,30 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
...
+
+    def __init__(self, name, stream, version=None):
+        self.name = name
+        self.stream = stream
+        self.version = version
+
+    @classmethod
+    def from_text(cls, text):
+        pieces = text.split(':')
+        if not 1 < len(pieces) < 4:
```
twaugh: Now we're instantiating the class this check is done for us. Could either leave it to propagate TypeError, or we could catch TypeError and re-raise as ValueError with the message as we do now.

athos-ribeiro: Should we catch attribute error on the line above (split) and re-raise as ValueError instead?

MartinBasti: So regexp would be good probably for this

athos-ribeiro: In atomic-reactor, @owtaylor  added a few comments about not allowing partial entries (which would allow some empty fields). Would this apply here or is it only valid when the CONTEXT field is used?

athos-ribeiro: OK, I am checking for empty strings so we will not be more restrictive than ODCS itself

osbs/repo_utils.py
```diff
@@ -60,10 +60,31 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
...
+            raise ValueError('Module specification should be NAME:STREAM or NAME:STREAM:VERSION')
+        if any(not len(p) for p in pieces):
```
MartinBasti: non-emptiness in python is tested by `not p`.

in your case you can use `not all(pieces)` to make it more readable
```
>>> l = ['something', 'something', '']
>>> all(l)
False
```

MartinBasti: s/non-emptiness/emptiness/

rcerven:
in ModuleSpec fromtext
reuse code from : atomic-reactor -> util.py -> split_module_spec

new method will also accept 'flatpak' param
and based on that it will do either check for flatpak or non-flatpak

in api.py after 631 line (call of _check_labels) call new split_module_spec to validate modules for non-flatpak scenario

athos-ribeiro:
@rcerven, I pulled in some extra bits from atomic_reactor.util. Once this is merged, we can remove the duplications from atomic_reactor and rely on the module validations in this patch.

osbs/repo_utils.py
```diff
@@ -60,10 +60,64 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
             with open(file_path) as f:
                 self.container = (yaml.load(f) or {})
 
+        modules = self.container.get('compose', {}).get('modules', [])
+        self.container_module_specs = [ModuleSpec.from_text(module) for module in modules]
```
rcerven: we should gather all errors first (for all modules) and raise error after, that way user will see all wrong module specs at once, and not 1 by 1

rcerven:
looks good

athos-ribeiro:
Thanks @owtaylor
@lkolacek mentioned that the time to get the feedback for an invalid module name can get quite long. In this case, it would be nice to have the validations here.

osbs/repo_utils.py
```diff
@@ -60,10 +60,73 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
...
+        if value_errors:
+            raise ValueError(value_errors)
+
     def is_autorebuild_enabled(self):
         return self._config_parser.getboolean('autorebuild', 'enabled')
 
 
+class ModuleSpec(object):
+    """
+    Specification for a to be requested module.
```
twaugh: Maybe write that as `to-be-requested`

osbs/repo_utils.py
```diff
@@ -60,10 +60,73 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
...
+        return result
+
+    def __repr__(self):
+        return "ModuleSpec({})".format(self.to_str())
+
+    def __eq__(self, other):
+        return self.__dict__ == other.__dict__
+
+    @classmethod
+    def from_text(cls, text):
```
twaugh: Nit: if this is the inverse of `to_str` should it be called `from_str`?

osbs/repo_utils.py
athos-ribeiro: Done

osbs/repo_utils.py
athos-ribeiro: Done!

athos-ribeiro:
Rebasing on master before merge

athos-ribeiro:
Release notes added:

* Always validate module NSVC format for faster feedback

PR #826, opened 2018-11-22
Extend build_json with images digests data

MartinBasti:
Atomic-reactor will use digests instead of tags for builds to prevent
race conditions caused by floating tags of paretn images. For this
reason osbs-client must provide this information in request

Jira: OSBS-5674

### DIFF
--- a/osbs/build/build_request.py
+++ b/osbs/build/build_request.py
@@ -99,6 +99,11 @@ class BuildRequest(object):
         :param auto_build_node_selector: dict, a nodeselector for auto builds
         :param isolated_build_node_selector: dict, a nodeselector for isolated builds
         :param is_auto: bool, indicates if build is auto build
+        :param parent_images_digests: dict, mapping image names with tags to platform specific
+                                      digests, example:
+                                      {'registry.fedorahosted.org/fedora:29': {
+                                          x86_64': 'registry.fedorahosted.org/fedora@sha256:....'}
+                                      }
         """
 
         # Here we cater to the koji "scratch" build type, this will disable
--- a/osbs/build/plugins_configuration.py
+++ b/osbs/build/plugins_configuration.py
@@ -507,6 +507,15 @@ class PluginsConfiguration(object):
 
         self.pt.set_plugin_arg(phase, plugin, 'tag_suffixes', tag_suffixes)
 
+    def render_pull_base_image(self):
+        """Configure pull_base_image"""
+        phase = 'prebuild_plugins'
+        plugin = 'pull_base_image'
+
+        if self.user_params.parent_images_digests.value:
+            self.pt.set_plugin_arg(phase, plugin, 'parent_images_digests',
+                                   self.user_params.parent_images_digests.value)
+
     def render(self):
         self.user_params.validate()
         # adjust for custom configuration first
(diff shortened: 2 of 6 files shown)


osbs/build/plugins_configuration.py
```diff
@@ -507,6 +507,14 @@ def render_tag_from_config(self):
 
         self.pt.set_plugin_arg(phase, plugin, 'tag_suffixes', tag_suffixes)
 
+    def render_pull_base_image(self):
+        """Configure pull_base_image"""
+        phase = 'prebuild_plugins'
+        plugin = 'pull_base_image'
+
+        self.pt.set_plugin_arg(phase, plugin, 'parent_images_digests',
```
MartinBasti: Now it works, but I have serious concerns about bw compatibility between this version of osbs-client, and older versions of atomic-reactor, because older versions of this plugin don't have keyword.

twaugh: Unknown parameters are ignored by plugins, so passing this extra parameter should be safe.

osbs/build/build_request.py
```diff
@@ -99,6 +99,7 @@ def set_params(self, **kwargs):
         :param auto_build_node_selector: dict, a nodeselector for auto builds
         :param isolated_build_node_selector: dict, a nodeselector for isolated builds
         :param is_auto: bool, indicates if build is auto build
+        :param parent_images_digests: dict, mapping image digests to names and platforms
```
twaugh: What exactly is the structure of this? What are the keys, and what are the values?

osbs/build/plugins_configuration.py
twaugh: Maybe we only want to set this if there is a value to set?

mlangsdorf:
What you have so far looks correct as far as I can tell, but where is this digest data coming from?

MartinBasti:
Digest data are downloaded from manifest lists during orchestrator phase, then digests are passed to workers which will use the same digests as orchestrator (platform specific per worker).

osbs/build/plugins_configuration.py
MartinBasti: Tested, worked.

osbs/build/build_request.py
MartinBasti: Reworded

osbs/build/plugins_configuration.py
MartinBasti: updated

osbs/build/build_request.py
```diff
@@ -99,6 +99,11 @@ def set_params(self, **kwargs):
         :param auto_build_node_selector: dict, a nodeselector for auto builds
         :param isolated_build_node_selector: dict, a nodeselector for isolated builds
         :param is_auto: bool, indicates if build is auto build
+        :param parent_images_digests: dict, mapping image names with tags to platform specific
```
rcerven: maybe we should just delete it from build_request and just keep it (with that example) in build_requestv2,
since only brv2 will use that param anyway

MartinBasti: It is deprecated, but still supported. IMO we should keep it. I don't think that removal of it from v1 will help in future consolidation.

MartinBasti:
Release notes:
> - worker build: allow to pass custom platform specific images digests to be used by workers.

PR #830, opened 2018-11-28
Fix spec file changelog bogus date

athos-ribeiro:


athos-ribeiro:
No release notes change: RPM package dockumentation change.

PR #832, opened 2018-12-04
Remove support for v1-only registries

athos-ribeiro:
We no longer need to handle the case where we are not pushing to a v2
registry.


twaugh:
Still WIP or should the title be changed?

athos-ribeiro:
@twaugh: changed. I would appreciate some insights on the documentation changes though! :)

athos-ribeiro:
Release notes entries added:

## Registry API version

- v1-only registry API is no longer supported
- URIs in `registry_uri` default to `v2` if no version is appended to the URI. Appending `/v1` to the URI is no longer supported

PR #833, opened 2018-12-04
Clean up for flake8 warnings

athos-ribeiro:


MartinBasti:
Would be possible to enable some pylint checks with these changes? Namely unused-import, bare-except ?

athos-ribeiro:
@MartinBasti: While I agree the checks should be enable to avoid regression, I believe they are different checks. unused-import throws false positives for our imported fixtures. Idid activate bare-except though.

athos-ribeiro:
No release notes added. Static analyzers clean ups only

PR #834, opened 2018-12-04
Handle null values for composes in container yaml

athos-ribeiro:
Container yaml file can may contain empty keys. We must handle them on
when fetching values from the file.

Also adding a couple tests to cover this case.

### DIFF
--- a/osbs/repo_utils.py
+++ b/osbs/repo_utils.py
@@ -60,7 +60,9 @@ class RepoConfiguration(object):
             with open(file_path) as f:
                 self.container = (yaml.load(f) or {})
 
-        modules = self.container.get('compose', {}).get('modules', [])
+        # container values may be set to None
+        container_compose = self.container.get('compose') or {}
+        modules = container_compose.get('modules') or []
 
         self.container_module_specs = []
         value_errors = []
--- a/tests/test_repo_utils.py
+++ b/tests/test_repo_utils.py
@@ -128,6 +128,27 @@ class TestRepoConfiguration(object):
                     assert spec.context == expected[3]
             assert spec.profile == expected[4]
 
+    def test_empty_yaml_compose(self, tmpdir):
+        with open(os.path.join(str(tmpdir), REPO_CONTAINER_CONFIG), 'w') as f:
+            f.write(dedent("""\
+                compose:
+                """))
+
+        conf = RepoConfiguration(dir_path=str(tmpdir))
+        assert conf.container['compose'] is None
+        assert conf.container_module_specs == []
+
+    def test_empty_yaml_modules(self, tmpdir):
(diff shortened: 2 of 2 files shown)


osbs/repo_utils.py
```diff
@@ -60,7 +60,9 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
             with open(file_path) as f:
                 self.container = (yaml.load(f) or {})
 
-        modules = self.container.get('compose', {}).get('modules', [])
+        # container values may be set to None
+        container_compose = self.container.get('compose', {}) or {}
```
MartinBasti: IMO we can use only get() without default value, to avoid duplicated definitions of default.

tests/test_repo_utils.py
```diff
@@ -128,6 +128,27 @@ def test_container_module_specs(self, tmpdir, module_nsv, should_raise, expected
...
 
+    def test_empty_yaml_compose(self, tmpdir):
+        with open(os.path.join(str(tmpdir), REPO_CONTAINER_CONFIG), 'w') as f:
+            f.write(dedent("""\
+                compose:
+                """))
+
+        conf = RepoConfiguration(dir_path=str(tmpdir))
+        assert conf.container['compose'] is None
+        assert not len(conf.container_module_specs)
```
MartinBasti: IMO `assert conf.container_module_specs == []` is better

tests/test_repo_utils.py
```diff
@@ -128,6 +128,27 @@ def test_container_module_specs(self, tmpdir, module_nsv, should_raise, expected
...
+    def test_empty_yaml_modules(self, tmpdir):
+        with open(os.path.join(str(tmpdir), REPO_CONTAINER_CONFIG), 'w') as f:
+            f.write(dedent("""\
+                compose:
+                    modules:
+                """))
+
+        conf = RepoConfiguration(dir_path=str(tmpdir))
+        assert conf.container['compose'] == {'modules': None}
+        assert not len(conf.container_module_specs)
```
MartinBasti: Here too

osbs/repo_utils.py
```diff
@@ -60,7 +60,9 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
             with open(file_path) as f:
                 self.container = (yaml.load(f) or {})
 
-        modules = self.container.get('compose', {}).get('modules', [])
+        # container values may be set to None
+        container_compose = self.container.get('compose') or {}
+        modules = container_compose.get('modules') or []
```
mlangsdorf: any particular reason for this change?

athos-ribeiro: The yaml file can contain empty arguments, which will be set to `None`.

If, for example, get('copose', {}) returns None, the next get (for modules) will fail.

athos-ribeiro:
No release notes added. The changes were already in November release.

PR #835, opened 2018-12-07
plugins configuration: remove special handling for custom base images

mlangsdorf:
now that custom base images are handled by the plugins themselves,
there's no need to edit the plugins JSON to deal with them.


mlangsdorf:
Added to Release Notes under JSON changes:
- no plugins are disabled or enabled for custom base images.  Each plugin will decide for itself whether to run for custom_base_images.

PR #836, opened 2018-12-11
Retry git clone operations

athos-ribeiro:
The retry code is based on atomic-reactor's backoff and retry operations [1]. A next step would be to reuse osbs git checkout function in atomic-reactor.

[1] https://github.com/projectatomic/atomic-reactor/blob/e8294b8df36660097db463060fe9610ddf7f7cbb/atomic_reactor/util.py#L305 (the ref points to master's HEAD on the day this PR was open)

### DIFF
--- a/osbs/constants.py
+++ b/osbs/constants.py
@@ -112,3 +112,9 @@ ANNOTATION_INSECURE_REPO = 'openshift.io/image.insecureRepository'
 
 # optional key path for filtering existing build config results
 FILTER_KEY = 'spec.source.git.uri'
+
+# number of retries for git clone operations
+GIT_MAX_RETRIES = 3
+
+# backoff factor for git clone operations - exponential backoff in seconds
+GIT_BACKOFF_FACTOR = 60
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -26,7 +26,10 @@ from datetime import datetime
 from io import BytesIO
 from hashlib import sha256
 from osbs.repo_utils import RepoConfiguration, RepoInfo, AdditionalTagsConfig
-from osbs.constants import OS_CONFLICT_MAX_RETRIES, OS_CONFLICT_WAIT
+from osbs.constants import (OS_CONFLICT_MAX_RETRIES, OS_CONFLICT_WAIT,
+                            GIT_MAX_RETRIES, GIT_BACKOFF_FACTOR)
+
+
 from six.moves import http_client
 from six.moves.urllib.parse import urlparse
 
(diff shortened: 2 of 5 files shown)


osbs/constants.py
```diff
@@ -112,3 +112,9 @@
 
 # optional key path for filtering existing build config results
 FILTER_KEY = 'spec.source.git.uri'
+
+# number of retries for git clone operations
+GIT_MAX_RETRIES = 3
+
+# backoff factor for git clone operations - exponential backoff in seconds
+GIT_BACKOFF_FACTOR = 5
```
twaugh: Wonder if 60s is a better value here. Each clone operation can take from several seconds up to a couple of minutes. The likely reason for failure is high load on the git server, so even the largest backoff of 40s seems pretty fast. Changing the backoff factor to 60 means we'll wait 15min in total which seems reasonable.

osbs/utils.py
```diff
@@ -188,29 +191,36 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
...
-            if git_branch:
-                args += ['-b', git_branch]
-
-            args.append(repo_path)
-            run_command(args)
-        except OsbsException as ex:
-            raise OsbsException("Unable to clone git repo '%s' "
-                                "branch '%s'" % (git_uri, git_branch),
-                                cause=ex, traceback=sys.exc_info()[2])
+        for counter in range(max_retries + 1):
```
MartinBasti: Could we separate retry logic from git clone logic here?

* construction of git checkout command is immutable and can be defined out of cycle

```
cmd = construct_git_checkout(...)
try:
    subprocess_call_with_retry(cmd, ....)  # emits subprocess.CalledProcessError when all tries failed
except subprocess.CalledProcessError as ex:
    raise OsbsException(...)
```

osbs/utils.py
```diff
@@ -188,29 +191,36 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
 
 
 @contextlib.contextmanager
-def checkout_git_repo(git_uri, git_ref, git_branch=None):
+def checkout_git_repo(git_uri, git_ref, git_branch=None, max_retries=GIT_MAX_RETRIES,
+                      cooldown=GIT_BACKOFF_FACTOR):
```
MartinBasti: can we just name this kwarg `backoff_factor`? :)

athos-ribeiro: I believe the name here is good to show the variable is related to git. We also have similar variables for HTTP

athos-ribeiro: Oh, I see what you mean. I will change it :)

athos-ribeiro:
- git-core was already a requirement for osbs-client. It was missing in the spec file. We have to add it here so it gets installed before tests.
- Added a couple lines to set tests git author to make CI happy

osbs/utils.py
```diff
@@ -187,30 +190,44 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
                 orig[k] = v
 
 
+def check_output_with_retry(args, max_retries, backoff_factor):
```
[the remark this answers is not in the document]
MartinBasti: or it can be named just `run_command`, wit default option it will run once, retries on demand.

osbs/utils.py
```diff
@@ -187,30 +190,44 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
...
+    for counter in range(max_retries + 1):
+        try:
+            subprocess.check_output(args)
+            break
+        except subprocess.CalledProcessError as ex:
+            if counter != max_retries:
```
MartinBasti: I have just bad feeling from this, is safer to use `counter < max_retries`

osbs/utils.py
```diff
@@ -187,30 +190,44 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
...
+                logger.info("retrying command '%s':\n '%s'", args, ex.output)
```
MartinBasti: would be nice here to add info `retrying command in X seconds` if sleep time is 0<.

athos-ribeiro:
Release notes added:
- Retry `git clone` operations on failure

PR #837, opened 2018-12-13
Log proper backoff messages on git clone

athos-ribeiro:


osbs/utils.py
```diff
@@ -205,10 +205,14 @@ def subprocess_check_output_with_retry(args, max_retries=0, backoff_factor=0):
         except subprocess.CalledProcessError as ex:
             if counter < max_retries:
                 backoff = backoff_factor * (2 ** counter)
-                if backoff > 0:
-                    logger.info("retrying command '%s' in backoff seconds:\n '%s'", args, ex.output)
-                else:
+                if backoff <= 0:
                     logger.info("retrying command '%s':\n '%s'", args, ex.output)
+                elif backoff == 1:
```
twaugh: Can avoid this test by wording it as "in {}s".

MartinBasti:
Can we drop completely logic about handling wording second/seconds?  I think that everybody can handle to see `1 seconds` in a log (or use `second(s)`)

osbs/utils.py
```diff
@@ -206,7 +206,8 @@ def subprocess_check_output_with_retry(args, max_retries=0, backoff_factor=0):
             if counter < max_retries:
                 backoff = backoff_factor * (2 ** counter)
                 if backoff > 0:
-                    logger.info("retrying command '%s' in backoff seconds:\n '%s'", args, ex.output)
+                    logger.info("retrying command '%s' in %s second(s):\n '%s'",
```
twaugh: This looks awkward to me. Why not just the simpler `"... in %ss:\n '%s'"`? "1s", "30s" etc are not even a compromise, they are just how those durations are written.

athos-ribeiro:
No release notes added. This was a fix for a bug introduced in this release cycle.

PR #839, opened 2019-01-08
Capture and re-raise DockerfileParser IOError

athos-ribeiro:
If osbs-client fails when a Dockerfile cannot be parsed, it fails with

`osbs - ERROR - Exception caught: OsbsException: IOError(2, 'No such file or directory') (from IOError(2, 'No such file or directory'))`

This commit captures a possible IOError when DockerfileParser tries to
read a Dockerfile and re-raise it with a friendly error message.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -516,7 +516,12 @@ class OSBS(object):
 
     def _check_labels(self, repo_info):
         df_parser = repo_info.dockerfile_parser
-        labels = utils.Labels(df_parser.labels)
+        # DockerfileParse does not ensure a Dockerfile exists during initialization
+        try:
+            labels = utils.Labels(df_parser.labels)
+        except IOError as e:
+            raise RuntimeError('Could not parse Dockerfile in {}: {}'
+                               .format(df_parser.dockerfile_path, e))
 
         required_missing = False
         req_labels = {}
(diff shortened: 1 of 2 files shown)


osbs/api.py
```diff
@@ -516,7 +516,11 @@ def _create_build_config_and_build(self, build_request):
 
     def _check_labels(self, repo_info):
         df_parser = repo_info.dockerfile_parser
-        labels = utils.Labels(df_parser.labels)
+        # DockerfileParse does not ensure a Dockerfile exists during initialization
+        try:
+            labels = utils.Labels(df_parser.labels)
+        except IOError:
+            raise IOError('Could not parse Dockerfile in %s' % df_parser.dockerfile_path)
```
MartinBasti: I'd like to see reason of failure in log entry (not found, permission, etc...)

Is a good idea to reuse IOError? It is stdlib exception and somehow violates it usage (i.e. by not setting the errno and other OS related stuff) 
```
except IOError as e:
    raise RuntimeError('Could not parse Dockerfile in {}: {}'.format(df_parser.dockerfile_path, e))
```

athos-ribeiro:
Release notes added:

-  Provide better error messages when Dockerfile is missing

PR #841, opened 2019-01-14
Allow isolated builds for custom base images

lcarva:
This is critical to allow Freshmaker to rebuild container images to
address CVEs.


athos-ribeiro:
Release notes added:

- - Allow isolated builds for custom base images (this may be used to address CVEs related rebuilds)

PR #842, opened 2019-01-17
Do not package customization files

### LINKED ISSUE #648: Don't package customization files?
I've noticed that all the `*_customize.json` files are being packaged. This means installing/uninstalling osbs-client (either via yum or pip) causes the customization files to be modified.

Do we want this behavior? I was under the impression that these files behaved similarly to a configuration file.

athos-ribeiro:
Closes #648

### DIFF
--- a/osbs-client.spec
+++ b/osbs-client.spec
@@ -15,7 +15,7 @@
 %global with_check 1
 %endif
 
-%global commit 5fa7fa287df56468146f13285c4ea42f557e154a
+%global commit cfcf69702c0f8e849c6d3624d119dbf0d4443498
 %global shortcommit %(c=%{commit}; echo ${c:0:7})
 
 # set to 0 to create a normal release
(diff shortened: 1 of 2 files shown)


osbs-client.spec
```diff
@@ -15,7 +15,7 @@
 %global with_check 1
 %endif
 
-%global commit 5fa7fa287df56468146f13285c4ea42f557e154a
+%global commit cfcf69702c0f8e849c6d3624d119dbf0d4443498
```
twaugh: Is this intentional? I think we generally only change this when cutting a new release.

athos-ribeiro: It was intentional. After the release was cut, version was bumped to 0.54-0.dev.git.%{shorcommit}, which led me to believe this would not be an issue.

Without pointing the spec file to the new reference, rpmbuild will throw warnings about duplicated entries in %files (although the customization files will not get packaged).

athos-ribeiro:
Release notes:
- Do not distribute customization files (*_customize.json) downstream (pypi/rpm)

PR #843, opened 2019-01-23
Fix exception indentation when waiting for build

athos-ribeiro:


athos-ribeiro:
Release note added:

## Bug Fixes
- raise exception when wait for build to be scheduled times out

PR #844, opened 2019-01-23
change note in stream_logs about which errors are silenced

rcerven:


athos-ribeiro:
@rcerven I believe you can also remove the IncompleteRead from the comment.

LGTM

rcerven:
not updating RN, as it is just comment change

PR #845, opened 2019-01-28
minimize the work of git clone

mlangsdorf:
Reduce the amount of load on the git servers:
* Pass the git repo configuration information (container.yaml, additional tags, and such) to atomic reactor as part of the user parameters of the build JSON, so that atomic reactor doesn't need to reclone the repo to get that information.
* Refactor plugins configuration slightly so that the additional tags for `tag_from_config` are taken from the build configuration, instead of recloning the git repo
* Calculate the depth of the commit when cloning the repo for the first time, and pass that information to as part of the git repo configuration information.  Subsequent clones of the repo clone (for worker builds, etc) clone using --depth, dramatically reducing the amount of data that needs to be transferred.

Atomic reactor will need changes to take advantage of this work.

### DIFF
--- a/osbs/build/user_params.py
+++ b/osbs/build/user_params.py
@@ -55,6 +55,7 @@ class BuildUserParams(BuildCommon):
         self.yum_repourls = BuildParam("yum_repourls")
         self.tags_from_yaml = BuildParam('tags_from_yaml', allow_none=True)
         self.additional_tags = BuildParam('additional_tags', allow_none=True)
+        self.git_commit_depth = BuildParam('git_commit_depth', allow_none=True)
 
         self.required_params = [
             self.build_json_dir,
@@ -84,10 +85,12 @@ class BuildUserParams(BuildCommon):
                    yum_repourls=None, signing_intent=None, compose_ids=None,
                    isolated=None, scratch=None, parent_images_digests=None,
                    tags_from_yaml=None, additional_tags=None,
+                   git_commit_depth=None,
                    **kwargs):
         self.git_uri.value = git_uri
         self.git_ref.value = git_ref
         self.git_branch.value = git_branch
+        self.git_commit_depth.value = git_commit_depth
         self.tags_from_yaml.value = tags_from_yaml
         self.additional_tags.value = additional_tags or set()
 
--- a/osbs/build/user_params.py
+++ b/osbs/build/user_params.py
@@ -157,7 +157,11 @@ class BuildUserParams(BuildCommon):
     def from_json(self, user_params_json):
         if not user_params_json:
             return
-        json_dict = json.loads(user_params_json)
+        try:
+            json_dict = json.loads(user_params_json)
+        except ValueError:
+            logger.debug('failed to convert %s', user_params_json)
+            raise
         for key, value in json_dict.items():
             try:
                 self.convert_dict[key].value = value
--- a/osbs/build/user_params.py
+++ b/osbs/build/user_params.py
@@ -53,6 +53,9 @@ class BuildUserParams(BuildCommon):
         self.signing_intent = BuildParam('signing_intent', allow_none=True)
         self.trigger_imagestreamtag = BuildParam('trigger_imagestreamtag')
         self.yum_repourls = BuildParam("yum_repourls")
+        self.tags_from_yaml = BuildParam('tags_from_yaml', allow_none=True)
+        self.additional_tags = BuildParam('additional_tags', allow_none=True)
+
         self.required_params = [
             self.build_json_dir,
             self.build_type,
@@ -80,10 +83,14 @@ class BuildUserParams(BuildCommon):
                    flatpak=None, reactor_config_map=None, reactor_config_override=None,
                    yum_repourls=None, signing_intent=None, compose_ids=None,
                    isolated=None, scratch=None, parent_images_digests=None,
+                   tags_from_yaml=None, additional_tags=None,
                    **kwargs):
         self.git_uri.value = git_uri
         self.git_ref.value = git_ref
         self.git_branch.value = git_branch
+        self.tags_from_yaml.value = tags_from_yaml
+        self.additional_tags.value = additional_tags or set()
+
         self.user.value = user
         self.component.value = component
         self.release.value = release
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -279,12 +279,12 @@ def reset_git_repo(target_dir, git_reference, retry_depth=None):
     :return: str and int, commit ID of HEAD and commit depth of git_reference
     """
     deepen = retry_depth or 0
-    commit_depth = None
+    base_commit_depth = 0
     for _ in range(GIT_FETCH_RETRY):
         try:
             if not deepen:
                 cmd = ['git', 'rev-list', '--count', git_reference]
-                commit_depth = int(subprocess.check_output(cmd, cwd=target_dir))
+                base_commit_depth = int(subprocess.check_output(cmd, cwd=target_dir)) - 1
             cmd = ["git", "reset", "--hard", git_reference]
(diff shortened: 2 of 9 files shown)


osbs/utils.py
```diff
@@ -252,7 +257,7 @@ def checkout_git_repo(git_uri, git_ref, git_branch=None, max_retries=GIT_MAX_RET
             raise OsbsException("Unable to reset branch to '%s'" % git_ref,
                                 cause=ex, traceback=sys.exc_info()[2])
 
-        yield repo_path
+        yield repo_path, depth
```
MartinBasti: I'd like to return at least named tuple

osbs/build/user_params.py
```diff
@@ -80,10 +85,17 @@ def set_params(self,
...
                    isolated=None, scratch=None, parent_images_digests=None,
+                   tags_from_yaml=None, additional_tags=None, container_yaml=None,
+                   git_commit_depth=-1,
                    **kwargs):
         self.git_uri.value = git_uri
         self.git_ref.value = git_ref
         self.git_branch.value = git_branch
+        self.git_commit_depth.value = git_commit_depth
+        self.tags_from_yaml.value = tags_from_yaml
+        self.additional_tags.value = additional_tags or {}
```
MartinBasti: did you mean `or set()` IIRC additional_tags var has type set.

osbs/build/user_params.py
```diff
@@ -80,10 +85,17 @@ def set_params(self,
...
+        self.container_yaml.value = container_yaml or {}
```
MartinBasti: I haven't seen this attribute used in this PR, only in init methods. What it does and what is type of that attr?

mlangsdorf: container_yaml is the existing RepoInfo's RepoConfiguration's container member, which is the yaml.load of REPO_CONTAINER_CONFIG.  So it should be the entire existing container.yaml, read once and passed around to atomic reactor so atomic reactor doesn't need to reclone the repo just to read the container.yaml.

I am currently running into the issue of how to pass something from user_params back up to input_osv3.

osbs/utils.py

osbs/build/user_params.py
mlangsdorf: Fixed, thanks.

osbs/utils.py
```diff
@@ -244,6 +247,9 @@ def checkout_git_repo(git_uri, git_ref, git_branch=None, max_retries=GIT_MAX_RET
                                 "branch '%s'" % (git_uri, git_branch),
                                 cause=ex, traceback=sys.exc_info()[2])
 
+        if git_branch and not depth:
+            git_depth_cmd = 'git log --oneline ...{0} | wc -l'.format(git_ref)
```
athos-ribeiro: Is git_ref being sanitized here?
Otherwise this would allow a user to execute arbitrary commands in the cluster.
Maybe we could drop the `git log --oneline` in favor of  `git rev-list --count` -1?

mlangsdorf: Didn't know about `git rev-list --count` and will happily use that instead.

tests/test_utils.py
```diff
@@ -558,9 +579,16 @@ def initialize_git_repo(rpath, files=[]):
...
     subprocess.Popen(['git', 'config', 'user.name', '"Gerald Host"'], cwd=rpath)
     subprocess.Popen(['git', 'config', 'user.email', '"ghost@example.com"'], cwd=rpath)
+    first_commit_ref = None
     for f in files:
+        subprocess.Popen(['touch', f], cwd=rpath)
         subprocess.Popen(['git', 'add', f], cwd=rpath)
+        subprocess.Popen(['git', 'commit', '-m', 'new file {0}'.format(f)], cwd=rpath)
+        if not first_commit_ref:
+            shortlog = subprocess.check_output(['git', 'log', '--oneline'], cwd=rpath)
+            first_commit_ref = shortlog.split(" ")[0]
```
athos-ribeiro: This fails for python 3 [1] with `TypeError: a bytes-like object is required, not 'str'` since check_output returns a byte string. You could check the `universal_newlines` param [2] or perform a conversion (using a byte literal b' ' should be ignored in python 2.

[1] https://travis-ci.org/projectatomic/osbs-client/jobs/485979150
[2] https://docs.python.org/3/library/subprocess.html#subprocess.check_output

tests/test_utils.py
```diff
@@ -519,23 +519,44 @@ def test_strip_registry_and_tag_from_image(img, expected):
...
+    with checkout_git_repo(repo_path, 'HEAD') as repo_info:
+        repo = repo_info[0]
         assert os.path.isdir(repo)
         assert os.path.isdir(os.path.join(repo, '.git'))
         assert subprocess.check_output(['git', 'status'], cwd=repo)
 
+    deep_repo_path = tmpdir.mkdir("deep_repo").strpath
+    first_commit_ref = initialize_git_repo(deep_repo_path, ['file1', 'file2', 'file3', 'file4'])
+    with checkout_git_repo(deep_repo_path, first_commit_ref, 'master') as deep_repo_info:
+        deep_repo = deep_repo_info[0]
```
MartinBasti: I hoped for accessing attributes by name
```python
deep_repo = deep_repo_info.repo_path 
depth = deep_repo_info.commit_depth
```

mlangsdorf: Fixed, thanks.

tests/test_utils.py
mlangsdorf: using rev-parse instead.

osbs/utils.py
```diff
@@ -233,7 +234,9 @@ def checkout_git_repo(git_uri, git_ref, git_branch=None, max_retries=GIT_MAX_RET
     # git will remove the empty directory (why?!)
     args = ['git', 'clone', git_uri]
     if git_branch:
-        args += ['-b', git_branch]
+        args += ['-b', git_branch, '--single-branch']
+    if depth > 0:
```
athos-ribeiro: If depth assumes its default value (None), it will fail here in Python 3.

osbs/utils.py
```diff
@@ -216,7 +217,7 @@ def subprocess_check_output_with_retry(args, max_retries=0, backoff_factor=0):
 
 @contextlib.contextmanager
 def checkout_git_repo(git_uri, git_ref, git_branch=None, max_retries=GIT_MAX_RETRIES,
-                      backoff_factor=GIT_BACKOFF_FACTOR):
+                      backoff_factor=GIT_BACKOFF_FACTOR, depth=None):
     """
     Context manager to check out a git repository, yielding a string with the path
     for the cloned repository.
```
athos-ribeiro: It would be nice to add an entry for depth in the docstring

osbs/utils.py

osbs/utils.py

rcerven:
LGTM, but waiting for others to approve since they had comments

osbs/build/user_params.py
```diff
@@ -157,7 +167,11 @@ def __repr__(self):
     def from_json(self, user_params_json):
         if not user_params_json:
             return
-        json_dict = json.loads(user_params_json)
+        try:
+            json_dict = json.loads(user_params_json)
+        except TypeError:
```
rcerven: missing coverage

mlangsdorf:
Testing showed that the git depth calculation was incorrect.  The 4th commit fixes that.

osbs/build/user_params.py
mlangsdorf: Fixed, thanks.

mlangsdorf:
added to RN under build JSON changes:
- add tags_from_yaml and git commit depth parameters.  Use tags_from_yaml to eliminate an unnecessary git clone operation when setting tags for atomic-reactor plugins.

PR #846, opened 2019-02-01
utils: refactor git clone operations to share code with atomic-reactor

mlangsdorf:
merge the logic of atomic-reactor's clone_git_repo with osbs-client's
checkout_git_repo while adding support for passing depth values to
both and performing shallow clone operations if possible.

### DIFF
--- a/osbs/repo_utils.py
+++ b/osbs/repo_utils.py
@@ -43,10 +43,11 @@ class RepoConfiguration(object):
         enabled = false
         """)
 
-    def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
+    def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE, depth=None):
 
         self._config_parser = ConfigParser()
         self.container = {}
+        self.depth = depth or 0
 
         # Set default options
         self._config_parser.readfp(StringIO(self.DEFAULT_CONFIG))
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -27,7 +27,8 @@ from io import BytesIO
 from hashlib import sha256
 from osbs.repo_utils import RepoConfiguration, RepoInfo, AdditionalTagsConfig
 from osbs.constants import (OS_CONFLICT_MAX_RETRIES, OS_CONFLICT_WAIT,
-                            GIT_MAX_RETRIES, GIT_BACKOFF_FACTOR)
+                            GIT_MAX_RETRIES, GIT_BACKOFF_FACTOR,
+                            GIT_FETCH_RETRY)
 
 
 from six.moves import http_client
@@ -48,6 +49,7 @@ from dockerfile_parse import DockerfileParser
 from osbs.exceptions import OsbsException, OsbsResponseException, OsbsValidationException
 
 logger = logging.getLogger(__name__)
+ClonedRepoData = namedtuple('ClonedRepoData', ['repo_path', 'commit_id', 'commit_depth'])
 
 
 class RegistryURI(object):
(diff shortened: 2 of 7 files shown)


tests/test_utils.py
```diff
@@ -517,26 +532,47 @@ def test_strip_registry_and_tag_from_image(img, expected):
...
+def test_clone_git_repo_commit_failure(tmpdir, commit, branch, depth):
+    tmpdir_path = str(tmpdir.realpath())
+    with pytest.raises(OsbsException) as exc:
+        clone_git_repo(TEST_DOCKERFILE_GIT, tmpdir_path, commit=commit, retry_times=1,
+                       branch=branch, depth=depth)
+    assert 'cannot find commit' in exc.value.message
+
+
+@requires_internet
+def test_clone_git_repo_total_failure(tmpdir):
```
athos-ribeiro: This test is prompting a GitHub username. Travis times out here.

mlangsdorf: Fixed.

osbs/utils.py
```diff
@@ -190,72 +192,120 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
...
+    commit = commit or "master"
+    logger.info("cloning git repo '%s'", git_url)
+    logger.debug("url = '%s', dir = '%s', commit = '%s'",
+                 git_url, target_dir, commit)
+
+    cmd = ["git", "clone"]
+    if branch:
+        cmd += ["-b", branch, "--single-branch"]
+        if depth:
+            cmd += ["--depth", '{0}'.format(depth)]
```
twaugh: Any reason not to use `str(depth)` here?

osbs/utils.py
```diff
@@ -190,72 +192,120 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
...
+    elif depth:
+        logger.warning("branch not provided for %s, depth setting ignored", git_url)
+        depth = None
+
+    cmd += [git_url, quote(target_dir)]
```
twaugh: Why do we need to quote this when we're passing in all the words as a vector?

osbs/utils.py
```diff
@@ -190,72 +192,120 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
...
-        except subprocess.CalledProcessError as ex:
-            raise OsbsException("Unable to clone git repo '%s' "
-                                "branch '%s'" % (git_uri, git_branch),
-                                cause=ex, traceback=sys.exc_info()[2])
+    hard reset git clone in target_dir to given git_reference
 
-        # Find the specific ref we want
+    :param target_dir: str, filesystem path where the repo is cloned
+    :param git_reference: str, any valid git reference
+    :param retry_deeper: int, if the repo was cloned with --shallow, this is the expected
```
twaugh: Should be `retry_depth`

osbs/utils.py
```diff
@@ -190,72 +192,120 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
...
+            cmd = ["git", "reset", "--hard", git_reference]
+            logger.debug("Resetting current HEAD: '%s'", cmd)
+            subprocess.check_call(cmd, cwd=target_dir)
+            break
+        except subprocess.CalledProcessError:
+            if not deepen:
+                raise OsbsException('cannot find commit %s in repo %s' %
+                                    (git_reference, target_dir))
+            deepen *= 2
+            cmd = ["git", "fetch", "--depth", "{0}".format(deepen)]
```
twaugh: Would it be useful to log that we're doing this?

osbs/utils.py
```diff
@@ -190,72 +192,120 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
...
+            subprocess.check_call(cmd, cwd=target_dir)
+    else:
+        raise OsbsException('cannot find commit %s in repo %s' % (git_reference, target_dir))
 
-        yield repo_path
+    cmd = ["git", "rev-parse", "HEAD"]
+    logger.debug("getting SHA-1 of provided ref '%s'", cmd)
```
twaugh: Looks like this will log `getting SHA-1 of provided ref '['git', 'rev-parse', 'HEAD']'`. Is that useful? Also, the command does not retrieve the hash for the *provided ref*, but for the commit we reset to... which the caller already knows.

I'm not sure why we're returning this value (or running the command).

osbs/utils.py
mlangsdorf: I'm bad at basic Python idiom.

osbs/utils.py
mlangsdorf: `reset_git_repo()` can take any valid git_reference, so things like HEAD^ can be the argument, in which case you don't know the hash.

I'm not sure I follow your objection.  We reset the HEAD to some arbitrary git reference, which may or may not be a hash, and then we return the hash of the new head.
ie:
```
$ git reset --hard master
HEAD is now at 031467b Merge pull request #112 from rcerven/metadata_for_scratch
$ git rev-parse HEAD
031467b48f8139935a7d4ab8b6ded9c30b653529
$ git reset --hard HEAD^^^
HEAD is now at 199d96c Automatic commit of package [koji-containerbuild] minor release [0.7.12-1].
$ git rev-parse HEAD
199d96ca229edfcf9f81fca4c5304823bfebd2f6
```

osbs/utils.py
mlangsdorf: I can add a log.

osbs/utils.py
twaugh: Ah, git_reference is not necessary a hash -- I had overlooked that. Thanks.
But is `getting SHA-1 of provided ref '['git', 'rev-parse', 'HEAD']'` what we want to log? Do we want to log what the provided ref actually was?

osbs/repo_utils.py
```diff
@@ -43,10 +43,11 @@ class RepoConfiguration(object):
         enabled = false
         """)
 
-    def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE):
+    def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE, depth=None):
 
         self._config_parser = ConfigParser()
         self.container = {}
+        self.depth = depth or -1
```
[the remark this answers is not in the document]
mlangsdorf: Fixed.

osbs/utils.py
mlangsdorf: That's how it was done.  Removed.

osbs/utils.py
mlangsdorf: Fixed, thanks.

tests/build_/test_arrangements.py
```diff
@@ -96,6 +96,7 @@ class MockConfiguration(object):
 
             module = container['compose']['modules'][0]
             container_module_specs = [ModuleSpec.from_str(module)]
+            depth = -1
```
twaugh: Should this be 0 now?

tests/test_api.py
```diff
@@ -119,6 +119,7 @@ def __init__(self, modules=['mod_name:mod_stream:mod_version']):
         self.container = {'compose': {'modules': modules}}
         safe_modules = modules or []
         self.container_module_specs = [ModuleSpec.from_str(module) for module in safe_modules]
+        self.depth = -1
```
twaugh: Should this be 0 now?

tests/build_/test_arrangements.py

osbs/utils.py
```diff
@@ -190,72 +191,122 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
...
-                                "branch '%s'" % (git_uri, git_branch),
-                                cause=ex, traceback=sys.exc_info()[2])
+    hard reset git clone in target_dir to given git_reference
 
-        # Find the specific ref we want
+    :param target_dir: str, filesystem path where the repo is cloned
+    :param git_reference: str, any valid git reference
+    :param retry_depth: int, if the repo was cloned with --shallow, this is the expected
+                        depth of the commit
+    :return: str, commit ID of HEAD
```
athos-ribeiro: It would be nice to document the second return value here (commit_depth)

osbs/utils.py
```diff
@@ -190,72 +191,122 @@ def buildconfig_update(orig, new, remove_nonexistent_keys=False):
...
+    """
+    deepen = retry_depth or 0
+    commit_depth = None
+    for _ in range(1, 10):
```
athos-ribeiro: Is there any reason for using this specific range here? Could we use just `range(9)` and set it as a constant in the constant modules?

osbs/utils.py

osbs/utils.py
mlangsdorf: No. Yes. Fixed.

mlangsdorf:
Not adding anything to the release notes because this is an internal only code change.

PR #847, opened 2019-02-05
Don't run pytest in verbose mode

MartinBasti:
We cannot read verbose output nicely in travis output because it is trimmed.


athos-ribeiro:
No release notes added. CI related change.

PR #849, opened 2019-03-04
Activate operators metadata uploader plugin

athos-ribeiro:


osbs/api.py
```diff
@@ -614,6 +614,7 @@ def _do_create_prod_build(self, git_uri, git_ref,
                               reactor_config_override=None,
                               parent_images_digests=None,
                               git_commit_depth=None,
+                              allowed_platform=None,
```
twaugh: I can't think of a better name for this :-(

inputs/worker_inner:6.json
```diff
@@ -83,6 +83,9 @@
         "publish": false
       }
     },
+    {
+      "name": "export_operator_manifests"
```
twaugh: Needs test-case change.

osbs/api.py
athos-ribeiro: maybe `operator_uploader` or `operator_uploader_platform`?

twaugh: The best I came up with was `noarch_platform`, i.e. the platform whose worker can perform operations for which it doesn't matter which platform is used. But the fact that I have to explain it probably means it's a bad name.

I prefer either of `operator_uploader` or `operator_uploader_platform` to `allowed_platform`.

twaugh: Or even `any_platform`.

MartinBasti: I vote for `operator_uploader_platform`, that names describes what is that param used for

MartinBasti: BTW, in design doc we agreed on `operator_manifests_extract_platform`

osbs/build/build_request.py
```diff
@@ -99,6 +99,7 @@ def set_params(self, **kwargs):
         :param auto_build_node_selector: dict, a nodeselector for auto builds
         :param isolated_build_node_selector: dict, a nodeselector for isolated builds
         :param is_auto: bool, indicates if build is auto build
+        :param allowed_platform: str, indicates which platform should upload operator manifests
```
mlangsdorf: is this a valid parameter to a v1 build_request?  You can't run the operators metadata plugin in any of the arrangements supported by v1 build_request.

osbs/build/user_params.py
```diff
@@ -125,6 +126,7 @@ def set_params(self,
                 self.build_imagestream.value = source_value
 
         self.parent_images_digests.value = parent_images_digests
+        self.allowed_platform.value = allowed_platform
```
mlangsdorf: you'll also need to add allowed_platform to atomic-reactor's user_params schema.

athos-ribeiro: Done in https://github.com/projectatomic/atomic-reactor/pull/1177
Thanks for catching this one :)

osbs/build/build_request.py
athos-ribeiro: Removed

osbs/utils.py
```diff
@@ -606,7 +608,8 @@ class Labels(object):
         LABEL_TYPE_HOST: ('com.redhat.build-host', 'Build_Host'),
         LABEL_TYPE_RUN: ('run', 'RUN'),
         LABEL_TYPE_INSTALL: ('install', 'INSTALL'),
-        LABEL_TYPE_UNINSTALL: ('uninstall', 'UNINSTALL')
+        LABEL_TYPE_UNINSTALL: ('uninstall', 'UNINSTALL'),
+        LABEL_TYPE_OPERATOR_MANIFESTS: ('com.redhat.delivery.appregistry',)
```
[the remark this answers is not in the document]
rcerven: generally we using it for labels which have variants, but we also use this, as only place where we define label names, so this way if in the future there will be added new name we change it just here and we are done

athos-ribeiro:
Release notes added:

## Build JSON changes
- add operator_manifests_extract_platform parameter to indicate which platform should upload operator manifests to koji

## Improvements
- Run new export_operator_manifests plugin in worker builds to upload files
- Add support for `com.redhat.delivery.appregistry` label in Dockerfile. When set to 'true', this label indicates the presence of operator manifests in the built image

PR #852, opened 2019-04-11
Fix py2/3 compatibility issues

chmeliik:
You don't want to know how long it took me to figure out why `from six.moves import input` broke tests.


MartinBasti:
Release notes: 
Bugfixes

> - python 2/3 compatibility

PR #853, opened 2019-04-23
Add hide files plugin

rcerven:
* OSBS-7298

### DIFF
--- a/inputs/worker_inner:6.json
+++ b/inputs/worker_inner:6.json
@@ -28,6 +28,9 @@
       "name": "change_from_in_dockerfile"
     },
     {
+      "name": "hide_files"
+    },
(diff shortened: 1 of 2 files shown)


inputs/orchestrator_inner:6.json
```diff
@@ -33,6 +33,9 @@
         "inspect_only": true
       }
     },
+    {
+      "name": "hide_files"
```
lcarva: I don't see why we need to run this on the orchestrator build. We need it in the worker build only.

rcerven: fixed

rcerven:
rn updated: - run new hide files plugin

PR #854, opened 2019-04-25
floating tags

rcerven:
OSBS-7169


rcerven:
@mlangsdorf  PTAL

twaugh:
Release notes drafts updated:
- better internal separation of floating tags and primary tags

and bumped minimum atomic-reactor version to 1.6.39.

PR #855, opened 2019-05-02
Use str() to report ScannerError

chmeliik:
* OSBS-5904

### DIFF
--- a/osbs/repo_utils.py
+++ b/osbs/repo_utils.py
@@ -9,6 +9,7 @@ of the BSD license. See the LICENSE file for details.
 
 from __future__ import absolute_import
 
+from osbs.exceptions import OsbsException
 from osbs.constants import REPO_CONFIG_FILE, ADDITIONAL_TAGS_FILE, REPO_CONTAINER_CONFIG
 from six import StringIO
 from six.moves.configparser import ConfigParser
@@ -61,7 +62,12 @@ class RepoConfiguration(object):
         file_path = os.path.join(dir_path, REPO_CONTAINER_CONFIG)
         if os.path.exists(file_path):
             with open(file_path) as f:
-                self.container = (yaml.load(f) or {})
+                try:
+                    self.container = yaml.load(f) or {}
+                except yaml.scanner.ScannerError as e:
+                    msg = ('Failed to parse YAML file "{file}": {reason}'
+                           .format(file=REPO_CONTAINER_CONFIG, reason=e))
+                    raise OsbsException(msg)
 
         # container values may be set to None
         container_compose = self.container.get('compose') or {}
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -46,9 +46,10 @@ from osbs.constants import (DEFAULT_OUTER_TEMPLATE, WORKER_OUTER_TEMPLATE,
                             ORCHESTRATOR_CUSTOMIZE_CONF,
                             BUILD_TYPE_WORKER, BUILD_TYPE_ORCHESTRATOR,
                             OS_CONFLICT_MAX_RETRIES,
-                            ANNOTATION_SOURCE_REPO, ANNOTATION_INSECURE_REPO)
+                            ANNOTATION_SOURCE_REPO, ANNOTATION_INSECURE_REPO,
+                            REPO_CONTAINER_CONFIG)
 from osbs import utils
-from osbs.repo_utils import RepoInfo, ModuleSpec
+from osbs.repo_utils import RepoInfo, RepoConfiguration, ModuleSpec
 
 from tests.constants import (TEST_ARCH, TEST_BUILD, TEST_COMPONENT, TEST_GIT_BRANCH, TEST_GIT_REF,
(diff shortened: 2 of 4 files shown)


tests/test_api.py
```diff
@@ -2943,3 +2943,32 @@ def labels(self):
             kwargs['flatpak'] = True
             response = osbs._do_create_prod_build(**kwargs)
             assert isinstance(response, BuildResponse)
+
+    def test_do_create_prod_build_invalid_yaml(self, osbs, tmpdir, monkeypatch):  # noqa
+        yaml_file = tmpdir.join('invalid.yaml')
+        yaml_file.write('\n'.join(['hallo: 1', 'bye']))
+
+        def get_repo_info(*_, **__):
```
MartinBasti: Is this a standardized way of doing `*args, **kwargs` for unused arguments?

chmeliik: I don't know, probably not.

MartinBasti: I'd stay with standard `*args, **kwargs` which are commonly used

chmeliik: ok

osbs/api.py
```diff
@@ -625,8 +626,13 @@ def _do_create_prod_build(self, git_uri, git_ref,
                 # fix for module requires will have to be determined from experience.
                 raise ValueError("Flatpak build cannot be isolated")
 
-        repo_info = utils.get_repo_info(git_uri, git_ref, git_branch=git_branch,
-                                        depth=git_commit_depth)
+        try:
+            repo_info = utils.get_repo_info(git_uri, git_ref, git_branch=git_branch,
+                                            depth=git_commit_depth)
+        except yaml.scanner.ScannerError as e:
```
MartinBasti: This doesn't seem right to me. This context should have no information about implementation detail, that we use yaml parser.

IMHO we should update `OSBSException` to use message from `str()` instead of `repr()` in constructor

mlangsdorf: If it's better to use str to get the error, we should do this inside get_repo_info.

chmeliik: So, any opinions on changing `OSBSException` to use `str()` vs moving the error handling to `get_repo_info()`?

MartinBasti: IMO, I'd go with OSBSException as our goal is to always use str() instead of repr() in future.

But on the other hand, exception context should be isolated, so if we know that somewhere in `get_rep_info` is expected to get Yaml exception, we should catch it and raise as OSBS exception

chmeliik: So I'm guessing I should do both?

Also, the actual line where the error happens is [in osbs.repo_utils.RepoConfiguration.\_\_init\_\_()](https://github.com/projectatomic/osbs-client/blob/f2e2ae3029bc368e4b6c1c07b420b39b24b5d61f/osbs/repo_utils.py#L64), so maybe the the try/except should be there?

MartinBasti: Yes, that's sounds about right

chmeliik:
Ready for review

osbs/repo_utils.py
```diff
@@ -61,7 +62,10 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE, depth=None):
         file_path = os.path.join(dir_path, REPO_CONTAINER_CONFIG)
         if os.path.exists(file_path):
             with open(file_path) as f:
-                self.container = (yaml.load(f) or {})
+                try:
+                    self.container = yaml.load(f) or {}
+                except yaml.scanner.ScannerError as e:
+                    raise OsbsException(str(e))
```
MartinBasti: Could be this err msg more descriptive, like "Failed to load YAML file {path}: {reason}"

MartinBasti: Respectively only related path `REPO_CONTAINER_CONFIG`, users don't care about our absolute/temp paths

twaugh: The reported error will look like this:
```
OsbsException('while scanning a simple key\n  in "/tmp/pytest-of-twaugh/pytest-3/test_create_build_invalid_yaml0/container.yaml", line 2, column 1\ncould not find expected \':\'\n  in "/tmp/pytest-of-twaugh/pytest-3/test_create_build_invalid_yaml0/container.yaml", line 2, column 4',)
```

I can't see a neat way to avoid repeating the absolute filename there, do you?

MartinBasti: Ah right, but still would be nice to see there that error is from YAML parsing

MartinBasti: `OsbsException('Failed to load YAML file containers.yaml: while scanning a simple key\n  in "/tmp/pytest-of-twaugh/pytest-3/test_create_build_invalid_yaml0/container.yaml", line 2, column 1\ncould not find expected \':\'\n  in "/tmp/pytest-of-twaugh/pytest-3/test_create_build_invalid_yaml0/container.yaml", line 2, column 4',)`

chmeliik:
Updated the error message as @MartinBasti requested.

MartinBasti:
Thank you!

MartinBasti:
Release notes:

> - Improved error reporting of YAML syntax errors

PR #856, opened 2019-05-02
Disallow usage of prod builds and arrangements <= 5

chmeliik:
* OSBS-6491

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -303,10 +303,8 @@ def cmd_cancel_build(args, osbs):
 def cmd_build(args, osbs):
     if args.worker:
         create_func = osbs.create_worker_build
-    elif args.orchestrator:
-        create_func = osbs.create_orchestrator_build
     else:
-        create_func = osbs.create_prod_build
+        create_func = osbs.create_orchestrator_build
 
     build_kwargs = {
         'git_uri': osbs.build_conf.get_git_uri(),
@@ -719,9 +717,7 @@ def cli():
     build_type_group.add_argument("--worker", action="store_true", required=False,
                                   default=False, help="create worker build")
     build_type_group.add_argument("--orchestrator", action="store_true", required=False,
-                                  default=False, help="create orchestrator build")
-    build_type_group.add_argument("--prod", action="store_true", required=False,
-                                  default=True, help="create prod build")
+                                  default=True, help="create orchestrator build")
 
     group = build_parser.add_mutually_exclusive_group()
     group.add_argument("--build-from", action='store', required=False,
--- a/tests/build_/test_arrangements.py
+++ b/tests/build_/test_arrangements.py
@@ -72,6 +72,30 @@ OSBS_WITH_PULP_PARAMS = {
     'kwargs': {'registry_uri': 'registry.example.com/v2'}}
 
 
+def unsupported_arrangement_version(version_test_class):
+    """
+    Mark a test class as unsupported to disable version validation.
+    Does not disable validation for classes that inherit from said class.
+    """
+    from osbs.api import validate_arrangement_version
+
+    def setup_class(cls):
+        import osbs.api
+        osbs.api.validate_arrangement_version = lambda version: None
+
+    def teardown_class(cls):
+        import osbs.api
+        # restore original validation logic
+        osbs.api.validate_arrangement_version = validate_arrangement_version
+        # prevent setup and teardown of child classes
+        del cls.setup_class, cls.teardown_class
+
+    version_test_class.setup_class = classmethod(setup_class)
+    version_test_class.teardown_class = classmethod(teardown_class)
+
+    return version_test_class
+
+
 class ArrangementBase(object):
     ARRANGEMENT_VERSION = None
     COMMON_PARAMS = {}
@@ -181,10 +205,10 @@ class ArrangementBase(object):
                 return registry.docker_uri
 
 
+@unsupported_arrangement_version    # TODO: move relevant tests to version 6
 class TestArrangementV1(ArrangementBase):
     """
-    This class tests support for the oldest supported arrangement
-    version, 1.
+    This arrangement version is no longer supported.
 
     NOTE! When removing this test class, *make sure* that any methods
     it provides for the test class for the next oldest supported
@@ -375,8 +399,11 @@ class TestArrangementV1(ArrangementBase):
             assert args['repos'] == params['yum_repourls']
 
 
+@unsupported_arrangement_version    # TODO: move relevant tests to version 6
 class TestArrangementV2(TestArrangementV1):
     """
+    This arrangement version is no longer supported.
+
     Differences from arrangement version 1:
     - add_filesystem runs with different parameters
     - add_filesystem also runs in orchestrator build
@@ -563,8 +590,11 @@ class TestArrangementV2(TestArrangementV1):
             assert 'koji_hub' in args
 
 
+@unsupported_arrangement_version    # TODO: move relevant tests to version 6
 class TestArrangementV3(TestArrangementV2):
     """
+    This arrangement version is no longer supported.
+
     Differences from arrangement version 2:
     - fetch_worker_metadata, koji_import, koji_tag_build, sendmail,
       check_and_set_rebuild, run in the orchestrator build
@@ -794,8 +824,11 @@ class TestArrangementV3(TestArrangementV2):
         assert match_args == args
 
 
+@unsupported_arrangement_version    # TODO: move relevant tests to version 6
 class TestArrangementV4(TestArrangementV3):
     """
+    This arrangement version is no longer supported.
+
     Orchestrator build differences from arrangement version 3:
     - tag_from_config enabled
     - pulp_tag enabled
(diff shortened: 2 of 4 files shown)


tests/build_/test_arrangements.py
```diff
@@ -181,1213 +181,83 @@ def get_pulp_sync_registry(self, conf):
...
-        }
-        if scratch:
-            additional_params['scratch'] = True
-        if worker:
-            additional_params['compose_ids'] = [42]
-            params, build_json = self.get_worker_build_request(osbs, additional_params)
-        else:
-            params, build_json = self.get_orchestrator_build_request(osbs, additional_params)
-        plugins = get_plugins_from_build_json(build_json)
+class UnsupportedArrangement(object):
```
MartinBasti: I'm curious if  is it worth to keep such testing. It has nice history overview of how arrangements version evolved but in reality it tests only validate_arrangement_version in a complicated way.

osbs/cli/main.py
```diff
@@ -306,7 +306,7 @@ def cmd_build(args, osbs):
     elif args.orchestrator:
         create_func = osbs.create_orchestrator_build
     else:
-        create_func = osbs.create_prod_build
```
MartinBasti: I'm not sure about this line, @others?

rcerven: I don't think there should be 'else'

there should be just worker & orchestrator build, that 'create_build' is used for koji-containerbuild when orchestrator isn't enabled or older osbs-client doesn't have create_orchestrator_build method

athos-ribeiro: Should we default to some of the options though (e.g, orchestrator build)? If not, we should raise an error if neither args are passed here.

MartinBasti:
Hmm, shouldn't we remove also `input/*_inner:X.json` for X<=5 ?

osbs/cli/main.py
rcerven: prod argument should be removed as well
https://github.com/projectatomic/osbs-client/blob/6e337b285f50d44e8526a779e202cd9663400011/osbs/cli/main.py#L723

mmm maybe we can default to orchestrator instead of raising

https://github.com/projectatomic/osbs-client/blob/6e337b285f50d44e8526a779e202cd9663400011/osbs/cli/main.py#L721
default=True

rcerven:
yeah we should remove unused jsons, for arrangements <=5 and prod

tests/build_/test_arrangements.py
chmeliik: Yeah, I kept those test classes mostly just for the documentation value, but if you don't want them they can go.

chmeliik:
There is also this section in `osbs/constants.py`:
```python
DEFAULT_OUTER_TEMPLATE = "prod.json"
WORKER_OUTER_TEMPLATE = "worker.json"
ORCHESTRATOR_OUTER_TEMPLATE = "orchestrator.json"
DEFAULT_INNER_TEMPLATE = "prod_inner.json"
WORKER_INNER_TEMPLATE = "worker_inner:{arrangement_version}.json"
ORCHESTRATOR_INNER_TEMPLATE = "orchestrator_inner:{arrangement_version}.json"
DEFAULT_ARRANGEMENT_VERSION = 6  # this should be the highest-numbered version
REACTOR_CONFIG_ARRANGEMENT_VERSION = 6
DEFAULT_CUSTOMIZE_CONF = "prod_customize.json"
WORKER_CUSTOMIZE_CONF = "worker_customize.json"
ORCHESTRATOR_CUSTOMIZE_CONF = "orchestrator_customize.json"
```
I assume it should look like this now?
```python
WORKER_OUTER_TEMPLATE = "worker.json"
ORCHESTRATOR_OUTER_TEMPLATE = "orchestrator.json"
WORKER_INNER_TEMPLATE = "worker_inner:{arrangement_version}.json"
ORCHESTRATOR_INNER_TEMPLATE = "orchestrator_inner:{arrangement_version}.json"
DEFAULT_ARRANGEMENT_VERSION = 6  # this should be the highest-numbered version
REACTOR_CONFIG_ARRANGEMENT_VERSION = 6
WORKER_CUSTOMIZE_CONF = "worker_customize.json"
ORCHESTRATOR_CUSTOMIZE_CONF = "orchestrator_customize.json"
DEFAULT_OUTER_TEMPLATE = ORCHESTRATOR_OUTER_TEMPLATE
DEFAULT_INNER_TEMPLATE = ORCHESTRATOR_INNER_TEMPLATE
DEFAULT_CUSTOMIZE_CONF = ORCHESTRATOR_CUSTOMIZE_CONF
```

chmeliik:
So, I removed those unused jsons and updated `osbs/constants.py` to use the orchestrator jsons by default. That broke over 70 tests (actual individual test methods, not just a few heavily parameterized ones). I've been able to fix a few, but am not sure what to do about the rest.

Most of the broken tests are in `tests/build_/test_build_request.py`, and if I am not mistaken, the `BuildRequest` class that is being tested there is not supposed to be used anymore, so maybe that file could just be dropped? Though some of the tests there may still have value, I'm not sure.

The remaining broken tests are in `tests/test_api.py`, I haven't really had time to look into those yet.

rcerven:
BuildRequestV2 inherits from BuildRequest

chmeliik:
> BuildRequestV2 inherits from BuildRequest

@rcerven Yeah, I know that, what I meant to say is that the tests for BuildRequest (old version) are now outdated, and I'm not sure if it would be OK to simply remove them all, or if some/all of them should be fixed and kept.

I figured out that many of the test failures were caused by `BuildRequest.render_add_labels_in_dockerfile`, many of the tests still fail though (see tests marked with `pytest.mark.skip`).

MartinBasti:
Please use the old configuration in tests (`prod_inner.json` probably) to keep tests passing. Fixing tests to work with arrangement 6 should be another PR to keep it cleaner.

tests/build_/test_arrangements.py
mlangsdorf: historical documentation value can be retrieved from git, so get rid of them please.

chmeliik:
So, I have good news and bad news.

The bad news:
* tests for `BuildRequest` still rely on old versions of `(orchestrator|worker)_inner` templates, so they cannot be removed (changing the arrangement version in the tests to 6 does not fix the issue)
* the defaults in `osbs/constants.py` cannot be updated yet, because it would break many tests

The good-ish news:
* the old input files can be moved to a different directory (currently `tests/old_inputs/`) and only temporarily copied to the `inputs/` directory during tests, which feels like a reasonable hack

MartinBasti:
I'd go only with first 2 commits. Let's open another issue for cleanup of tests and arrangement files

osbs/cli/main.py
```diff
@@ -719,9 +717,7 @@ def cli():
     build_type_group.add_argument("--worker", action="store_true", required=False,
                                   default=False, help="create worker build")
     build_type_group.add_argument("--orchestrator", action="store_true", required=False,
-                                  default=False, help="create orchestrator build")
-    build_type_group.add_argument("--prod", action="store_true", required=False,
```
MartinBasti: May dropping of this option cause any issues? But, well we noticed others about deprecation of this one early

rcerven:
we shouldn't remove all tests from previous arrangements in tests/build_/test_arrangements.py

for example tests: test_flatpak, test_not_flatpak are only in arrangement 4, but they are valid for arrangement 6 as well, all tests from previous arrangements which are still valid should be moved to arrangement 6

tests/test_api.py
```diff
@@ -803,10 +807,10 @@ class MockParser(object):
             .with_args(TEST_GIT_URI, TEST_GIT_REF, git_branch=TEST_GIT_BRANCH, depth=None)
             .and_return(self.mock_repo_info()))
         flexmock(OSBS, _create_build_config_and_build=request_as_response)
-        req = osbs.create_prod_build(TEST_GIT_URI, TEST_GIT_REF,
-                                     TEST_GIT_BRANCH, TEST_USER,
-                                     TEST_COMPONENT, TEST_TARGET,
-                                     TEST_ARCH)
+        req = osbs.create_build(component=TEST_COMPONENT,
```
rcerven: maybe we should actually try to pass different component than is in Dockerfile, since we can't tell now from where was that req.spec.component.value set

chmeliik:
@rcerven 
> all tests from previous arrangements which are still valid should be moved to arrangement 6

Most of those tests will not work though, because of 
```python
class TestArrangementV6(ArrangementBase):
    """
    ...

    Inherit from ArrangementBase, not the previous arrangements, because argument handling is
    different now and all previous tests break.

    ...
    """
```
In fact, the only test that would work is `test_not_flatpak`, because it does not test any plugin arguments.

tests/test_api.py
chmeliik: Nice catch, the test was actually completely broken.

rcerven:
it isn't about plugin arguments, but about the fact that we are using buildrequestv2 and user_params

but I still think we should preserve tests which are still valid

chmeliik:
Rebased on master and re-added the tests for previous arrangement versions.

The test classes for old versions are now marked with `@unsupported_arrangement_version` to temporarily disable version validation in those classes.

chmeliik:
@athos-ribeiro @MartinBasti @rcerven PR is ready for review

PR #857, opened 2019-05-28
Remove v1 content building support

athos-ribeiro:
* OSBS-6665

https://github.com/projectatomic/osbs-docs/pull/76


athos-ribeiro:
Changed

## Registry API version

- v1-only registry API is no longer supported
- URIs in `registry_uri` default to `v2` if no version is appended to the URI. Appending `/v1` to the URI is no longer supported

to

## Registry API version

- v1 registry API is no longer supported
- URIs in `registry_uri` default to `v2` if no version is appended to the URI.

and removed the following entry:


## Deprecation warning
- Support for container registry API V1 will be dropped in future releases

PR #858, opened 2019-05-30
update test_arrangements

mlangsdorf:
Addresses osbs-7503

Some of the tests for arrangement 1-5 are still valid but are no longer run because arrangements 1-5 are not valid.  Move those tests into arrangement 6, updating them as necessary.


tests/build_/test_arrangements.py
```diff
@@ -967,86 +647,42 @@ def test_tag_from_config(self, osbs, params, build_type, has_plat_tag, has_prima
 
         primary_tags = args['tag_suffixes']['primary']
         if has_primary_tag:
-            assert set(primary_tags) == set(['latest', '{version}', '{version}-{release}'])
+            assert set(primary_tags) == set(['{version}-{release}'])
 
     @pytest.mark.parametrize('osbs', [  # noqa:F811
-        {'platform_descriptors': {'x86_64': {'enable_v1': True}},
+        {
+            'platform_descriptors': {'x86_64': {'enable_v1': True}},
```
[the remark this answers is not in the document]
mlangsdorf: platform descriptors are only used in build_request_v1 code.

tests/build_/test_arrangements.py
```diff
@@ -1703,3 +1729,319 @@ def test_export_operator_manifests(self, osbs, extract_platform):
...
+    def test_koji_import(self, osbs):  # noqa:F811
+        additional_params = {
+            'base_image': 'fedora:latest',
+            'koji_upload_dir': 'upload',
+        }
+        params, build_json = self.get_orchestrator_build_request(osbs, additional_params)
+        plugins = get_plugins_from_build_json(build_json)
+
+        with pytest.raises(NoSuchPluginException):
+            get_plugin(plugins, 'postbuild_plugins', PLUGIN_KOJI_UPLOAD_PLUGIN_KEY)
```
chmeliik: Does this actually test koji_import? Maybe we could at least add something like this:
```python
get_plugin(plugins, 'exit_plugins', PLUGIN_KOJI_IMPORT_PLUGIN_KEY)
with pytest.raises(KeyError):
    plugin_value_get(plugins, 'exit_plugins', PLUGIN_KOJI_IMPORT_PLUGIN_KEY, 'args')
```

tests/build_/test_arrangements.py
```diff
@@ -1703,3 +1729,319 @@ def test_export_operator_manifests(self, osbs, extract_platform):
...
+        expected_args = {
+            'publish': False,
+        }
+
+        assert args == expected_args
+
+    @pytest.mark.parametrize('osbs, use_pulp', [  # noqa:F811
+        ({'platform_descriptors': None, 'additional_config': None, 'kwargs': None}, False),
+        (OSBS_WITH_PULP_PARAMS, True)], indirect=['osbs'])
+    def test_pulp_publish(self, osbs, use_pulp):
```
chmeliik: `use_pulp` seems to be unused here. If the outcome of the test is not expected to change based on that, is it necessary to parameterize it?

tests/build_/test_arrangements.py
```diff
@@ -1703,3 +1729,319 @@ def test_export_operator_manifests(self, osbs, extract_platform):
...
+
+    @pytest.mark.parametrize('osbs, with_group', [  # noqa:F811
+        ({'platform_descriptors': {'x86_64': {'architecture': 'amd64'}},
+            'additional_config': get_pulp_additional_config(),
+            'kwargs': {'registry_uri': 'registry.example.com/v2'}}, False),
+        ({'platform_descriptors': {'x86_64': {'architecture': 'amd64'}},
+            'additional_config': get_pulp_additional_config(with_group=True),
+            'kwargs': {'registry_uri': 'registry.example.com/v2'}}, True)],
+        indirect=['osbs'])
+    def test_group_manifests(self, osbs, with_group):
```
chmeliik: `with_group` seems to be unused here, same concern as `use_pulp` above.

tests/build_/test_arrangements.py
```diff
@@ -1703,3 +1729,319 @@ def test_export_operator_manifests(self, osbs, extract_platform):
...
+        if has_primary_tag:
+            assert set(primary_tags) == set(['{version}-{release}'])
+
+    @pytest.mark.parametrize('osbs', [  # noqa:F811
+        {
+            'additional_config': get_pulp_additional_config(),
+            'kwargs': {'registry_uri': 'registry.example.com/v2'},
+        }
+    ], indirect=True)
+    def test_pulp_push(self, osbs):
```
chmeliik: The original test has a `# TODO: remove this test after the pulp push plugin is removed`. Is that not relevant anymore?

mlangsdorf: pulp_push has not been removed.

tests/build_/test_arrangements.py
mlangsdorf: Thanks, fixed.

tests/build_/test_arrangements.py
mlangsdorf: Thanks, fixed.

tests/build_/test_arrangements.py

tests/build_/test_arrangements.py
chmeliik: Do you know if there are still plans to remove it? Maybe the TODO could be kept there, although I expect the test will start failing anyway if the plugin is removed so it's probably unnecessary.

mlangsdorf: Yes, pulp will be removed soon now that we can use quay.io instead.

chmeliik: Ok, thanks, I found the issue that we have filed for it, so the TODO is probably not needed.

tests/fake_api.py
```diff
@@ -509,7 +509,7 @@ def openshift(request):
 @pytest.fixture(params=[{'kwargs': None, 'additional_config': None, 'platform_descriptors': None}])
 def osbs(request, openshift):
     kwargs = request.param['kwargs'] or {}
-    platform_descriptors = request.param['platform_descriptors'] or {}
+    platform_descriptors = request.param.get('platform_descriptors') or {}
```
rcerven: request.param.get('platform_descriptors', {})

tests/build_/test_arrangements.py
```diff
@@ -1703,3 +1729,313 @@ def test_export_operator_manifests(self, osbs, extract_platform):
...
+        unique_tags = args['tag_suffixes']['unique']
+        assert len(unique_tags) == 1
+        unique_tag_suffix = ''
+        if has_plat_tag:
+            unique_tag_suffix = '-' + additional_params.get('platform')
+        assert unique_tags[0].endswith(unique_tag_suffix)
+
+        primary_tags = args['tag_suffixes']['primary']
+        if has_primary_tag:
+            assert set(primary_tags) == set(['{version}-{release}'])
```
rcerven: add also:
floating_tags = args['tag-suffixes']['floating']
assert(floating_tags) == set(['latest', '{version}'])

tests/fake_api.py
mlangsdorf: That's not working in my tests, I don't know why.

mlangsdorf:
```
        if platform_descriptors is None:
            print("params: {0} pds {1} adds {2}".format(request.param, request.param.get('platform_descriptors', {}), request.param.get('additional_config', {})))
            raise
```
returns
`params: {u'additional_config': None, u'platform_descriptors': None, u'kwargs': None} pds None adds None`

I don't get it.

rcerven: nvm leave what you have, I missed that the key is actually defined but set to None,
so in that case it won't use that value, since key exists

MartinBasti:
Please remove this bin file from PR ` BIN +16 KB tests/build_/.test_build_request.py.swp `

MartinBasti:
So can we now remove old arrangement json files and prod build?

mlangsdorf:
Sure, but that's not this PR.

MartinBasti:
ok, thanks

chmeliik:
@MartinBasti @mlangsdorf I'm pretty sure removing prod jsons would require at least changes to `osbs/constants.py` and possibly fixing some tests, and removing old arrangement jsons would require removing/fixing old `BuildRequest` tests (if I'm not mistaken, this PR does not remove any of those).

rcerven:
not updating release note, just tests cleanup

PR #862, opened 2019-06-11
Remove BuildSpec and BuildRequest v1

mlangsdorf:
address osbs-6573

Remove BuildSpec and v1 of BuildRequest.  Migrate some tests out of their test functions and into the BuildUserParams and BuildRequestV2 test functions.


rcerven:
looks great,
test_arrangements.py and test_api.py are still trying to import spec from BuildSpec

osbs/api.py
```diff
@@ -1240,21 +1231,25 @@ def get_compression_extension(self):
...
+        build_request = BuildRequestV2(build_json_store=self.os_conf.get_build_json_store())
+        build_request.set_params(git_uri="/", git_ref="HEAD", git_branch="master", user="user",
+                                 build_image=self.build_conf.get_build_image(),
+                                 build_imagestream=self.build_conf.get_build_imagestream(),
+                                 build_from=self.build_conf.get_build_from(),
+                                 base_image="base_image", name_label='name', build_type='worker',
+                                 registry_uris=self.build_conf.get_registry_uris())
+
+        plugins_config = PluginsConfiguration(build_request.user_params)
+        build_json = json.loads(plugins_config.render())
```
chmeliik: I'm not a huge fan of getting the JSON from `plugins_config` and then converting it back to a dict immediately, though I wasn't able to find a better way. Technically, you could do this
```python
plugins_config.render()
build_json = plugins_config.pt.template
```
but that relies entirely on the side effects of `plugins_config.render()` and I think I like that even less then the ~~Hobbit:~~ JSON: There and Back Again. Probably better to keep it as it is for now, up to you.

tests/test_api.py
```diff
@@ -2313,9 +2304,12 @@ def test_reactor_config_secret(self):
...
 
-        req = osbs_obj.create_build(target=TEST_TARGET,
-                                    architecture=TEST_ARCH,
-                                    **REQUIRED_BUILD_ARGS)
+        kwargs = copy.deepcopy(REQUIRED_BUILD_ARGS)
+        kwargs.update({
+            'build_type': 'orchestrator',
+            'reactor_config_override': {'required_secrets': ['mysecret']},
+        })
+        req = osbs_obj.create_build(target=TEST_TARGET, architecture=TEST_ARCH, **kwargs)
```
chmeliik: Is it necessary to copy `REQUIRED_BUILD_ARGS` here? Maybe `build_type` and `reactor_config_override` could just be passed to `osbs_obj.create_build` directly.

osbs/api.py
mlangsdorf: converting dictionaries back and forth is the least sketchy element of this particular function.  I'll leave it as is.

tests/test_api.py
mlangsdorf: Done, thanks!

rcerven:
we should also remove manipulate.py
as that was used only in BuildRequestv1
@twaugh  @MartinBasti @athos-ribeiro

mlangsdorf:
I totally forgot about manipulate.py; it's been removed.

osbs/api.py
mlangsdorf: actually, I'll just kill the function entirely.  Even better solution!

mlangsdorf:
The coveralls coverage decrease is an artifact of removing three tests suites that had high coverage percentages.

mlangsdorf:
Added to Release Notes under API Changes
"-  support for v1 BuildRequest and BuildRequest removed as part of removing  `arrangement_version` <= 5"

PR #863, opened 2019-06-17
inputs: delete old arrangements

mlangsdorf:
part of osbs-6573

Now that all the code that uses arrangements 1-5 is gone, delete
the JSON files that defined them.


MartinBasti:
IMHO `prod_inner.json` should be unused and removed too.

MartinBasti:
Hmm, but it is actually used as `DEFAULT_INNER_TEMPLATE` but only in tests.

mlangsdorf:
Tests updated, prod*json removed.

mlangsdorf:
Not adding anything to release notes because this is housekeeping for a feature already added to release notes.

PR #865, opened 2019-06-19
CVE-2019-10135: insecure yaml load of user input

MartinBasti:
`yaml.safe_load` must be used for user input to prevent injecting
malicious objects.


MartinBasti:
CC @cverna

PR #868, opened 2019-07-12
build request: set a completion deadline

mlangsdorf:
For osbs #7519

Set a completionDeadline of 3 hours in a worker or 4 hours in the orchestrator.  This should automatically cancel stuck builds.  Alternate values can be set in the config file and the deadline can be disabled altogether by using negative values.

### DIFF
--- a/osbs/build/build_requestv2.py
+++ b/osbs/build/build_requestv2.py
@@ -101,6 +101,8 @@ class BuildRequestV2(object):
         :param operator_manifests_extract_platform: str, indicates which platform should upload
                                                     operator manifests to koji
         :param parent_images_digests: dict, mapping image digests to names and platforms
+        :param worker_deadline: int, worker completion deadline in hours
+        :param orchestrator_deadline: int, orchestrator deadline in hours
         """
 
         # Here we cater to the koji "scratch" build type, this will disable
@@ -316,6 +318,8 @@ class BuildRequestV2(object):
         self.adjust_for_isolated(self.user_params.release)
         self.render_node_selectors(self.user_params.build_type.value)
 
+        self.set_deadline(self.user_params.build_type.value)
+
         # Set our environment variables
         custom_strategy['env'].append({
             'name': 'USER_PARAMS',
@@ -326,6 +330,7 @@ class BuildRequestV2(object):
             if env['name'] == 'ATOMIC_REACTOR_PLUGINS':
                 del custom_strategy['env'][index]
                 break
+
         # Log build json
         # Return build json
         self.build_json = self.template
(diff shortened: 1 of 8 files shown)


osbs/build/build_requestv2.py
```diff
@@ -444,6 +447,18 @@ def render_resource_limits(self):
             resources['limits'] = limits
             self.template['spec']['resources'] = resources
 
+    def set_deadline(self, build_type):
+        deadline_hours = 16
```
MartinBasti: is possible to make this configurable from a config instead of hardcoding it?

mlangsdorf: Yes, done.

osbs/build/user_params.py
```diff
@@ -280,6 +284,15 @@ def set_params(self,
         self.signing_intent.value = signing_intent
         self.compose_ids.value = compose_ids or []
 
+        try:
+            self.orchestrator_deadline.value = string.atoi(orchestrator_deadline)
```
athos-ribeiro: `string.atoi` is not available in python3 (and is deprecated in python2). You may want to use the `int()` builtin here. https://docs.python.org/2/library/string.html#string.atoi

osbs/build/user_params.py
```diff
@@ -280,6 +284,15 @@ def set_params(self,
...
+        except (ValueError, TypeError):
+            self.orchestrator_deadline.value = 4
```
athos-ribeiro: It would be nice to have these fallback values in the constants module

osbs/build/user_params.py
```diff
@@ -280,6 +284,15 @@ def set_params(self,
...
+        try:
+            self.worker_deadline.value = string.atoi(worker_deadline)
```
athos-ribeiro: See comment above on `atoi`

osbs/build/user_params.py
```diff
@@ -280,6 +284,15 @@ def set_params(self,
...
+        except (ValueError, TypeError):
+            self.worker_deadline.value = 3
```
athos-ribeiro: see comment above on having this in the constants module

osbs/build/build_requestv2.py
```diff
@@ -444,6 +449,17 @@ def render_resource_limits(self):
             resources['limits'] = limits
             self.template['spec']['resources'] = resources
 
+    def set_deadline(self, build_type):
+        if build_type == BUILD_TYPE_WORKER:
+            deadline_hours = self.user_params.worker_deadline.value
+        else:
+            deadline_hours = self.user_params.orchestrator_deadline.value
+
+        deadline_seconds = deadline_hours * 3600
```
twaugh: It would be great if we could disable the deadline entirely using config. For example, if `deadline_hours == 0` here.

osbs/build/user_params.py
mlangsdorf: Fixed.

osbs/build/user_params.py
mlangsdorf: Fixed.

osbs/build/build_requestv2.py

mlangsdorf:
Added to release notes under improvements:
- builds will automatically cancel themselves if any worker takes more than 3 hours to complete or the entire task takes more than 4 hours to complete.  The run time values can be set with `worker_max_run_hours` and `orchestrator_max_run_hours` in the configuration file.

PR #869, opened 2019-08-21
Don't show hide files magic in Dockerfile inside image in /root/build…

rcerven:
…info/

* OSBS-7865


rcerven:
rn updated
- Don't show hide files magic in Dockerfile inside image in /root/buildinfo/

PR #871, opened 2019-08-22
more stricter required labels checking,

rcerven:
name and component needs to have explicit value,
version, release may be using env variable

* OSBS-7697

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -513,6 +513,9 @@ class OSBS(object):
 
         required_missing = False
         req_labels = {}
+        # required labels which needs to have explicit value (not from env variable)
+        explicit_labels = [utils.Labels.LABEL_TYPE_NAME,
+                           utils.Labels.LABEL_TYPE_COMPONENT]
         # version label isn't used here, but is required label in Dockerfile
         # and is used and required for atomic reactor
         # if we don't catch error here, it will fail in atomic reactor later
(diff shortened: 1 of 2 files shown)


osbs/api.py
```diff
@@ -513,6 +513,9 @@ def _check_labels(self, repo_info):
 
         required_missing = False
         req_labels = {}
+        # required labels which needs to have explicit value (not from env variable)
+        strict_labels = [utils.Labels.LABEL_TYPE_NAME,
```
athos-ribeiro: could we call them explicit_labels?

osbs/api.py
```diff
@@ -521,14 +524,19 @@ def _check_labels(self, repo_info):
                       utils.Labels.LABEL_TYPE_VERSION]:
             try:
                 _, req_labels[label] = labels.get_name_and_value(label)
+
+                if label in strict_labels and not req_labels[label]:
```
athos-ribeiro: Could this block be in the try/except `else` clause?

osbs/api.py
```diff
@@ -521,14 +524,19 @@ def _check_labels(self, repo_info):
...
+                                 labels.get_name(label))
             except KeyError:
                 required_missing = True
                 logger.error("required label missing from Dockerfile : %s",
                              labels.get_name(label))
 
         try:
             _, release_value = labels.get_name_and_value(utils.Labels.LABEL_TYPE_RELEASE)
-            if not RELEASE_LABEL_FORMAT.match(release_value):
+            if release_value and not RELEASE_LABEL_FORMAT.match(release_value):
```
athos-ribeiro: Could this block also be moved to the `else` clause?

athos-ribeiro:
Looks good!

Could we rephrase the commit message to something like "Enforce explicit labels" ?

osbs/api.py

osbs/api.py
rcerven: I'd rather leave it here as it is more visible than in else after except

osbs/api.py
rcerven: I'd rather leave it here as it is more visible than in else after except

rcerven:
commit message changed

rcerven:
rn updated: - Enforce explicit required labels, name and component needs to have explicit value, version, release may be using env variable

PR #872, opened 2019-08-26
Use repr to log JSON response content

athos-ribeiro:
* OSBS-6632

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -628,7 +628,7 @@ class Openshift(object):
 
             encoding = None
             for line in response.iter_lines():
-                logger.debug(line)
+                logger.debug('%r', line)
 
                 if not encoding:
                     encoding = guess_json_utf(line)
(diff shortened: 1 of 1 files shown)


osbs/core.py
```diff
@@ -741,7 +741,7 @@ def adjust_attributes_on_object(self, collection, name, things, values, how):
         """
         url = self._build_url("%s/%s" % (collection, name))
         response = self._get(url)
-        logger.debug("before modification: %s", response.content)
+        logger.debug("before modification: %s", response.text)
```
twaugh: There is:
* the actual encoding of the response data
* the *declared* encoding of the response data in the headers
* the output encoding for logger.debug

This looks like it will fix a mismatch between the actual response encoding and the output encoding, but it won't help if the declared encoding is wrong -- whereas the `content` property will.

Don't we want `"%r", response.content` instead? That ought to cover all encoding mismatches.

athos-ribeiro: Sounds good. I will run a few tests in osbs-box and then squash the revert and the new commit.

athos-ribeiro: Done

athos-ribeiro:
Release notes added:

- Fix errors triggered when logging information from Dockerfiles containing non-ascii characters

PR #873, opened 2019-08-28
Use repr when printing logs for cli builds

athos-ribeiro:
This will prevent encoding errors when running osbs-client CLI

* OSBS-6632


athos-ribeiro:
Entry for this fix already present in RN

PR #878, opened 2019-09-25
Use OCP Groupified APIs

MartinBasti:
OCP4 doesn't support `oapi/` endpoint anymore and also `oapi/` is
deprecated in OCP3

* OSBS-5497


osbs/core.py
```diff
@@ -134,11 +140,12 @@ def _build_k8s_url(self, url, _prepend_namespace=True, **query):
             url += ("?" + urlencode(query))
         return urljoin(self.k8s_api_url, url)
 
-    def _build_url(self, url, _prepend_namespace=True, **query):
+    def _build_url(self, api_version, url, _prepend_namespace=True, **query):
```
twaugh: Should it infer this by looking up the first part of `url` in `OCP_RESOURCE_API_VERSION_MAP`?

twaugh: Oh, I see, /watch/... breaks that.

osbs/core.py
```diff
@@ -45,6 +45,12 @@
 WAIT_RETRY_HOURS = 12
 WAIT_RETRY = WAIT_RETRY_HOURS * 3600 // (WATCH_RETRY_SECS * WATCH_RETRY)
 
+OCP_RESOURCE_API_VERSION_MAP = {
+    'builds': 'build.openshift.io/v1',
+    'buildconfigs': 'build.openshift.io/v1',
+    'imagestreams': 'image.openshift.io/v1',
```
twaugh: Maybe we should give symbolic names to these strings?
```
OCP_RESOURCE_API_VERSION_BUILD = 'build.openshift.io/v1'
OCP_RESOURCE_API_VERSION_IMAGE = 'image.openshift.io/v1'
```
to prevent typos and to put the actual versions we're using in a central location.

MartinBasti:
From following snippet:
```
        self.os = Openshift(openshift_api_url=self.os_conf.get_openshift_api_uri(),
                            openshift_api_version=self.os_conf.get_openshift_api_version(),
                            openshift_oauth_url=self.os_conf.get_openshift_oauth_api_uri(),
                            k8s_api_url=self.os_conf.get_k8s_api_uri(),
                            ....

```
`openshift_api_version` is used only for k8s endpoint. Shall I rename it and also rename `get_openshift_api_version()` or do we have to keep it as public API and don't change interface?

osbs/core.py
MartinBasti: I ahve it in core.py, shall I move it to constants.py? (It's not used elsewhere)

MartinBasti:
arg `openshift_api_version` was removed as unused

MartinBasti:
Conflicts resolved

MartinBasti:
RN:
```
## Minimum required version of Openshift/OKD

- 3.6
```

```
## Improvements
- Support for OCP groupified APIs has been added, usage of `oapi/` endpoint was deprecated
```

PR #879, opened 2019-09-26
pass is_autorebuild to update_build_config_with_trigger to change label

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -440,10 +440,11 @@ class OSBS(object):
         return existing_bc
 
     @retry_on_conflict
-    def _update_build_config_with_triggers(self, build_json, triggers):
+    def _update_build_config_with_triggers(self, build_json, triggers, is_autorebuild=False):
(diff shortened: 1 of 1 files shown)


osbs/api.py
```diff
@@ -440,10 +440,11 @@ def _update_build_config_when_exist(self, build_json):
         return existing_bc
 
     @retry_on_conflict
-    def _update_build_config_with_triggers(self, build_json, triggers):
+    def _update_build_config_with_triggers(self, build_json, triggers, is_autorebuild="false"):
         existing_bc = self._get_existing_build_config(build_json)
         existing_bc['spec']['triggers'] = triggers
         build_config_name = existing_bc['metadata']['name']
+        existing_bc['metadata']['labels']['is_autorebuild'] = is_autorebuild
```
MartinBasti: Could we stay with boolean and only transform it here to str?

`... = "true" if is_autorebuild else "false" `

rcerven: changed

PR #882, opened 2019-10-16
new triggered_after_koji_task parameter

rcerven:

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -505,6 +505,9 @@ class OSBS(object):
                     imstreamtag['image']['dockerImageReference']
                 is_autorebuild = True
 
+            if build_request.triggered_after_koji_task is not None:
+                is_autorebuild = True
+
             existing_bc = self._update_build_config_with_triggers(build_json, triggers,
                                                                   is_autorebuild)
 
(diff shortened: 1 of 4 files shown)


MartinBasti:
Please add ticket number and more details about this feature into commit message. Otherwise LGTM

osbs/build/build_requestv2.py
```diff
@@ -105,6 +106,7 @@ def set_params(self, **kwargs):
         :param worker_deadline: int, worker completion deadline in hours
         :param orchestrator_deadline: int, orchestrator deadline in hours
         :param skip_build: bool, if we should skip build and just set buildconfig for autorebuilds
+        :param triggered_after_koji_task: str, koji task ID from which was autorebuild triggered
```
twaugh: Wouldn't it be an integer?

rcerven: will change

rcerven:
updated commit message

osbs/api.py
```diff
@@ -505,6 +505,9 @@ def _create_build_config_and_build(self, build_request):
                     imstreamtag['image']['dockerImageReference']
                 is_autorebuild = True
 
+            if build_request.triggered_after_koji_task is not None:
+                is_autorebuild = True
```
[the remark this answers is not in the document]
rcerven: we are passing triggered_after_koji_task, so we can use it later in koji build metadata, that way we will know from koji build which is the original task we are rebuilding

rcerven:
rn updated: - New triggered_after_koji_task parameter, when provided set is_autorebuild in build config

PR #887, opened 2019-10-29
utils: better error message when we fail to parse registry URI

ktdreyer:
Prior to this change, when a user specified a registry URI that completely fails to match our validation regex, `match()` returned None, and we called `.groups()` on that `None` value. This led to the following error message:

```
osbs.exceptions.OsbsException: AttributeError("'NoneType' object has no attribute 'groups'")
```

Handle the case where our regex does not match the user input and raise
a more helpful error message to the user.

- [ ] "Signed-off-by:" line is present in each commit

### DIFF
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -61,7 +61,10 @@ class RegistryURI(object):
     versionre = re.compile(r'((https?://)?([^/]*))(/(v\d+)?)?$')
 
     def __init__(self, uri):
-        groups = self.versionre.match(uri).groups()
+        match = self.versionre.match(uri)
+        if not match:
+            raise ValueError('Invalid registry URI {}'.format(uri))
+        groups = match.groups()
         self.docker_uri = groups[2]
         self.version = groups[4] or 'v2'
         self.scheme = groups[1] or ''


ktdreyer:
I'm setting up an OSBS instance following the [developer docs](https://github.com/containerbuildsystem/osbs-client/blob/master/docs/development-setup.md). I'm trying to figure out what to set for `registry_uri` because I didn't see any obvious examples on the internet. I blindly used this:

```
registry_uri = quay.io/ktdreyer/osbs
```

And when I ran `osbs ... build ...` the program crashes with the `NoneType' object has no attribute 'groups'` error. It took a few minutes of studying the backtrace to figure out that my `registry_uri` setting was the problem.

MartinBasti:
IIRC `registry_uri` and `source_registry_uri` are deprecated. It's now being a part of atomic-reactor config map. I cannot find usage of `registry_uri` in atomic reactor (passed from osbs-client in user build params)

ktdreyer:
Thanks Martin. What needs to change in osbs-client, in that case?

MartinBasti:
I'd keep it as is, in future we may do breaking change and remove those options completely

osbs/utils.py
```diff
@@ -61,7 +61,10 @@ class RegistryURI(object):
     versionre = re.compile(r'((https?://)?([^/]*))(/(v\d+)?)?$')
 
     def __init__(self, uri):
-        groups = self.versionre.match(uri).groups()
+        match = self.versionre.match(uri)
+        if not match:
+            raise OsbsValidationException('Invalid registry URI {}'.format(uri))
```
twaugh: As Martin suggested, I'd prefer `ValueError` here.

PR #888, opened 2019-10-29
user_params: better error for missing ":" in build-from

ktdreyer:
Prior to this change, when a user specified a `--build-from` option that does not contain a "`:`" character, `split()` returned one value, but we tried to assign the result to two values (`source_type` and `source_value`). This led to the following error:

```
  osbs.exceptions.OsbsException: ValueError('not enough values to unpack (expected 2, got 1)')
```

Handle the case where the user fails to specify "`:`" by returning a more helpful error message to the user.

### DIFF
--- a/osbs/build/user_params.py
+++ b/osbs/build/user_params.py
@@ -244,6 +244,9 @@ class BuildUserParams(BuildCommon):
                            "use build_from instead")
 
         if build_from:
+            if ':' not in build_from:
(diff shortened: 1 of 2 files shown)


MartinBasti:
Could you please add test for this change?

ktdreyer:
Sure, I've added a test to assert the error message string.

PR #909, opened 2019-11-22
add handling for setting a release version environmental variable

mlangsdorf:
Fixes osbs-7244

Add a new user param, release_env_var.  If it has a value, add it it to the Dockerfile after the FROM lines as an ENV with a value of the current release.

### DIFF
--- a/osbs/build/build_requestv2.py
+++ b/osbs/build/build_requestv2.py
@@ -377,6 +377,7 @@ class BuildRequestV2(BaseBuildRequest):
         :param operator_manifests_extract_platform: str, indicates which platform should upload
                                                     operator manifests to koji
         :param parent_images_digests: dict, mapping image digests to names and platforms
+        :param release_env_var: str, optional name of an ENV variable to hold release data
         :param worker_deadline: int, worker completion deadline in hours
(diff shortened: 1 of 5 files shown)


osbs/build/build_requestv2.py
```diff
@@ -571,11 +572,16 @@ def _set_deadline(self):
...
         self.set_deadline(deadline_hours)
 
-    def adjust_for_repo_info(self):
+    def adjust_for_repo_info(self, release_env_var, release):
         if not self._repo_info:
             logger.warning('repo info not set')
             return
 
+        if release_env_var.value:
+            release_var = "ENV {}={}".format(release_env_var.value, release.value)
```
athos-ribeiro: Could the release value be None at this point? If so, should we warn about setting the env var with an empty value?

twaugh:
This adds a configuration field to osbs.conf, but this file is controlled by site admins. Really we want the container.yaml file in the git repository to have this configuration field.

It also updates the Dockerfile in the git repository as checked out on the Koji builder, but really we want the atomic-reactor checkout to be updated.

Should this be a PR against atomic-reactor and not osbs-client?

mlangsdorf:
Redoing as an atomic-reactor change.

PR #911, opened 2019-11-27
imagestreamtag race

rcerven:

### DIFF
--- a/osbs/core.py
+++ b/osbs/core.py
@@ -23,7 +23,8 @@ from osbs.constants import (DEFAULT_NAMESPACE, BUILD_FINISHED_STATES, BUILD_RUNN
 from osbs.exceptions import (OsbsResponseException, OsbsException,
                              OsbsWatchBuildNotFound, OsbsAuthException,
                              ImportImageFailed, ImportImageFailedServerError)
-from osbs.utils import graceful_chain_get, retry_on_conflict, retry_on_exception
+from osbs.utils import (graceful_chain_get, retry_on_conflict, retry_on_exception,
+                        retry_on_not_found)
 
 import requests
 from requests.utils import guess_json_utf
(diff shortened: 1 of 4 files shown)


osbs/core.py
```diff
@@ -841,6 +842,16 @@ def get_image_stream_tag(self, tag_id):
         check_response(response, log_level=logging.DEBUG)
         return response
 
+    @retry_on_not_found
+    def get_image_stream_tag_with_retry(self, tag_id):
```
twaugh: We don't have any test coverage for this -- is this ok?

osbs/utils.py
```diff
@@ -702,6 +702,19 @@ def should_retry_cb(ex):
     return retry
 
 
+def retry_on_not_found(func):
+    @wraps(func)
+    def retry(*args, **kwargs):
+        # Only retry when OsbsResponseException was raised due to not found
+        def should_retry_cb(ex):
+            return ex.status_code == http_client.NOT_FOUND
```
twaugh: We don't have coverage here according to coveralls, but we do on the equivalent part of `retry_on_conflict`.

osbs/core.py
rcerven: that should be covered by https://github.com/containerbuildsystem/osbs-client/pull/911/files#diff-625efbbf35a2290861f596fee1b19838R2172

not sure why it is showing that it isn't

osbs/utils.py
rcerven: I will try to add it , I think I have idea

rcerven: idea failed

osbs/core.py
rcerven: nvm that test is testing method in api, and we are always mocking methods in core

rcerven:
rn updated: - prevent tag race condition when updating imagestreamtag for autorebuilds in multiple builds at the same time

PR #934, opened 2020-02-10
Adjust the pre-build Flatpak plugins

owtaylor:
atomic-reactor is being changed to use the 'resolve_composes' plugin
for Flatpaks, instead of having a separate 'resolve_module_compose',
and to add a new 'flatpak_update_dockerfile' plugin that runs after
'resolve_composes'.
    
The 'resolve_composes' plugin is moved up a bit in the sequence of
orchestrator plugins so that 'flatpak_update_dockerfile' can set the
release label before bump_release runs. As far as I can tell, it should
work as well in the new location as the old location.

See https://github.com/containerbuildsystem/atomic-reactor/pull/1373

### DIFF
--- a/inputs/orchestrator_inner:6.json
+++ b/inputs/orchestrator_inner:6.json
@@ -18,9 +18,6 @@
       "required": false
     },
     {
-      "name": "resolve_module_compose"
-    },
-    {
       "name": "flatpak_create_dockerfile"
     },
     {
@@ -41,6 +38,9 @@
     },
     {
       "name": "resolve_composes"
+    },
+    {
+      "name": "flatpak_update_dockerfile"
     },
     {
       "name": "bump_release"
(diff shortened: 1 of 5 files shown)


MartinBasti:
Please resolve conflicts

rcerven:
bump_release should run before add_labels_in_dockerfile

also because it is now reserving builds in koji, it should also run before resolve_compose which might take a while

owtaylor:
> bump_release should run before add_labels_in_dockerfile

That's the case with this patch set - unless I'm missing something I didn't change that.

> also because it is now reserving builds in koji, it should also run before resolve_compose which might take a while

The dependency chain is that flatpak_update_dockerfile now sets the release tag based on the results of resolve_compose. So, that's why the patch switches things to do bump_release after resolve_compose. Can you explain the rationale about timing a bit more? Does Koji time-out reservations?

MartinBasti:
Release notes:
> - Flatpak plugins we rearranged to better support inheriting parent composes

PR #937, opened 2020-02-18
api: report incomplete positional arguments to _do_create_build

mlangsdorf:
Fixes OSBS-8026

Catch when there are an insufficient number of positional arguments to _do_create_prod_build and report which one is missing.

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -89,6 +89,8 @@ def osbsapi(func):
     return catch_exceptions
 
 
+_REQUIRED_PARAM = object()
+
 logger = logging.getLogger(__name__)
 
 LogEntry = namedtuple('LogEntry', ['platform', 'line'])
@@ -678,7 +680,8 @@ class OSBS(object):
         return user_params
 
     def _do_create_prod_build(self,
-                              git_uri, git_ref, git_branch,
+                              git_uri=_REQUIRED_PARAM, git_ref=_REQUIRED_PARAM,
+                              git_branch=_REQUIRED_PARAM,
                               inner_template=None,
                               outer_template=None,
                               customize_conf=None,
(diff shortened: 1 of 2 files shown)


MartinBasti:
`_do_create_prod_build` is used everywhere only with `**kwargs`. IMHO it would be better to update `_do_create_prod_build`:

```
_REQUIRED_PARAM = object()
```

```diff
diff --git a/osbs/api.py b/osbs/api.py
index 72bf147..ccb90d0 100644
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -654,9 +654,14 @@ class OSBS(object):
             utils.Labels.LABEL_TYPE_COMPONENT: module.name
         }, None
 
-    def _do_create_prod_build(self, git_uri, git_ref,
-                              git_branch,
-                              user,
+
+
+
+    def _do_create_prod_build(self,
+                              git_uri=_REQUIRED_PARAM,
+                              git_ref=_REQUIRED_PARAM,
+                              git_branch=_REQUIRED_PARAM,
+                              user=_REQUIRED_PARAM,
                               component=None,
                               target=None,
                               architecture=None, yum_repourls=None,
```
And check it in the body of function if all reuired params are set` (would be super easy with py3 only:  `def _do_create_prod_build(self, *, git_uri, ...`)

mlangsdorf:
Nice!  Thanks, much cleaner.

osbs/api.py
```diff
@@ -691,6 +694,13 @@ def _do_create_prod_build(self,
                               target=None,
                               **kwargs):
 
+        if git_uri == _REQUIRED_PARAM:
```
MartinBasti: Could be better to use `is` as we care exactly about the same isntance.

MartinBasti: the same as we use for `is None`

osbs/api.py
```diff
@@ -691,6 +694,13 @@ def _do_create_prod_build(self,
...
+            raise OsbsException('required parameter git_uri missing')
+        if git_ref == _REQUIRED_PARAM:
+            raise OsbsException('required parameter git_ref missing')
+        if git_branch == _REQUIRED_PARAM:
+            raise OsbsException('required parameter git_branch missing')
+
```
rcerven: would be better to first check all params which are missing and then raise,
because if all 3 are missing users will always get just the 1st missing

mlangsdorf:
Added to RN under "API changes":
- Failing to provide the git URI, git ref, or git branch will raise an obvious error specifying which arguments are missing, instead of a confusing error specifying that some arguments are missing.

PR #938, opened 2020-02-19
repo_utils: validate container.yaml against the schema

mlangsdorf:
Fixes OSBS-5734

Move the validation of the container.yaml out of atomic-reactor and into osbs-client.

### DIFF
--- a/osbs/repo_utils.py
+++ b/osbs/repo_utils.py
@@ -7,14 +7,18 @@ of the BSD license. See the LICENSE file for details.
 """
 
 
-from __future__ import absolute_import
+from __future__ import print_function, absolute_import, unicode_literals
 
 from osbs.exceptions import OsbsException
 from osbs.constants import REPO_CONFIG_FILE, ADDITIONAL_TAGS_FILE, REPO_CONTAINER_CONFIG
 from six import StringIO
 from six.moves.configparser import ConfigParser
+from pkg_resources import resource_stream
 from textwrap import dedent
 
+import codecs
+import json
+import jsonschema
 import logging
 import os
 import re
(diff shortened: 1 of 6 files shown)


osbs/repo_utils.py
```diff
@@ -82,13 +132,12 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE, depth=None,
...
-            with open(file_path) as f:
-                try:
-                    self.container = yaml.safe_load(f) or {}
-                except yaml.scanner.ScannerError as e:
-                    msg = ('Failed to parse YAML file "{file}": {reason}'
-                           .format(file=REPO_CONTAINER_CONFIG, reason=e))
-                    raise OsbsException(msg)
+            try:
+                self.container = read_yaml_from_file_path(file_path, 'schemas/container.json') or {}
+            except yaml.scanner.ScannerError as e:
```
rcerven: except Exception as e:

msg => Failed to load and validate ....

mlangsdorf: Why go from the specific error type to the general?  The linter seems to prefer specific errors.

Regardless, I improved the error message.

rcerven: read_yaml will return multiple different exceptions, so we want to wrap those in here with OsbsException

mlangsdorf: Okay then, fixed.

athos-ribeiro:
LGTM! Thanks, Mark.

It seems we will lose some test coverage (see [1] vs [2]) when moving the functions to read the yaml schemas here. Are these cases we do care about testing?

[1] https://coveralls.io/builds/29073151/source?filename=atomic_reactor/util.py
[2] https://coveralls.io/builds/28978947/source?filename=osbs%2Frepo_utils.py#L44

mlangsdorf:
We weren't really testing all those variations in the utils test, though, since the actual test only checked for success:
```
@pytest.mark.parametrize('from_file', [True, False])
@pytest.mark.parametrize('config', [
    ("""\
      version: 1
      clusters:
        ignored:
        - name: foo
          max_concurrent_builds: 2
        platform:
        - name: one
          max_concurrent_builds: 4
        - name: two
          max_concurrent_builds: 8
          enabled: true
        - name: three
          max_concurrent_builds: 16
          enabled: false
    """),
    REACTOR_CONFIG_MAP,
])
def test_read_yaml_file_or_yaml(tmpdir, from_file, config):
    expected = yaml.safe_load(config)

    if from_file:
        config_path = os.path.join(str(tmpdir), 'config.yaml')
        with open(config_path, 'w') as fp:
            fp.write(config)
        output = read_yaml_from_file_path(config_path, 'schemas/config.json')
    else:
        output = read_yaml(config, 'schemas/config.json')

    assert output == expected
```

There's another test in test_reactor_config that does check those variations, and it still exists.  I can try to port it over to osbs-client if you want better locality for tests.

mlangsdorf:
Ported over the test and another another test.  I can't figure out how to raise the jsonschema error properly, so that one will still be without a test but at least it's not reducing coverage anymore.

PR #942, opened 2020-03-03
Change tests and travis for correct working with Fedora 31

SalatskySal:

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -2,14 +2,16 @@
 set -eux
 
 # Prepare env vars
+ENGINE=${ENGINE:="podman"}
 OS=${OS:="centos"}
 OS_VERSION=${OS_VERSION:="7"}
 PYTHON_VERSION=${PYTHON_VERSION:="2"}
 ACTION=${ACTION:="test"}
 IMAGE="$OS:$OS_VERSION"
-docker_mounts="-v $PWD:$PWD:z"
+# Use arrays to prevent globbing and word splitting 
+engine_mounts=(-v "$PWD":"$PWD":z)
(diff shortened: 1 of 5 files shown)


test.sh
```diff
@@ -2,14 +2,15 @@
...
 OS=${OS:="centos"}
 OS_VERSION=${OS_VERSION:="7"}
 PYTHON_VERSION=${PYTHON_VERSION:="2"}
 ACTION=${ACTION:="test"}
 IMAGE="$OS:$OS_VERSION"
-docker_mounts="-v $PWD:$PWD:z"
+engine_mounts="$PWD:$PWD:z"
 for dir in ${EXTRA_MOUNT:-}; do
-  docker_mounts="${docker_mounts} -v $dir:$dir:z"
+  engine_mounts="${engine_mounts} -v $dir:$dir:z"
```
chmeliik: If `"$engine_mounts"` is expected to be used with quotes, then this will not work. I don't think I ever had any use for `$EXTRA_MOUNT`, maybe we can just drop it? Or, if we need it, then `$engine_mounts` will have to be used without quotes (maybe just disable the shellcheck warning, or see https://github.com/koalaman/shellcheck/wiki/SC2086 for other ways to fix)

athos-ribeiro: I'd be OK with either adding the quotes or adding an extra commit to remove the EXTRA_MOUNT from this script. This shell script is already too bloated. It would be nice to check with others if they ever use this option if we are to remove it.

FYI, it was introduce in `92e7ba9` to "make it easier to mount more volumes in the test container".

chmeliik: I'd vote to remove EXTRA_MOUNT

PR #945, opened 2020-03-05
Improve "cannot find commit" error message

SalatskySal:
*OSBS-8552

Added new exception for better understanding,
changed error message by adding information about branch,
changed tests for correct working.

Signed-off-by: Salatskyi Serhii <ssalatsk@redhat.com>

### DIFF
--- a/osbs/exceptions.py
+++ b/osbs/exceptions.py
@@ -88,6 +88,10 @@ class OsbsWatchBuildNotFound(OsbsException):
     """ watch stream ended and build was not found """
 
 
+class OsbsCommitNotFound(OsbsException):
+    """Commit was not found in repo"""
+
+
 class ImportImageFailed(OsbsException):
     """Import image via ImageStream failed"""
 
--- a/osbs/utils.py
+++ b/osbs/utils.py
@@ -46,7 +46,8 @@ except ImportError:
     from calendar import timegm
 
 from dockerfile_parse import DockerfileParser
-from osbs.exceptions import OsbsException, OsbsResponseException, OsbsValidationException
+from osbs.exceptions import (OsbsException, OsbsResponseException,
+                             OsbsValidationException, OsbsCommitNotFound)
 
 logger = logging.getLogger(__name__)
 ClonedRepoData = namedtuple('ClonedRepoData', ['repo_path', 'commit_id', 'commit_depth'])
@@ -256,7 +257,11 @@ def clone_git_repo(git_url, target_dir=None, commit=None, retry_times=GIT_MAX_RE
             # we are using check_output, even though we aren't using
             # the return value, but we will get 'output' in exception
             subprocess.check_output(cmd, stderr=subprocess.STDOUT)
-            repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
+            try:
+                repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
+            except OsbsCommitNotFound as exc:
+                raise OsbsCommitNotFound("Commit {} is not reachable in branch {}, reason: {}"
+                                         .format(commit, branch, exc))
             break
         except subprocess.CalledProcessError as exc:
             if counter != retry_times:
@@ -293,15 +298,16 @@ def reset_git_repo(target_dir, git_reference, retry_depth=None):
             break
         except subprocess.CalledProcessError:
             if not deepen:
-                raise OsbsException('cannot find commit %s in repo %s' %
-                                    (git_reference, target_dir))
+                raise OsbsCommitNotFound('cannot find commit {} in repo {}'.format(
+                                         git_reference, target_dir))
             deepen *= 2
             cmd = ["git", "fetch", "--depth", str(deepen)]
             subprocess.check_call(cmd, cwd=target_dir)
             logger.debug("Couldn't find commit %s, increasing depth with '%s'", git_reference,
                          cmd)
     else:
-        raise OsbsException('cannot find commit %s in repo %s' % (git_reference, target_dir))
+        raise OsbsCommitNotFound('cannot find commit {} in repo {}'.format(
+                                  git_reference, target_dir))
 
     cmd = ["git", "rev-parse", "HEAD"]
     logger.debug("getting SHA-1 of provided ref '%s'", git_reference)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -31,7 +31,7 @@ from osbs.utils import (buildconfig_update,
                         get_instance_token_file_name, Labels, sanitize_version,
                         has_triggers, strip_registry_and_tag_from_image,
                         clone_git_repo, get_repo_info)
-from osbs.exceptions import OsbsException
+from osbs.exceptions import OsbsException, OsbsCommitNotFound
 from tests.constants import (TEST_DOCKERFILE_GIT, TEST_DOCKERFILE_SHA1, TEST_DOCKERFILE_INIT_SHA1,
                              TEST_DOCKERFILE_BRANCH)
 import osbs.kerberos_ccache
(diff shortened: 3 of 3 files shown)


osbs/utils.py
```diff
@@ -293,15 +298,15 @@ def reset_git_repo(target_dir, git_reference, retry_depth=None):
...
+                raise OsbsException('Cannot find commit %s in repo %s' %
                                     (git_reference, target_dir))
             deepen *= 2
             cmd = ["git", "fetch", "--depth", str(deepen)]
             subprocess.check_call(cmd, cwd=target_dir)
             logger.debug("Couldn't find commit %s, increasing depth with '%s'", git_reference,
                          cmd)
     else:
-        raise OsbsException('cannot find commit %s in repo %s' % (git_reference, target_dir))
+        raise OsbsException('Cannot find commit %s in repo %s' % (git_reference, target_dir))
```
MartinBasti: I'd keep lowercase letter here

osbs/utils.py
```diff
@@ -293,15 +298,15 @@ def reset_git_repo(target_dir, git_reference, retry_depth=None):
             break
         except subprocess.CalledProcessError:
             if not deepen:
-                raise OsbsException('cannot find commit %s in repo %s' %
+                raise OsbsException('Cannot find commit %s in repo %s' %
```
MartinBasti: I'd keep lowercase letter here

osbs/exceptions.py
```diff
@@ -75,6 +75,10 @@ class OsbsAuthException(OsbsException):
     pass
 
 
+class OsbsCommitNotFound(OsbsException):
```
MartinBasti: I was thinking of having this defined only in module where it's used (not general OSBS exception)

tests/test_utils.py
```diff
@@ -579,10 +579,10 @@ def test_calc_depth_git_repo(tmpdir):
 @requires_internet
 def test_clone_git_repo_commit_failure(tmpdir, commit, branch, depth):
     tmpdir_path = str(tmpdir.realpath())
-    with pytest.raises(OsbsException) as exc:
+    with pytest.raises(OsbsCommitNotFound) as exc:
         clone_git_repo(TEST_DOCKERFILE_GIT, tmpdir_path, commit=commit, retry_times=1,
                        branch=branch, depth=depth)
-    assert 'cannot find commit' in exc.value.message
+    assert re.match(r'^Commit \S+ is not reachable in branch \S+, reason: (\S+\s?)*$', exc.value.message)
```
MartinBasti: instead of `message` please use `str(exc.value)`

osbs/exceptions.py
SalatskySal: @MartinBasti I need make a correct checking by tests this part of code, so I must share this in test_utils.py

SalatskySal: Oh, I suggested bad idea. Understood and resolved :)

tests/test_utils.py
```diff
@@ -582,7 +582,8 @@ def test_clone_git_repo_commit_failure(tmpdir, commit, branch, depth):
     with pytest.raises(OsbsException) as exc:
         clone_git_repo(TEST_DOCKERFILE_GIT, tmpdir_path, commit=commit, retry_times=1,
                        branch=branch, depth=depth)
-    assert 'cannot find commit' in exc.value.message
+    assert re.match(r'^Commit \S+ is not reachable in branch \S+, '
+                    r'reason: (\S+\s?)*$', str(exc.value))
```
chmeliik: Instead of using regex matching, could the error message just be formatted with the expected commit and branch?

osbs/utils.py
```diff
@@ -52,6 +52,10 @@
 ClonedRepoData = namedtuple('ClonedRepoData', ['repo_path', 'commit_id', 'commit_depth'])
 
 
+class OsbsCommitNotFound(OsbsException):
```
MartinBasti: Please add docstring

osbs/utils.py
```diff
@@ -256,7 +260,11 @@ def clone_git_repo(git_url, target_dir=None, commit=None, retry_times=GIT_MAX_RE
             # we are using check_output, even though we aren't using
             # the return value, but we will get 'output' in exception
             subprocess.check_output(cmd, stderr=subprocess.STDOUT)
-            repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
+            try:
+                repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
+            except OsbsException as exc:
+                raise OsbsCommitNotFound("Commit %s is not reachable in branch %s, reason: %s"
+                                         % (commit, branch, exc.message))
```
MartinBasti: Please don't use `exc.message`. It's deprecated. Use just plain exception object it will be converted to string when you use `%s` formatting.

Also I'd prefer `.format()` formatting instead of `%`. It's never and better for future compatibility with F-string in py36+

osbs/utils.py
```diff
@@ -256,7 +260,11 @@ def clone_git_repo(git_url, target_dir=None, commit=None, retry_times=GIT_MAX_RE
             # we are using check_output, even though we aren't using
             # the return value, but we will get 'output' in exception
             subprocess.check_output(cmd, stderr=subprocess.STDOUT)
-            repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
+            try:
+                repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
+            except OsbsException as exc:
```
MartinBasti: You made this in reversed way.

I wanted something like:
```
try:
    repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
except GitReferenceNotFound as e:
   raise OsbsException(....)
```

osbs/utils.py
```diff
@@ -256,7 +260,11 @@ def clone_git_repo(git_url, target_dir=None, commit=None, retry_times=GIT_MAX_RE
...
+                raise OsbsCommitNotFound("Commit %s is not reachable in branch %s, reason: %s"
```
MartinBasti: `s/is not reachable/not found/`?

osbs/utils.py
MartinBasti: Also it should't inherit from `OsbsException` as  this exception should never reach user. Use just `Exception`

osbs/utils.py
```diff
@@ -256,7 +260,11 @@ def clone_git_repo(git_url, target_dir=None, commit=None, retry_times=GIT_MAX_RE
             # we are using check_output, even though we aren't using
             # the return value, but we will get 'output' in exception
             subprocess.check_output(cmd, stderr=subprocess.STDOUT)
-            repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
+            try:
+                repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
+            except OsbsCommitNotFound as exc:
+                raise OsbsException("Commit {} is not reachable in branch {}, reason: {}".format(
+                                    commit, branch, str(exc)))
```
MartinBasti: no need for `str(exc)`. Format will retype it into string automatically

osbs/utils.py
```diff
@@ -256,7 +260,11 @@ def clone_git_repo(git_url, target_dir=None, commit=None, retry_times=GIT_MAX_RE
             # we are using check_output, even though we aren't using
             # the return value, but we will get 'output' in exception
             subprocess.check_output(cmd, stderr=subprocess.STDOUT)
-            repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
+            try:
+                repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
```
MartinBasti: I realized that this is used in atomic-reactor. So we have to rework exception to be subclass of `OsbsException`


IMHO we can re-raise the same exception type with bigger context.

```
try:
    ....
except OsbsCommitNotFound as exc:
    raise OsbsCommitNotFound("......".format(branch, exc))
```

MartinBasti:
Release notes:
> - When git reference doesn't exist error messages reports also which branch is checked out

PR #952, opened 2020-03-11
Make sure json schemas are installed by setup.py

chmeliik:
Signed-off-by: Adam Cmiel <acmiel@redhat.com>

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ setup(
           'console_scripts': ['osbs=osbs.cli.main:main'],
     },
     install_requires=_install_requirements(),
+    package_data={'osbs': ['schemas/*.json']},
     data_files=data_files.items(),
     setup_requires=[],
     tests_require=_get_requirements('tests/requirements.txt'),


MartinBasti:
Please rebase so we will if RPMs are build fine in copr

MartinBasti:
Shall we also update this section in specfile?
```
%ghost %config(noreplace) %{_datadir}/osbs/orchestrator_customize.json
%ghost %config(noreplace) %{_datadir}/osbs/prod_customize.json
%ghost %config(noreplace) %{_datadir}/osbs/worker_customize.json
```

chmeliik:
> Shall we also update this section in specfile?

@MartinBasti what exactly does that do? The "noreplace" part makes me think it marks that file as something that should not be replaced during a `dnf upgrade` - probably not what we want

athos-ribeiro:
That's right. The files are ghosted because they are not being shipped by the package, but if they are there, they are owned by the package. Configuration files should not be replaced by new releases: we do not want to override user configurations. That is not true for **most** OSBS json files though. with exception for the *_customize.json files there, we usually want to overwrite these JSON files with plugin order (or schemas) on upgrades.

athos-ribeiro:
Although, it would be nice to be explicit in the .spec file that we want to package everything under the schema dirs. Maybe adding sth like `%{datadir}/osbs/schemas/*.json` to the files section

chmeliik:
@athos-ribeiro the schema does not go in /usr/share/osbs though, it goes in /usr/lib/python*/site-packages/osbs*. That's already in the spec as `%{python3_sitelib}/osbs*`

PR #954, opened 2020-03-12
Use tito for releases

MartinBasti:
Make releases automated as much as possible

- [ ] "Signed-off-by:" line is present in each commit


MartinBasti:
@StykMartin :eyes:

osbs-client.spec
```diff
@@ -32,20 +29,17 @@
...
-%endif
+Version:        0.64.1
+Release:        1%{?dist}
 
 Summary:        Python command line client for OpenShift Build Service
 Group:          Development/Tools
 License:        BSD
 URL:            https://github.com/containerbuildsystem/osbs-client
-Source0:        https://github.com/containerbuildsystem/osbs-client/archive/%{commit}/osbs-client-%{commit}.tar.gz
+Source0:        %{name}-%{version}.tar.gz
```
StykMartin: ugh. This should point to an upstream resource for your tarball. 
The goal is to separate upstream source and vendor modifications.

setup.py
```diff
@@ -34,7 +34,7 @@ def _install_requirements():
 setup(
     name="osbs-client",
     description='Python module and command line client for OpenShift Build Service',
-    version="0.65.dev",
+    version="0.64.1",
```
StykMartin: So you keeping here version, which is already released in PyPI.
What will happen if someone will try to use dev and version is same as released one

MartinBasti: Hmm, while I'm thinking about this, probably we should just follow pattern from https://github.com/containerbuildsystem/koji-containerbuild and don't bump to dev version. It makes tito workflow easy. Dev version must be force installed.

osbs-client.spec
```diff
@@ -145,7 +127,7 @@ This package contains osbs Python 2 bindings.
 
 
 %prep
-%setup -qn %{name}-%{commit}
+%setup -q
```
athos-ribeiro: Could we use autosetup here? (nitpick)

https://rpm.org/user_doc/autosetup.html

setup.py
athos-ribeiro: Could we fetch this version from a single place, such as osbs/version.py?

osbs-client.spec
MartinBasti: I;d do it in separate PRs, I;d like to keep it in sync with other projects

setup.py
MartinBasti: We could, but it;s not nice and tito handles this for us

MartinBasti:
No release notes, change to release process only

PR #971, opened 2020-04-23
Plugins decide for themselves whether to run (osbs-client doesn't)

rcerven:
* CLOUDBLD-111

### DIFF
--- a/docs/build_process.md
+++ b/docs/build_process.md
@@ -29,9 +29,6 @@ This document mentions several components that communicate with each other:
    1. The `BuildConfig`'s source git ref names the git branch the build is from
       ― The `bump_release` configuration names the branch's commit the initial
       build is from
-   1. The `stop_autorebuild_if_disabled` configuration names the config file
-      where autorebuilds are enabled/disabled, see [OSBS Repo Configuration
-      File][]
 1. osbs-client starts the Openshift build from `BuildConfig` for *IM*
 1. Openshift spawns a new build container that contains atomic-reactor inside
 1. atomic-reactor's `CheckAndSetRebuildPlugin` is run. This determines whether
--- a/osbs/build/plugins_configuration.py
+++ b/osbs/build/plugins_configuration.py
@@ -174,68 +174,6 @@ class PluginsConfigurationBase(object):
         """
         raise NotImplementedError
 
-    def has_tag_suffixes_placeholder(self):
-        phase = 'postbuild_plugins'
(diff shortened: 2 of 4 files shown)


docs/build_process.md
```diff
@@ -29,9 +29,6 @@ This document mentions several components that communicate with each other:
    1. The `BuildConfig`'s source git ref names the git branch the build is from
       ― The `bump_release` configuration names the branch's commit the initial
       build is from
-   1. The `stop_autorebuild_if_disabled` configuration names the config file
```
athos-ribeiro, line 32: Should this be in a separate commit?

osbs/build/plugins_configuration.py
```diff
@@ -174,68 +174,6 @@ def render(self):
...
-    def has_tag_suffixes_placeholder(self):
-        phase = 'postbuild_plugins'
-        plugin = 'tag_from_config'
-        if not self.pt.has_plugin_conf(phase, plugin):
-            logger.debug('no tag suffix placeholder')
-            return False
-
-        placeholder = '{{TAG_SUFFIXES}}'
-        plugin_conf = self.pt.get_plugin_conf('postbuild_plugins', 'tag_from_config')
-        return plugin_conf.get('args', {}).get('tag_suffixes') == placeholder
```
athos-ribeiro, line 186: Why is this being removed without adding an atomic reactor counterpart? Should it be in a different commit?

rcerven: we don't need this, that placeholder was always in json inner template, and this wasn't providing any value, IF you don't want to plugin to run, you can just simply remove it from template

docs/build_process.md
rcerven: not sure if it is worth, just obsolete plugin in doc

osbs/build/plugins_configuration.py
athos-ribeiro: This seems to be unrelated to the commit that is moving code to atomic-reactor. Could it be in a separate commit? Alternatively, you could add a longer description to the commit message (e.g. bullets) explaining this specific change and the removal of `stop_autorebuild_disabled` saying both are a consequence of this change.

rcerven: updated commit msg

rcerven:
rn updated: Plugins decide for themselves whether to run (osbs-client doesn't

PR #981, opened 2020-05-21
create imagestreams with registry name

rcerven:
* CLOUDBLD-746

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -43,8 +43,7 @@ from osbs.constants import (BUILD_RUNNING_STATES, WORKER_OUTER_TEMPLATE,
                             ORCHESTRATOR_CUSTOMIZE_CONF, BUILD_TYPE_WORKER,
                             BUILD_TYPE_ORCHESTRATOR, BUILD_FINISHED_STATES,
                             DEFAULT_ARRANGEMENT_VERSION, REACTOR_CONFIG_ARRANGEMENT_VERSION,
-                            ANNOTATION_SOURCE_REPO, ANNOTATION_INSECURE_REPO, FILTER_KEY,
-                            RELEASE_LABEL_FORMAT,
+                            FILTER_KEY, RELEASE_LABEL_FORMAT,
                             ORCHESTRATOR_SOURCES_OUTER_TEMPLATE,
                             USER_PARAMS_KIND_IMAGE_BUILDS,
                             USER_PARAMS_KIND_SOURCE_CONTAINER_BUILDS,
@@ -55,8 +54,7 @@ from osbs.exceptions import (OsbsException, OsbsValidationException, OsbsRespons
 from osbs.utils.labels import Labels
 # import utils in this way, so that we can mock standalone functions with flexmock
 from osbs import utils
-from osbs.utils import (retry_on_conflict, graceful_chain_get, RegistryURI,
-                        strip_registry_and_tag_from_image)
+from osbs.utils import (retry_on_conflict, graceful_chain_get, RegistryURI, ImageName)
 
 from six.moves import http_client, input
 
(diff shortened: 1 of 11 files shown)


osbs/api.py
```diff
@@ -425,33 +423,68 @@ def _create_build_directly(self, build_request, unique=None):
...
         return BuildResponse(self.os.create_build(build_json).json(), self)
 
+    @retry_on_conflict
+    def _get_or_create_imagestream(self, imagestream_name, build_request):
+        insecure = False
+        source_registry_uri = RegistryURI(build_request.source_registry['url']).docker_uri
+        source_registry_insecure = build_request.source_registry.get('insecure', False)
+
+        docker_image_repo = ImageName.parse(build_request.base_image)
+        if build_request.organization:
```
lcarva, line 433: This must be done conditionally based on the _registry_ configuration. You mentioned this to me in IRC, just adding a comment here so I don't forget it :)

rcerven: already fixed :)

osbs/build/build_requestv2.py
```diff
@@ -402,10 +447,14 @@ def set_data_from_reactor_config(self, reactor_config_data):
             if self.user_params.flatpak.value and not self.user_params.base_image.value:
                 raise OsbsValidationException(
```
lcarva, line 448: Should we always raise `OsbsValidationException` if `reactor_config_data` is falsey?

osbs/build/build_requestv2.py
```diff
@@ -402,10 +447,14 @@ def set_data_from_reactor_config(self, reactor_config_data):
...
                     "Flatpak base_image must be be set in container.yaml or reactor config")
-            else:
-                return
+        else:
+            self._set_flatpak(reactor_config_data)
 
-        self._set_flatpak(reactor_config_data)
+        if not self.source_registry:
+            raise RuntimeError('mandatory "source_registry" is not defined in reactor_config')
+        source_registry = RegistryURI(self.source_registry['url']).docker_uri
```
lcarva, line 455: Is it safe to always assume the source registry here? Why not take the pull registries into consideration?

osbs/cli/main.py
```diff
@@ -135,7 +135,8 @@ def cmd_list_builds(args, osbs):
                             key=lambda x: x.get_time_created_in_seconds()):
             unique_image = build.get_image_tag()
             try:
-                image = strip_registry_from_image(build.get_repositories()["primary"][0])
+                image = \
+                    ImageName.parse(build.get_repositories()["primary"][0]).to_str(registry=False)
             except (TypeError, KeyError, IndexError):
```
lcarva, line 140: Are these exception still valid when `ImageName` is used?

osbs/api.py
```diff
@@ -425,33 +423,68 @@ def _create_build_directly(self, build_request, unique=None):
...
+                            insecure = pull_reg.get('insecure', False)
+                            allowed_registry = True
+                            break
+
+                if not allowed_registry:
+                    raise RuntimeError('Not allowed explicitly specified registry: {}'.
+                                       format(docker_image_repo.registry))
+
+        try:
+            imagestream = self.get_image_stream(imagestream_name)
```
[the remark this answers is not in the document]
rcerven: it is : imagestream_name contains registry
it gets it from build_request.trigger_imagestreamtag

osbs/build/build_requestv2.py
rcerven: well it will fail anyway few lines bellow if it doesn't fine source_registry

osbs/build/build_requestv2.py
rcerven: we need source registry to name imagestream and imagestream_tag for result of build

osbs/cli/main.py
rcerven: I guess since we are passing (build.get_repositories()["primary"][0])
seems to me like all could happen

osbs/build/build_requestv2.py
lcarva: Ah! Missed that this is setting up for the destination location. The source registry is the only possible destination registry.

osbs/api.py
```diff
@@ -425,33 +423,70 @@ def _create_build_directly(self, build_request, unique=None):
...
+        else:
+            if docker_image_repo.registry == source_registry_uri:
+                insecure = source_registry_insecure
+
+            else:
+                allowed_registry = False
+
+                if build_request.pull_registries:
+                    for pull_reg in build_request.pull_registries:
+                        if docker_image_repo.registry == pull_reg['url']:
```
chmeliik, line 446: I'm not sure if there is any agreement on whether the `url` for a pull registry must be just a hostname or if it can be `https://registry.example.com`, but technically, nothing is stopping anyone from using a full URL. It might be safer to check this the same way as the source registry url, i.e. `RegistryURI(...).docker_uri`

rcerven: good point will do

rcerven: fixed

chmeliik:
Waiving osbs-docs, imagestream names are an implementation detail

rcerven:
rn updated: create imagestreams with registry name

PR #983, opened 2020-05-28
Propagate build_env variables from config map

chmeliik:
* CLOUDBLD-1228

Add variables to buildconfig, let openshift propagate them into
buildroot.

Signed-off-by: Adam Cmiel <acmiel@redhat.com>

### DIFF
--- a/osbs/build/build_requestv2.py
+++ b/osbs/build/build_requestv2.py
@@ -88,6 +88,11 @@ class BaseBuildRequest(object):
         if koji_task_id is not None:
             self.set_label('koji-task-id', str(koji_task_id))
 
+        # Should run near the end of render() to avoid setting any special
+        # environment variables (such as USER_PARAMS, REACTOR_CONFIG) in case
+        # someone sets them in reactor_config
+        self.set_build_env_vars(data)
+
         return self.template
 
     def render_custom_strategy(self):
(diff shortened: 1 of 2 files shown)


osbs/build/build_requestv2.py
```diff
@@ -304,6 +304,14 @@ def set_data_from_reactor_config(self, reactor_config_data):
...
         self.set_registry_organization(reactor_config_data)
         self.set_required_secrets(reactor_config_data)
+        self.set_build_env(reactor_config_data)
+
+    def set_build_env(self, reactor_config_data):
+        """
+        Propagates build_env variables set in config map to build environment
+        """
+        build_env = reactor_config_data.get('build_env', [])
+        self.template['spec']['strategy']['customStrategy']['env'].extend(build_env)
```
MartinBasti, line 314: :eyes: can we trust that configmap has correct format here?

MartinBasti:
I'm thinking about blacklisting some names. like USER_PARAMS, BUILD, etc. We use for different purpose to avoid overwrites (or making sure that these won't be overwritten somehow else)

osbs/build/build_requestv2.py
chmeliik: Well it does get validated by schema, I guess I could write this in a way that is more likely to fail if we change the schema

chmeliik:
> I'm thinking about blacklisting some names. like USER_PARAMS, BUILD, etc. We use for different purpose to avoid overwrites (or making sure that these won't be overwritten somehow else)

Couple changes:
- `set_build_env_vars()` will not set duplicate variables
- moved the method out of `set_data_from_reactor_config()`, because it is now important for it to run at the end of `render()`
- reverted old unit test and added a new one specifically for setting build_env_vars

osbs/build/build_requestv2.py
MartinBasti: It's too late for validation by atomic-rector :)

osbs/build/build_requestv2.py
```diff
@@ -305,6 +310,26 @@ def set_data_from_reactor_config(self, reactor_config_data):
...
+        build_env_vars = reactor_config_data.get('build_env_vars', [])
+        for var in build_env_vars:
+            if var['name'] not in existing_vars:
+                # Could be just `env.append(var)`, but here we are using the reactor config
+                # without validating it (that happens later, in atomic-reactor). Do it this
+                # way to fail early if format does not match schema.
+                env.append({'name': var['name'], 'value': var['value']})
+                existing_vars.add(var['name'])
+                logger.info('Set environment variable from reactor config: %s', var['name'])
+            else:
```
MartinBasti, line 329: I'd prefer fail, this warning will nobody notice and this shouldn't happen

chmeliik: Good point, changed

osbs/build/build_requestv2.py
chmeliik: Yeah I realized, should we do some more thorough validation here? Or is it fine the way it is now?

MartinBasti: I'd say it's fine now

chmeliik:
Rebased, should fix py2 failure

chmeliik:
Release notes: Environment variables can be defined in reactor config map, they will be propagated to build environment

PR #984, opened 2020-05-28
Fix test.sh breakage for Cent7xPy2...

ben-alkov:
...due to setuptools dropping Py2 support

Signed-off-by: Ben Alkov <ben.alkov@redhat.com>

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -71,14 +71,8 @@ function setup_osbs() {
     $RUN mkdir -p /usr/local/lib/python3.6/site-packages/
   fi
 
-  # CentOS needs to have setuptools updates to use wildcards in requirements.txt and to make pytest-cov work
-  if [[ $OS != "fedora" ]]; then
-    $RUN $PIP install -U setuptools
-
-    # Watch out for https://github.com/pypa/setuptools/issues/937
-    $RUN curl -O https://bootstrap.pypa.io/2.6/get-pip.py
-    $RUN $PYTHON get-pip.py
-  fi
+  $RUN $PIP install -U pip
+  $RUN $PIP install -U setuptools
   $RUN $PYTHON setup.py install
 
   # Install packages for tests


ben-alkov:
FYI, I don't *believe* that this effects atomic-reactor, but I could be wrong :)

test.sh
```diff
@@ -73,8 +73,12 @@ function setup_osbs() {
 
   # CentOS needs to have setuptools updates to use wildcards in requirements.txt and to make pytest-cov work
   if [[ $OS != "fedora" ]]; then
-    $RUN $PIP install -U setuptools
-
+    if [[ ${PYTHON_VERSION} == '2' ]]; then
```
MartinBasti, line 76: In this case you can just always upgrade pip, don't care about distro

ben-alkov: OK, wasn't sure if we had some reason for *not* doing it...

ben-alkov:
> I'd prefer to have a unified solution for this (note: dockerfile-parse may need this fix too)

WDYGT should be the solution? I'm partial to the one here (Martin's idea) - pinning setuptools is very brittle; Martin's point (to me) was that (an up-to-date) pip will figure out the needed setuptools version based on platfrom/Py version (which does work as evidenced by the Travis runs in here :) )

ben-alkov:
Even better is to just force-update both pip and setuptools for all platforms/interpreters :smile:

chmeliik:
> @chmeliik, @athos-ribeiro, @MartinBasti; WDYT should be the solution? I'm partial to the one here (Martin's idea) - pinning setuptools is very brittle; Martin's point (to me) was that (an up-to-date) pip will figure out the needed setuptools version based on platfrom/Py version (which _does_ work, as evidenced by the Travis runs in here :) )

@ben-alkov, I think the reason why this was not done in koji-containerbuild is that we are using the native version of pip (or, in the case of CentOS, installed from EPEL), which may be more reliable than the PyPI version (downstream patches etc.).

OTOH, this solution has the advantage of protecting against other packages potentially becoming incompatible in the future. I don't mind doing it this way, but I'll defer to others here :)

athos-ribeiro:
I'm OK with the proposal. If we go with this one, it would be nice to propagate the change to all other repositories for consistency.

chmeliik:
No release notes, fix for testing only

PR #986, opened 2020-06-09
Refactor user params

chmeliik:
* CLOUDBLD-124

### DIFF
--- a/osbs/build/user_params.py
+++ b/osbs/build/user_params.py
@@ -82,9 +82,7 @@ def load_user_params_from_json(user_params_json):
     json_dict = json.loads(user_params_json)
     kind = json_dict.get(KIND_KEY, BuildUserParams.KIND)  # BW comp. default to BuildUserParams
     user_params_class = user_param_kinds[kind]
-    user_params = user_params_class()
-    user_params.from_json(user_params_json)
-    return user_params
+    return user_params_class.from_json(user_params_json)
 
 
 class BuildCommon(BuildParamsBase):
@@ -230,19 +228,18 @@ class BuildCommon(BuildParamsBase):
             missing_repr = ", ".join(repr(p.name) for p in missing)
             raise OsbsValidationException("Missing required params: {}".format(missing_repr))
 
-    def from_json(self, user_params_json):
+    @classmethod
+    def from_json(cls, user_params_json):
         if not user_params_json:
-            return
+            return cls()
         try:
             json_dict = json.loads(user_params_json)
         except ValueError:
             logger.debug('failed to convert %s', user_params_json)
             raise
-        for key, value in json_dict.items():
-            try:
-                setattr(self, key, value)
-            except AttributeError:
-                continue
+        # Drop invalid keys
+        json_dict = {k: v for k, v in json_dict.items() if cls.get_param(k) is not None}
+        return cls(**json_dict)
 
     def to_dict(self, keys):
         retdict = {}
--- a/osbs/build/user_params.py
+++ b/osbs/build/user_params.py
@@ -7,14 +7,12 @@ of the BSD license. See the LICENSE file for details.
 """
 from __future__ import print_function, absolute_import, unicode_literals
 
-from abc import abstractproperty, ABCMeta
 import logging
 import re
 import random
 import json
 
-import six
-
+from osbs.build.user_params_meta import BuildParam, BuildParamsBase
 from osbs.constants import (DEFAULT_GIT_REF, REACTOR_CONFIG_ARRANGEMENT_VERSION,
                             DEFAULT_CUSTOMIZE_CONF, RAND_DIGITS,
                             WORKER_MAX_RUNTIME, ORCHESTRATOR_MAX_RUNTIME,
@@ -41,61 +39,37 @@ def register_user_params(klass):
     return klass
 
 
-class BuildParam(object):
-    """ One parameter of a spec """
-
-    def __init__(self, name, default=None, allow_none=False):
-        self.name = name
-        self.allow_none = allow_none
-        self._value = default
-
-    @property
-    def value(self):
-        return self._value
-
-    @value.setter
-    def value(self, val):
-        logger.debug("%s = '%s'", self.name, val)
-        self._value = val
-
-    def __repr__(self):
-        return "BuildParam(%s='%s')" % (self.name, self.value)
-
-
 class UserParam(BuildParam):
     """ custom class for "user" parameter with postprocessing """
-    name = "user"
 
-    def __init__(self):
-        super(UserParam, self).__init__(self.name)
+    def __init__(self, **kwargs):
+        super(UserParam, self).__init__("user", **kwargs)
 
 
 class BuildIDParam(BuildParam):
     """ validate build ID """
-    name = "name"
 
-    def __init__(self):
-        super(BuildIDParam, self).__init__(self.name)
+    def __init__(self, **kwargs):
+        super(BuildIDParam, self).__init__("name", **kwargs)
 
-    @BuildParam.value.setter  # pylint: disable=no-member
-    def value(self, val):  # pylint: disable=W0221
+    def __set__(self, obj, value):
         # build ID has to conform to:
         #  * 63 chars at most
         #  * (([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?
 
-        if len(val) > 63:
+        if len(value) > 63:
             # component + timestamp > 63
-            new_name = val[:63]
-            logger.warning("'%s' is too long, changing to '%s'", val, new_name)
-            val = new_name
+            new_name = value[:63]
+            logger.warning("'%s' is too long, changing to '%s'", value, new_name)
+            value = new_name
 
         build_id_re = re.compile(r"^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$")
-        match = build_id_re.match(val)
+        match = build_id_re.match(value)
         if not match:
-            logger.error("'%s' is not valid build ID", val)
+            logger.error("'%s' is not valid build ID", value)
             raise OsbsValidationException("Build ID '%s', doesn't match regex '%s'" %
-                                          (val, build_id_re))
-        BuildParam.value.fset(self, val)  # pylint: disable=no-member
+                                          (value, build_id_re))
+        super(BuildIDParam, self).__set__(obj, value)
 
 
 def load_user_params_from_json(user_params_json):
@@ -113,55 +87,40 @@ def load_user_params_from_json(user_params_json):
     return user_params
 
 
-@six.add_metaclass(ABCMeta)
-class BuildCommon(object):
-    """Abstract class for user parameters"""
+class BuildCommon(BuildParamsBase):
+    """Common user parameters, class should be considered abstract"""
 
-    @abstractproperty
-    def KIND(self):
-        return 'DEFINE_KIND_NAME_IN_SUBCLASS'
+    # Must be defined in subclasses
+    KIND = NotImplemented
 
-    def __init__(self, build_json_store=None):
-        self.arrangement_version = BuildParam(
-            "arrangement_version",
-            allow_none=True,
-            default=REACTOR_CONFIG_ARRANGEMENT_VERSION)
-        # build_from contains the full build_from string, including the source type prefix
-        self.build_from = BuildParam('build_from')
-        # build_image contains the buildroot name, whether the buildroot is a straight image or an
-        # imagestream.  buildroot_is_imagestream indicates what type of buildroot
-        self.build_image = BuildParam('build_image')
-        self.buildroot_is_imagestream = BuildParam('buildroot_is_imagestream', default=False)
-        self.build_json_dir = BuildParam('build_json_dir', default=build_json_store)
-        self.kind = BuildParam(KIND_KEY, default=self.KIND)
-        self.component = BuildParam('component')
-        self.image_tag = BuildParam("image_tag")
-        self.koji_target = BuildParam("koji_target", allow_none=True)
-        self.koji_task_id = BuildParam('koji_task_id', allow_none=True)
-        self.platform = BuildParam("platform", allow_none=True)
-        self.orchestrator_deadline = BuildParam('orchestrator_deadline', allow_none=True)
-        self.reactor_config_map = BuildParam("reactor_config_map", allow_none=True)
-        self.reactor_config_override = BuildParam("reactor_config_override", allow_none=True)
-        self.scratch = BuildParam('scratch', allow_none=True)
-        self.signing_intent = BuildParam('signing_intent', allow_none=True)
-        self.user = UserParam()
-        self.worker_deadline = BuildParam('worker_deadline', allow_none=True)
+    arrangement_version = BuildParam("arrangement_version",
+                                     default=REACTOR_CONFIG_ARRANGEMENT_VERSION)
+    # build_from contains the full build_from string, including the source type prefix
+    build_from = BuildParam("build_from")
+    # build_image contains the buildroot name, whether the buildroot is a straight image or an
+    # imagestream.  buildroot_is_imagestream indicates what type of buildroot
+    build_image = BuildParam("build_image")
+    buildroot_is_imagestream = BuildParam("buildroot_is_imagestream", default=False)
+    build_json_dir = BuildParam("build_json_dir", required=True)
+    component = BuildParam("component")
+    image_tag = BuildParam("image_tag")
+    koji_target = BuildParam("koji_target")
+    koji_task_id = BuildParam("koji_task_id")
+    platform = BuildParam("platform")
+    orchestrator_deadline = BuildParam("orchestrator_deadline")
+    reactor_config_map = BuildParam("reactor_config_map")
+    reactor_config_override = BuildParam("reactor_config_override")
+    scratch = BuildParam("scratch")
+    signing_intent = BuildParam("signing_intent")
+    user = UserParam(required=True)
+    worker_deadline = BuildParam("worker_deadline")
 
-        self.required_params = [
-            self.build_json_dir,
-            self.koji_target,
-            self.user,
-        ]
-        self.convert_dict = {}
+    def __init__(self, build_json_dir=None, **kwargs):
+        super(BuildCommon, self).__init__(build_json_dir=build_json_dir, **kwargs)
 
-    def attrs_finalizer(self):
-        for _, param in self.__dict__.items():
-            if isinstance(param, BuildParam):
-                # check that every parameter has a unique name
-                if param.name in self.convert_dict:
-                    raise OsbsValidationException(
-                        'Two user params with the same name: {}'.format(param.name))
-                self.convert_dict[param.name] = param
+    def __setattr__(self, name, value):
+        super(BuildCommon, self).__setattr__(name, value)
+        logger.debug("%s = %s", name, value)
 
     def set_params(self,
                    build_conf=None,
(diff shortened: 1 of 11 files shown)


osbs/build/user_params.py
```diff
@@ -202,22 +162,24 @@ def set_params(self,
         :param reactor_config_map: str, name of the config map containing the reactor environment
         :param worker_deadline: int, worker completion deadline in hours
         """
+        # TODO: see if setting params from build_conf can be separated from setting other params,
+        #   if yes, move most of this method to __init__
```
chmeliik, line 157: Here's the solutions that I came up with for refactoring `set_params()`:
* make it `__init__()`
  * `+` validation could be done right away during init
  * `-` the `build_conf` argument is not for a BuildParam (other BuildParams are set from build_conf), so it might be confusing
* put most of it into `__init__()`, add a `set_params_from_build_conf()` method
  * `+` separates basic param setting from the additional build_conf logic
  * `-` set_params does some cross-validation between build_conf and normal params (specifically, it checks that `build_from` is set either in build_conf or as a normal param) - this would be lost
  * `-` validation would have to be done by calling `validate()`
* turn it into a classmethod that creates a new instance
  * `+` validation could be done right away in `<name of method>`
  * `-` what should be the name of the method?

osbs/build/user_params.py
```diff
@@ -113,55 +87,41 @@ def load_user_params_from_json(user_params_json):
...
+    orchestrator_deadline = BuildParam("orchestrator_deadline")
+    reactor_config_map = BuildParam("reactor_config_map")
+    reactor_config_override = BuildParam("reactor_config_override")
+    scratch = BuildParam("scratch")
+    signing_intent = BuildParam("signing_intent")
+    user = UserParam(required=True)
+    worker_deadline = BuildParam("worker_deadline")
+
+    def __init__(self, build_json_dir=None, **kwargs):
+        # TODO: why is is build_json_dir special?
```
chmeliik, line 110: Was there any specific reason why `build_json_dir` had to be set in `__init__()` rather than together with the other params?

osbs/build/user_params.py
```diff
@@ -288,22 +243,22 @@ def from_json(self, user_params_json):
...
-    def set_if_exists(self, json_dict, param):
-        if self.convert_dict[param].value:
-            json_dict[param] = self.convert_dict[param].value
-
     def to_dict(self, keys):
         retdict = {}
         for key in keys:
-            self.set_if_exists(retdict, key)
+            value = getattr(self, key)
+            # TODO: why are we not including falsy values?
```
chmeliik, line 244: Is there a good reason for this? Unit tests expect this behaviour, but it is strange.

osbs/build/user_params.py
```diff
@@ -312,60 +267,54 @@ class BuildUserParams(BuildCommon):
...
+    auto_build_node_selector = BuildParam("auto_build_node_selector",
+                                          include_in_json=False)
+    explicit_build_node_selector = BuildParam("explicit_build_node_selector",
+                                              include_in_json=False)
+    isolated_build_node_selector = BuildParam("isolated_build_node_selector",
+                                              include_in_json=False)
+    platform_node_selector = BuildParam("platform_node_selector",
+                                        include_in_json=False)
+    scratch_build_node_selector = BuildParam("scratch_build_node_selector",
+                                             include_in_json=False)
```
chmeliik, line 304: Previously, these were not defined as `BuildParam` => they would not be included in `to_json()` output. Unlike `repo_info`, node selectors have JSON-compatible values. Is there a reason why they should not appear in JSON?

osbs/build/user_params.py
```diff
@@ -312,60 +267,54 @@ class BuildUserParams(BuildCommon):
...
+
+    # TODO: see if this can (and should) be moved elsewhere
+    repo_info = BuildParam("repo_info", include_in_json=False)
```
chmeliik, line 307: `repo_info` does not sound like a user param. Should this be moved elsewhere (probably to `BuildRequestV2.__init__()`)? It would still have to be passed to `BuildUserParams` because, just like `build_conf`, it may be used for setting values of other params.

chmeliik:
Still a WIP, but please take a look. I've commented on the most important TODOs, which I will likely not be able to figure out on my own.

osbs/build/user_params.py
chmeliik: Tried option 3, see last commit (the validation part is not complete)

chmeliik: Looks like there might be a case for not validating during creation: https://github.com/containerbuildsystem/osbs-client/blob/56d648c020948db68c74143086d2cdc06add6aa9/osbs/build/build_requestv2.py#L73-L74

Not sure if there really is a reason for not validating user params, but for now, I'm calling it good enough.

osbs/build/user_params.py
chmeliik: After refactoring set_params, it does not seem to be the case

chmeliik:
Should be mostly done now.

Not all changes necessarily have to be included. When this reaches final form, some commits could be squashed.

tkdchen:
@chmeliik Remove "[WIP]" from the last two commits?

chmeliik:
@tkdchen removed

Also changed how kwargs are handled in `make_params()`, because the previous implementation could lead to `TypeError: ... got multiple values for keyword argument ...` if, for example, `reactor_config_map` was passed as a kwarg to `make_params()` (this does actually happen in atomic-reactor, though I'm not entirely sure why)

osbs/build/user_params.py
MartinBasti: Not in this PR, you can open issue

osbs/build/user_params.py
MartinBasti: Don't include it into user_params. We should have it in BuildRequest directly

MartinBasti: Keep it as is here for now. (No export to json)

osbs/build/user_params.py
MartinBasti: You are fast and you have fixed it already.
Good

osbs/build/user_params_meta.py
```diff
@@ -0,0 +1,176 @@
...
+            if param is not None:
+                return param
+        return None
+
+    @property
+    def params(cls):
+        """
+        Get all params for a class
+        """
+        # pylint: disable=no-member; pylint does not understand metaclass properties
```
MartinBasti, line 132: :eyes: poor pylint

osbs/build/user_params.py
MartinBasti: :eyes:

osbs/build/user_params.py
```diff
@@ -202,22 +162,24 @@ def set_params(self,
         :param reactor_config_map: str, name of the config map containing the reactor environment
         :param worker_deadline: int, worker completion deadline in hours
         """
+        # TODO: see if setting params from build_conf can be separated from setting other params,
```
MartinBasti: :eyes:

osbs/build/user_params.py
MartinBasti: It's probably meant to not include `None` I cannot figure out why falsy values should be skiped here

osbs/build/user_params.py
```diff
@@ -463,6 +411,8 @@ def set_params(self,
         :param git_ref: str, commit ID of the branch to be pulled
         :param git_uri: str, uri of the git repository for the source
         """
+        # TODO: see if setting params from build_conf and repo_info can be separated from setting
```
MartinBasti: :eyes:

osbs/build/user_params.py
```diff
@@ -576,12 +523,13 @@ def set_params(
                                               to fetch sources
         :return:
         """
+        # TODO: if the set_params methods of other classes can be scrapped, scrap this one too
```
MartinBasti: :eyes:

osbs/build/user_params.py
```diff
@@ -233,19 +231,18 @@ def validate(self):
             missing_repr = ", ".join(repr(p.name) for p in missing)
             raise OsbsValidationException("Missing required params: {}".format(missing_repr))
 
-    def from_json(self, user_params_json):
+    @classmethod
```
MartinBasti, line 245: <3

osbs/build/user_params_meta.py
```diff
@@ -0,0 +1,176 @@
...
+    def include_in_json(self):
+        return self._include_in_json
+
+    def __repr__(self):
+        return "{self.__class__.__name__}({self.name!r})".format(self=self)
+
+    def __get__(self, obj, objtype=None):
+        try:
+            # Bypass potential __getattr__ redefinition in class
+            return object.__getattribute__(obj, self._mangled_name)
```
lcarva, line 70: Could you elaborate on why simply using something like `self.val` is not sufficient? I don't fully understand why we're setting this attribute on the owner object, e.g. `BuildParamsBase`.

osbs/build/user_params.py
```diff
@@ -141,61 +145,75 @@ def set_params(self,
...
+        return params
 
-        self._populate_image_tag()
+    @classmethod
+    def _make_params_super(cls, *args, **kwargs):
+        # Pylint cannot properly infer the return type of an overriden classmethod
+        # that returns cls(). This is an ugly workaround that prevents pylint from
+        # inferring any type at all (thus preventing false-positive warnings).
+        # See https://github.com/PyCQA/pylint/issues/981
+        return BuildCommon.make_params.__func__(cls, *args, **kwargs)
```
tkdchen, line 216: How do we usually track such workaround?

Just a quick idea, add a `TODO:` for this method to tell "remove this workaround once the pylint issue is fixed"?

osbs/build/user_params_meta.py
chmeliik: One of the main goals of this refactoring was to change `BuildParam` to a class attribute instead of an instance attribute (basically the opposite of [this](https://github.com/containerbuildsystem/osbs-client/commit/a2b06d4e0b1d4b7c0241cf2500409d55008476e7), as you mentioned).

If the value was stored in the `BuildParam` instance, it would be shared among all instances of `BuildParamsBase`. Example:

```python
class SharedValue(object):
    def __init__(self):
        self._value = None

    def __get__(self, obj, objtype=None):
        return self._value

    def __set__(self, obj, value):
        self._value = value


class Foo(object):
    x = SharedValue()


foo = Foo()
bar = Foo()

foo.x = 1
print(foo.x, bar.x)  # 1 1
bar.x = 2
print(foo.x, bar.x)  # 2 2
```

That is exactly what we *do not* want, so the attribute has to be set on the owner to prevent this behaviour. See this test: https://github.com/chmeliik/osbs-client/blob/0b4e0fe81cc21372055bdfae647becd5a63f6912/tests/build_/test_user_params_meta.py#L206-L219

osbs/build/user_params_meta.py
chmeliik: yeah, pylint really had a bad time with my PR :D

osbs/build/user_params.py
chmeliik: This (and most of the remaining TODOs) is removed in later commits. Should I remove the TODO from this commit as well to make it look like it never existed?

osbs/build/user_params_meta.py
```diff
@@ -0,0 +1,176 @@
...
+    def name(self):
+        return self._name
+
+    @property
+    def required(self):
+        return self._required
+
+    @property
+    def include_in_json(self):
+        return self._include_in_json
```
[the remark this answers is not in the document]
chmeliik: I'm paranoid so I often use properties to prevent others (and myself) from changing attributes that should be immutable. There is also the possibility that some subclass may want to override the `include_in_json` (or, more likely, `required`) property. But I'm not strictly opposed to making this a simple attribute, let's see what the others think.

osbs/build/user_params.py
chmeliik: @MartinBasti usually dislikes TODOs, but in this case it might be valid?

chmeliik:
Minor cleanup (1 unneeded change, 1 typo, 1 TODO)

chmeliik:
I realized that default values for BuildParam are useless if they can be immediately overridden by setting `None` - see last commit. Not sure if the behaviour isn't a little confusing, especially this: https://github.com/chmeliik/osbs-client/blob/4d30155629cce61564c334ef654eb522e5cf393f/tests/build_/test_user_params_meta.py#L93-L94

WDYT?

osbs/build/user_params.py
MartinBasti: TODOs will rot there forever, if you have intention to fix in future open tracking ticket.
I don't think that it's worth tracking.

MartinBasti:
> I realized that default values for BuildParam are useless if they can be immediately overridden by setting `None` - see last commit. Not sure if the behaviour isn't a little confusing, especially this: https://github.com/chmeliik/osbs-client/blob/4d30155629cce61564c334ef654eb522e5cf393f/tests/build_/test_user_params_meta.py#L93-L94
> 
> WDYT?

You set none but it returns default? It is confusing

chmeliik:
> You set none but it returns default? It is confusing

That's the problem of using `None` for `<not set>` in a nutshell, I guess I could make a compromise and only ignore `None` in `__init__`.

Currently, the problem is that params will **never** have their default value, because you have:
```python
def make_params(cls, some_param_with_default=None):
    # some_param_with_default is immediately set to None, default is overwritten
    return cls(some_param_with_default=some_param_with_default)
```

Or, I could remove all the named params from `make_params` and keep only `**kwargs` and the docstring, that's not very nice though.

osbs/build/user_params.py
chmeliik: Ok, not adding TODO, not tracking

chmeliik:
Dropped last commit (`Do not overwrite defaults by setting None`), fixed by ignoring `None` in `make_params()`

osbs/build/user_params.py
chmeliik: I think I'll leave it as it is for now, it worked fine all this time and, apparently, adding unnecessary things to the json output might be a bad idea.

MartinBasti:
Interesting F32 test build error. We can waive this

```
=================================== FAILURES ===================================
______________________________ test_get_repo_info ______________________________

tmpdir = local('/tmp/pytest-of-mockbuild/pytest-0/test_get_repo_info0')

    def test_get_repo_info(tmpdir):
        repo_path = tmpdir.mkdir("repo").strpath
        with open(os.path.join(repo_path, REPO_CONTAINER_CONFIG), 'w') as f:
            f.write(dedent("""\
                compose:
                    modules:
                    - n:s:v
                """))
    
>       initialize_git_repo(repo_path, files=[REPO_CONTAINER_CONFIG])

tests/utils/test_utils.py:531: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/utils/test_utils.py:547: in initialize_git_repo
    first_commit_ref = subprocess.check_output(['git', 'rev-parse', 'HEAD'], cwd=rpath)
/usr/lib64/python3.8/subprocess.py:411: in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

input = None, capture_output = False, timeout = None, check = True
popenargs = (['git', 'rev-parse', 'HEAD'],)
kwargs = {'cwd': '/tmp/pytest-of-mockbuild/pytest-0/test_get_repo_info0/repo', 'stdout': -1}
process = <subprocess.Popen object at 0x7f9d6230b940>, stdout = b'HEAD\n'
stderr = None, retcode = 128

    def run(*popenargs,
            input=None, capture_output=False, timeout=None, check=False, **kwargs):
        """Run command with arguments and return a CompletedProcess instance.
    
        The returned instance will have attributes args, returncode, stdout and
        stderr. By default, stdout and stderr are not captured, and those attributes
        will be None. Pass stdout=PIPE and/or stderr=PIPE in order to capture them.
    
        If check is True and the exit code was non-zero, it raises a
        CalledProcessError. The CalledProcessError object will have the return code
        in the returncode attribute, and output & stderr attributes if those streams
        were captured.
    
        If timeout is given, and the process takes too long, a TimeoutExpired
        exception will be raised.
    
        There is an optional argument "input", allowing you to
        pass bytes or a string to the subprocess's stdin.  If you use this argument
        you may not also use the Popen constructor's "stdin" argument, as
        it will be used internally.
    
        By default, all communication is in bytes, and therefore any "input" should
        be bytes, and the stdout and stderr will be bytes. If in text mode, any
        "input" should be a string, and stdout and stderr will be strings decoded
        according to locale encoding, or by "encoding" if set. Text mode is
        triggered by setting any of text, encoding, errors or universal_newlines.
    
        The other arguments are the same as for the Popen constructor.
        """
        if input is not None:
            if kwargs.get('stdin') is not None:
                raise ValueError('stdin and input arguments may not both be used.')
            kwargs['stdin'] = PIPE
    
        if capture_output:
            if kwargs.get('stdout') is not None or kwargs.get('stderr') is not None:
                raise ValueError('stdout and stderr arguments may not be used '
                                 'with capture_output.')
            kwargs['stdout'] = PIPE
            kwargs['stderr'] = PIPE
    
        with Popen(*popenargs, **kwargs) as process:
            try:
                stdout, stderr = process.communicate(input, timeout=timeout)
            except TimeoutExpired as exc:
                process.kill()
                if _mswindows:
                    # Windows accumulates the output in a single blocking
                    # read() call run on child threads, with the timeout
                    # being done in a join() on those threads.  communicate()
                    # _after_ kill() is required to collect that and add it
                    # to the exception.
                    exc.stdout, exc.stderr = process.communicate()
                else:
                    # POSIX _communicate already populated the output so
                    # far into the TimeoutExpired exception.
                    process.wait()
                raise
            except:  # Including KeyboardInterrupt, communicate handled that.
                process.kill()
                # We don't call process.wait() as .__exit__ does that for us.
                raise
            retcode = process.poll()
            if check and retcode:
>               raise CalledProcessError(retcode, process.args,
                                         output=stdout, stderr=stderr)
E               subprocess.CalledProcessError: Command '['git', 'rev-parse', 'HEAD']' returned non-zero exit status 128.

/usr/lib64/python3.8/subprocess.py:512: CalledProcessError
----------------------------- Captured stdout call -----------------------------
Initialized empty Git repository in /tmp/pytest-of-mockbuild/pytest-0/test_get_repo_info0/repo/.git/
----------------------------- Captured stderr call -----------------------------
fatal: not in a git directory

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: empty ident name (for <"ghost@example.com">) not allowed
fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

osbs/build/user_params_meta.py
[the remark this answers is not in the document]
chmeliik: Will keep it as it is for now

chmeliik:
Removed TODOs from the commit which introduced them (they were removed in later commits anyway, so overall it makes no difference)

osbs/build/user_params_meta.py
lcarva: So we want a class attribute to behave like an instance attribute? It's unclear to me why we want this. Python's syntax for accessing such attributes is exactly the same. The only difference I can think of is where those attributes are defined:

```python
class Foobar():
  spam = BuildParam()

  def __init__(self, ...):
    sef.spam = ...

# versus

class Foobar():
  def __init__(self, ...):
    self.spam = BuildParam(...)
```

Those values need to be initialized anyways in `__init__`, so why not also define them there?

lcarva:
It looks like my comment is hidden by default, but I'm concerned about the complexity vs value of this approach: https://github.com/containerbuildsystem/osbs-client/pull/986#pullrequestreview-431761744

osbs/build/user_params_meta.py
lcarva: I've misunderstood how data descriptors are working under the hood. The example I posted above is invalid and does not work as expected. They must be set as a class attributes.

MartinBasti:
No release notes, internal change only

PR #989, opened 2020-06-30
Add the "remote_source.packages" key to the JSON schema

mprahl:
This is to support forwarding new parameters to Cachito to
support multiple packages in a single source repository.

Relates to CLOUDBLD-1598

Signed-off-by: mprahl <mprahl@users.noreply.github.com>

### DIFF
--- a/osbs/schemas/container.json
+++ b/osbs/schemas/container.json
@@ -106,6 +106,24 @@
                         "enable-party-popper"
                       ]
                   }
+                },
+                "packages": {
+                  "description": "the packages object to pass to the Cachito request",
+                  "type": ["object", "null"],
+                  "examples": [
+                    {"npm": [{"path": "client"}, {"path": "proxy"}]}
+                  ],
+                  "additionalProperties" : {
+                    "type" : "array",
+                    "items": {
+                      "type": "object",
+                      "examples": [
+                        {"path": "client"},
+                        {"path": "proxy"}
+                      ],
+                      "minProperties": 1
+                    }
+                  }
                 }
               },
               "additionalProperties": false,


osbs/schemas/container.json
```diff
@@ -106,6 +106,24 @@
...
                       ]
                   }
+                },
+                "packages": {
+                  "description": "the packages object to pass to the Cachito request",
+                  "type": ["object", "null"],
+                  "examples": [
+                    {"npm": [{"path": "client"}, {"path": "proxy"}]}
+                  ],
+                  "additionalProperties" : {
```
athos-ribeiro, line 116: Why isn't this "properties" instead?

We are usually quite restrictive in the container.yaml files in dist-git. So if a user has a typo or is trying to use some invalid key here, we just halt early during validation.

That said, if there is no specific reason for using additional properties here, could we change it to `properties` and set `additionalProperties` to `false`?

osbs/schemas/container.json
```diff
@@ -106,6 +106,24 @@
...
+                    "type" : "array",
+                    "items": {
+                      "type": "object",
```
athos-ribeiro, line 119: is this object properties restricted to `path`? If so, can we make it explicit here?

Note that this file is linked in the upstream documentation (in hopes that users will refer to it when writing their container.yaml files; some do).

osbs/schemas/container.json
mprahl: @athos-ribeiro I used `additionalProperties` so that the schema does not need to be updated every time a new package manager is added to Cachito.

osbs/schemas/container.json
mprahl: It is as of now, but it could be expanded as use-cases pop up. I specifically made this less restrictive so that the schema does not need to be updated every time Cachito adds support for a new property.

rcerven:
rn upated: - Add the "remote_source.packages" key to schema for container.yaml

PR #992, opened 2020-08-07
Build only content sets used just for building content

rcerven:
* CLOUDBLD-2046

### DIFF
--- a/osbs/schemas/container.json
+++ b/osbs/schemas/container.json
@@ -268,6 +268,21 @@
                 "rhel-8.2.1-modules"
               ]
           }
+        },
+        "build_only_content_sets": {
+          "desription": "Content sets used only for building content, not for distribution",
+          "type": ["object", "null"],
+          "patternProperties": {
+            "^[a-z].*$": {
+              "type": "array",
+              "description": "List of content sets for specific architecture",
+              "minItems": 1,
+              "items": {
+                "type": "string",
+                "pattern": "^.*-rpms$"
+              }
+            }
+          }
         }
       },
       "additionalProperties": false


rcerven:
related atomic-reactor PR https://github.com/containerbuildsystem/atomic-reactor/pull/1539

osbs/schemas/container.json
```diff
@@ -268,6 +268,21 @@
                 "rhel-8.2.1-modules"
               ]
           }
+        },
+        "dev_content_sets": {
```
MartinBasti, line 272: Still not usre if this should be rather build_content_sets than dev

rcerven: @lcarva @athos-ribeiro @ben-alkov  ?

ben-alkov: I'm not sure that I have an opinion, but my observation here is that the word 'build' is HUGELY overloaded in all of OSBS... (what we really mean here is "build-time", or "build-dep")

rcerven: agreed on : build_only_content_sets

athos-ribeiro: As per the atomic-reactor PR [1], this seems to be just regular content sets. In other words, there are no restrictions on how the users will use these content sets (we are just trusting they will not accidentally ship any RPMs from those content sets).

@lcarva, is this consequence worth the benefit? i.e., are false positives better than false negatives when assessing the container image contents? Alternatively, should we ensure these packages are not used in the final state somehow?

Back to the naming issue, these are purposefully unaccounted content sets. They are unaccounted because they are only used in intermediate stages. In other words, they are content sets whose contents are not shipped. So, how about `unshipped_content_sets` or `nonshipped_content_sets`?

[1] https://github.com/containerbuildsystem/atomic-reactor/pull/1539

lcarva: There's current testing to ensure the content used in the final image matches the defined content sets in content_sets.yml. However, there are some cases, where ODCS integration is used, that additional content sets are needed only for building the image.

Testing should help identify when content_sets are incorrectly being listed as `build_only_content_sets`. Furthermore, failure to distinguish between these different usages of content sets makes it more difficult to make the existing tests mandatory.

rcerven:
rn updated: Build only content sets used just for building content

PR #998, opened 2020-09-02
Dashes should not be allowed in version tag

rcerven:
* CLOUDBLD-2222

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -43,7 +43,7 @@ from osbs.constants import (BUILD_RUNNING_STATES, WORKER_OUTER_TEMPLATE,
                             ORCHESTRATOR_CUSTOMIZE_CONF, BUILD_TYPE_WORKER,
                             BUILD_TYPE_ORCHESTRATOR, BUILD_FINISHED_STATES,
                             DEFAULT_ARRANGEMENT_VERSION, REACTOR_CONFIG_ARRANGEMENT_VERSION,
-                            FILTER_KEY, RELEASE_LABEL_FORMAT,
+                            FILTER_KEY, RELEASE_LABEL_FORMAT, VERSION_LABEL_FORBIDDEN_CHARS,
                             ORCHESTRATOR_SOURCES_OUTER_TEMPLATE,
                             USER_PARAMS_KIND_IMAGE_BUILDS,
                             USER_PARAMS_KIND_SOURCE_CONTAINER_BUILDS,
@@ -652,6 +652,22 @@ class OSBS(object):
                 raise OsbsValidationException("release label doesn't have proper format")
         except KeyError:
             pass
+
+        try:
+            _, version_value = labels.get_name_and_value(Labels.LABEL_TYPE_VERSION)
+        # version doesn't exist
+        except KeyError:
+            pass
+        else:
+            if version_value:
+                wrong_chars = \
+                    [denied for denied in VERSION_LABEL_FORBIDDEN_CHARS if denied in version_value]
+
+                if wrong_chars:
+                    msg = "version label '{}' contains not allowed chars : '{}'".\
+                        format(version_value, wrong_chars)
+                    logger.error(msg)
+                    raise OsbsValidationException(msg)
 
         if required_missing:
             raise OsbsValidationException("required label missing from Dockerfile")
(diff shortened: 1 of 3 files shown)


osbs/api.py
```diff
@@ -653,6 +653,17 @@ def _check_labels(self, repo_info):
...
             pass
 
+        try:
+            _, version_value = labels.get_name_and_value(Labels.LABEL_TYPE_VERSION)
+            if (version_value and
+                    any(denied in version_value for denied in VERSION_LABEL_FORBIDDEN_CHARS)):
+
+                logger.error("version label '%s' contains not allowed chars : %s", version_value,
+                             VERSION_LABEL_FORBIDDEN_CHARS)
+                raise OsbsValidationException("version label contains not allowed chars")
```
MartinBasti, line 663: Can we specify which one?

MartinBasti:
do we know allowed values? I.e. just having check in opposite way?

osbs/api.py
ben-alkov: +1 for `denied` instead of `VERSION_LABEL_FORBIDDEN_CHARS`

ben-alkov:
@rcerven; Does this effect users in any way (i.e. do we need a doc update/release note)?

osbs/api.py
rcerven: sure

rcerven:
> @rcerven; Does this effect users in any way (i.e. do we need a doc update/release note)?

not really, because that limitation is already in koji, so without this fix, build will fail anyway in koji_import

rcerven:
> do we know allowed values? I.e. just having check in opposite way?

in koji source I've found only raise with '-'

osbs/api.py
MartinBasti: I meant in that exception, so it's shown to a user directly

rcerven: added there as well :)

osbs/api.py
```diff
@@ -653,6 +653,20 @@ def _check_labels(self, repo_info):
...
+            _, version_value = labels.get_name_and_value(Labels.LABEL_TYPE_VERSION)
+            if version_value:
+                wrong_chars = \
+                    [denied for denied in VERSION_LABEL_FORBIDDEN_CHARS if denied in version_value]
+
+                if wrong_chars:
+                    logger.error("version label '%s' contains not allowed chars : %s",
+                                 version_value, wrong_chars)
+                    raise OsbsValidationException("version label contains not allowed chars %s" %
+                                                  wrong_chars)
```
MartinBasti, line 666:
```suggestion
                    msg = "version label '{}' contains not allowed chars : '{}'".format(version_value, wrong_chars)
                    logger.error(msg)
                    raise OsbsValidationException(msg)
```

osbs/api.py
```diff
@@ -653,6 +653,20 @@ def _check_labels(self, repo_info):
...
+        except KeyError:
+            pass
```
tkdchen, line 668: Which line above could raise KeyError? Suggest to write a comment to tell why this error can be ignored.

tests/test_api.py
```diff
@@ -3046,51 +3047,53 @@ def mock_get_image_stream(*args, **kwargs):
...
-                [general]
-                build_json_dir = inputs
-                [default]
-                build_from = image:buildroot:latest
-                openshift_url = /
-                """))
-            fp.flush()
-            config = Configuration(fp.name)
-            osbs_obj = OSBS(config, config)
+    def test_release_and_version_label_validation(self, osbs, label_type, label_value,
```
tkdchen, line 3071: This test case tests invalid release and version label only, but the name `test ... validation` is too generic. Suggest to rename it to something specifically like `test_raise_error_if_release_or_version_label_is_invalid`.

osbs/api.py
rcerven: ok

osbs/api.py
rcerven: getting version label, added comment

tests/test_api.py
rcerven: sure

osbs/api.py
MartinBasti: I really like:
```python
try:
except:
else:
```
for such cases, when try block is as small as possible

rcerven: done

MartinBasti (approved):

tkdchen (approved):

ben-alkov (approved):

rcerven:
rn updated: Dashes should not be allowed in version tag

PR #999, opened 2020-09-07
Validate tags type in container.yaml

tkdchen:
* CLOUDBLD-1560

Signed-off-by: Chenxiong Qi <cqi@redhat.com>


rcerven:
code looks good, but we should actually reconsider the approach

tkdchen:
@rcerven I added a new commit to allow writing tags in either string or number in container.yaml. Please have a look at the result behavior. If it is expected, I'll squash the commits.

osbs/schemas/container.json
```diff
@@ -351,7 +351,7 @@
     "tags": {
       "type": "array",
       "items": {
-        "type": "string"
+        "type": ["string", "number"]
```
chmeliik, line 354: Allowing floats *could* have unexpected results, e.g.:
```python
>>> str(yaml.safe_load('1.1111111111111111'))                                             
'1.1111111111111112'
```
No idea if tags with this many "decimal places" are ever used, just something to be careful about.

tkdchen: Yeah, this could an actual problem if a user would like to use such kind of tag.

I personally prefer to restrict the data in the input side as much as possible, that is user should know that he/she should write the tag name in the `tags:` section in a string format with double-quotes. This should be easy enough for both user and OSBS to handle the data without any potential problems like such one.

Meanwhile I also think it makes sense and is natural for user to realize that the data written in the container.yaml should be in correct data format.

rcerven (approved):

osbs/schemas/container.json
ben-alkov: +1 for early user input validation

MartinBasti (approved):

rcerven:
rn updated: Validate tags type in container.yaml

PR #1000, opened 2020-09-30
Log HTTP error responses

MartinBasti:
Log response text from OCP
    
Added hook to requests to log response when a HTTP request failed.
    
However with retries it's tricky, if failure limit is reached, it will
just fail withou passing response to hooks. To be able to process
request while using retries, Retry(raise_on_status=False) has to be
used.
    
However this param is available from urllib3 > 1.15, in RHEL7 there is
urllib3 ~= 1.10, so response text will be logged only with newer urllib3
version.
    
<s>Also Retry() rasises exception by default when a HTTP request was not
successful. To keep this behavior and not break current code, another
hook has been added to raise exception when request was not successful.</s>
(no need for this, code checks responses and this change would even break unittests)
    
* CLOUDBLD-2800

### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -11,6 +11,7 @@ abstraction on top of http api calls
 
 from __future__ import print_function, absolute_import, unicode_literals
 
+from distutils.version import LooseVersion
 import sys
 import logging
 import json
@@ -31,6 +32,7 @@ try:
 except ImportError:
     HTTPKerberosAuth = None
 
+from urllib3 import __version__ as urllib3_version
 from urllib3.exceptions import InsecureRequestWarning
 from urllib3.util import Retry
 from urllib3 import disable_warnings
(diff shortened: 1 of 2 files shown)


MartinBasti:
PR 1000 :1st_place_medal:

mkosiarc (approved):

tests/test_http.py
```diff
@@ -26,6 +28,12 @@ def s():
     return HttpSession(verbose=True)
 
 
+@pytest.fixture(scope='module')
+def error_response_logged():
+    """Error response loggin works only with urllib3>1.15"""
```
ben-alkov, line 33: Typo: "loggin"

osbs/http.py
```diff
@@ -75,6 +77,18 @@ def request(self, url, *args, **kwargs):
...
 
+def make_retry(**kwargs):
+    """Make initialized Retry object based on urllib3 version
+
+    :param kwargs: kwargs acceptable by urllib3.util.Retry class
+    :return: urllib3.util.Retry object
+    """
+    if LooseVersion(urllib3_version) < LooseVersion('1.15'):
+        # `raise_on_status` is not supported with older versions of urllib3 (RHEL7)
+        kwargs.pop('raise_on_status', None)
```
ben-alkov, line 88: Clever :wink:

lkolacek (approved):

tests/test_http.py
MartinBasti: fixed

ben-alkov (approved):

MartinBasti:
Release notes:
- OCP error response content is logged

PR #1001, opened 2020-10-01
HttpStream: fix close method

MartinBasti:
`.close()` is called from `__del__` and because this happend during
garbage collecting nothing is guaranted [1] and it results into following
traceback printed directly to stderr because it happened in GC context:
```
  File "/home/mbasti/repos-upstream/osbs-client/osbs/http.py", line 221, in __del__
    self.close()
  File "/home/mbasti/repos-upstream/osbs-client/osbs/http.py", line 214, in close
    if not self.closed:
AttributeError: 'HttpStream' object has no attribute 'closed'
```

Using getattr instead to avoid traceback.

TLDR: use context managers if you want to be sure that connection is closed
properly.

[1] https://docs.python.org/3/reference/datamodel.html#object.__del__

### DIFF
--- a/osbs/http.py
+++ b/osbs/http.py
@@ -188,11 +188,12 @@ class HttpStream(object):
             return
 
     def close(self):
-        if not self.closed:
+        # using getattr and hasattr because this may be called from __del__
+        if not getattr(self, 'closed', True):
             logger.debug("cleaning up")
             if hasattr(self, 'req'):
                 del self.req
-        self.closed = True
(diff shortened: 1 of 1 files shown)


MartinBasti:
Or we can just delete `__del__` method

rcerven (approved):

mkosiarc (approved):

MartinBasti:
No release notes, minor fix

PR #1005, opened 2020-10-21
api.py: Catch mismatches between the flatpak parameter and the repo

owtaylor:
If a build is started with the 'flatpak' parameter, but doesn't have a
container.yaml with a 'flatpak' section, or a build is started without
the 'flatpak' parameter, but has a Flatpak container.yaml, the user
would get super-confusing failures, since the repo-parsing part of the code is
independent of the part of the code that checks for build type.

Check for consistency, and throw helpful messages.

Signed-off-by: Owen W. Taylor <otaylor@fishsoup.net>

### DIFF
--- a/osbs/api.py
+++ b/osbs/api.py
@@ -756,6 +756,17 @@ class OSBS(object):
 
         repo_info = utils.get_repo_info(git_uri, git_ref, git_branch=git_branch,
                                         depth=git_commit_depth)
+
+        if flatpak and not repo_info.configuration.is_flatpak:
+            raise OsbsException(
+                "Flatpak build, "
+                "but repository doesn't have a container.yaml with a flatpak: section")
+
+        if not flatpak and repo_info.configuration.is_flatpak:
+            raise OsbsException(
+                "Not a flatpak build, "
+                "but repository has a container.yaml with a flatpak: section")
+
         req_labels = self._check_labels(repo_info)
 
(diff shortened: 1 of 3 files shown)


MartinBasti:
4 tests has to be fixed

osbs/api.py
```diff
@@ -756,6 +756,17 @@ def _do_create_prod_build(self,
 
         repo_info = utils.get_repo_info(git_uri, git_ref, git_branch=git_branch,
                                         depth=git_commit_depth)
+
+        if flatpak and not repo_info.configuration.is_flatpak:
+            raise OsbsException(
+                "Flatpak build, " +
```
MartinBasti, line 762:
```suggestion
                "Flatpak build, "
```

owtaylor:
Pushed a new version that fixes the broken tests and uses literal concatenation for the line breaks :-)

MartinBasti:
> Pushed a new version that fixes the broken tests and uses literal concatenation for the line breaks :-)

The same test failures and commit looks the same as previous one to me

owtaylor:
> The same test failures and commit looks the same as previous one to me

Bah! Actually amended my commit before pushing now.

MartinBasti (approved):

lkolacek (approved):

MartinBasti:
Release notes:
- catch mismatched between flatpak parameter and repo configuration

PR #1008, opened 2021-01-21
Fail build if container or content_sets file have unexpected extension

mkosiarc:
Closes CLOUDBLD-3471

- [n/a] Pull request has a link to an osbs-docs PR for user documentation updates

### DIFF
--- a/osbs/constants.py
+++ b/osbs/constants.py
@@ -79,6 +79,11 @@ ATOMIC_REACTOR_LOGGING_FMT = \
 REPO_CONFIG_FILE = '.osbs-repo-config'
 ADDITIONAL_TAGS_FILE = 'additional-tags'
 REPO_CONTAINER_CONFIG = 'container.yaml'
+REPO_CONTAINER_CONFIG_POSSIBLE_TYPOS = {'container.yml'}
+REPO_CONTENT_SETS_FILE = 'content_sets.yml'
+REPO_CONTENT_SETS_FILE_POSSIBLE_TYPOS = {'content_sets.yaml',
+                                         'content-sets.yml',
+                                         'content-sets.yaml'}
 
 # number of retries for http requests
 HTTP_MAX_RETRIES = 8
--- a/osbs/repo_utils.py
+++ b/osbs/repo_utils.py
@@ -10,7 +10,12 @@ of the BSD license. See the LICENSE file for details.
 from __future__ import print_function, absolute_import, unicode_literals
 
 from osbs.exceptions import OsbsException, OsbsValidationException
-from osbs.constants import REPO_CONFIG_FILE, ADDITIONAL_TAGS_FILE, REPO_CONTAINER_CONFIG
+from osbs.constants import (REPO_CONFIG_FILE,
+                            ADDITIONAL_TAGS_FILE,
+                            REPO_CONTAINER_CONFIG,
+                            REPO_CONTAINER_CONFIG_POSSIBLE_TYPOS,
+                            REPO_CONTENT_SETS_FILE,
+                            REPO_CONTENT_SETS_FILE_POSSIBLE_TYPOS)
 from osbs.utils.labels import Labels
 from osbs.utils.yaml import read_yaml_from_file_path
 from six import StringIO
@@ -129,6 +134,7 @@ class RepoConfiguration(object):
         self.git_uri = git_uri
         self.git_branch = git_branch
         self.git_ref = git_ref
+        self.dir_path = dir_path
 
         # Set default options
         self._config_parser.readfp(StringIO(self.DEFAULT_CONFIG))   # pylint: disable=W1505; py2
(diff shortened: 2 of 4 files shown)


tests/test_repo_utils.py
```diff
@@ -342,6 +347,47 @@ def test_tags_type_validation(self, tmpdir, tag, should_fail):
...
+                            correct_file=REPO_CONTENT_SETS_FILE,
+                            wrong_file=REPO_CONTENT_SETS_FILE_WRONG_EXTENSION,
+                        ),
+                ),
+                ([REPO_CONFIG_FILE], ''),
+                ([REPO_CONTENT_SETS_FILE], ''),
+        ),
+    )
+    def test_repo_files_extensions(self, tmpdir, files, error_msg):
+        for file in files:
```
ben-alkov, line 382: Don't shadow builtin `file`; maybe `repo_file`?

mkosiarc: noticed it, changed it, and then changed it again because I like your suggestion more.

osbs/repo_utils.py
```diff
@@ -171,6 +170,54 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE, depth=None,
...
+                                            REPO_CONTAINER_CONFIG_WRONG_EXTENSION)
+
+        if (os.path.exists(container_config_path) and
+                os.path.exists(container_config_wrong_ext_path)):
+            msg = ('This repo contains both {correct_file} and {wrong_file}'
+                   .format(correct_file=REPO_CONTAINER_CONFIG,
+                           wrong_file=REPO_CONTAINER_CONFIG_WRONG_EXTENSION))
+            raise OsbsException(msg)
+        elif os.path.exists(container_config_wrong_ext_path):
+            msg = ('Container config file has wrong extension please change it to {correct_file}'
```
ben-alkov, line 191: Maybe?
```suggestion
            msg = ('Container config filename has wrong extension; please change it to {correct_file}'
```

osbs/repo_utils.py
```diff
@@ -171,6 +170,54 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE, depth=None,
...
+        content_sets_path = os.path.join(dir_path, REPO_CONTENT_SETS_FILE)
+        content_sets_wrong_ext_path = os.path.join(dir_path, REPO_CONTENT_SETS_FILE_WRONG_EXTENSION)
+
+        if os.path.exists(content_sets_path) and os.path.exists(content_sets_wrong_ext_path):
+            msg = ('This repo contains both {correct_file} and {wrong_file}'
+                   .format(correct_file=REPO_CONTENT_SETS_FILE,
+                           wrong_file=REPO_CONTENT_SETS_FILE_WRONG_EXTENSION))
+            raise OsbsException(msg)
+        elif os.path.exists(content_sets_wrong_ext_path):
+            msg = ('Content sets file has wrong extension please change it to {correct_file}'
```
ben-alkov, line 217: As above

osbs/constants.py
```diff
@@ -79,6 +79,10 @@
 REPO_CONFIG_FILE = '.osbs-repo-config'
 ADDITIONAL_TAGS_FILE = 'additional-tags'
 REPO_CONTAINER_CONFIG = 'container.yaml'
+REPO_CONTAINER_CONFIG_WRONG_EXTENSION = 'container.yml'
+REPO_CONTENT_SETS_FILE = 'content_sets.yml'
+REPO_CONTENT_SETS_FILE_WRONG_EXTENSION = 'content_sets.yaml'
```
MartinBasti, line 84: I have crazy idea what about having sets of potential typos?

```python
REPO_CONTAINER_CONFIG_WRONG = {'container.yml', 'contianer.yaml'}
REPO_CONTENT_SETS_FILE_WRONG = {'content_sets.yaml', 'content-sets.yml'}
```

osbs/repo_utils.py
```diff
@@ -171,6 +170,54 @@ def __init__(self, dir_path='', file_name=REPO_CONFIG_FILE, depth=None,
     def is_autorebuild_enabled(self):
         return self._config_parser.getboolean('autorebuild', 'enabled')
 
+    def _check_and_validate_container_config_file(self, dir_path):
```
MartinBasti, line 173: Seeing `*_and_*` pattern in functions makes me cry. 

`check` can be just single single function used with parameters (eventually 2 partial function made from it)

When checks passes just call another function `validate` after

osbs/constants.py
tkdchen: I have another crazy idea after reading this PR which makes me remind of another task that validates a few user config files inside an atomic-reactor plugin with predefined JSON schema.

To be short, we could abstract a container repository into a dedicated object.

The object could provides methods to check and validate user config files with JSON schema and properties to get the file names or read file content. For example:

```python
class ContainerRepo:
    def __init__(self, root_dir):
        self._root_dir = root_dir
        # check files, e.g. ensure the correct container.yaml file name

    def validate(self):
        # validate various user configs, e.g. content_set.yml
        # there could be separate validation method for each specific user config file

    @property
    def dockerfile(self):
        return os.path.join(self._root_dir, 'Dockerfile')

    @property
    def content_set_yml(self):
        return os.path.join(self._root_dir, 'content_sets.yml')

    @property
    def container_yml(self):
        return os.path.join(self._root_dir, 'container.yml')

    def read_content_sets(self):
        """Return content of the content_sets.yml"""
```

With this abstraction, all the code related to handling the files inside a container repo will be in one place, not spread into osbs-client and atomic-reactor, and we could handle more things inside this object as well.

Another benefit would be by reading this object, it should be much easier to know what files could be included in a container repo and what files should be included for a successful container build.

I'm not requesting changes to this PR based on this thought. Just sharing my thought and FYI.

osbs/repo_utils.py
mkosiarc: fixed

osbs/repo_utils.py
mkosiarc: fixed

osbs/constants.py
mkosiarc: @tkdchen feel free to file a story, mbasti said he likes the idea too.

osbs/repo_utils.py
mkosiarc: fixed

MartinBasti (approved):

rcerven (approved):

tkdchen (approved):

PR #1010, opened 2021-01-22
Use newer version fedora image in CI

tkdchen:
* CLOUDBLD-3581

Signed-off-by: Chenxiong Qi <cqi@redhat.com>

### DIFF
--- a/.github/workflows/unittests.yaml
+++ b/.github/workflows/unittests.yaml
@@ -23,12 +23,12 @@ jobs:
             engine: docker
 
           - name: fedora
-            version: 31
+            version: 32
             python: 3
             engine: docker
 
           - name: fedora
-            version: 32
+            version: 33
             python: 3
             engine: docker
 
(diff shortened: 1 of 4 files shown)


mkosiarc (approved):

.github/workflows/unittests.yaml
```diff
@@ -23,12 +23,12 @@ jobs:
             engine: docker
 
           - name: fedora
-            version: 31
```
mkosiarc, line 26: this commit says replace f31 with f33, how should I understand this ?

MartinBasti: it was changed on both places, F31->F32 and F32->F33

MartinBasti (approved):

.github/workflows/unittests.yaml
tkdchen: I just changed the order of the list :)

mkosiarc (approved):

PR #1015, opened 2021-03-15
Complete loggers by new level and message handler for user warnings

ssalatsk:
*CLOUDBLD-4640

Signed-off-by: Serhii Salatskyi <ssalatsk@redhat.com>

### DIFF
--- a/osbs/constants.py
+++ b/osbs/constants.py
@@ -76,6 +76,10 @@ RAND_DIGITS = 5
 ATOMIC_REACTOR_LOGGING_FMT = \
     '%(asctime)s platform:%(arch)s - %(name)s - %(levelname)s - %(message)s'
 
+# Numeric representation of user warnings loggers' level
+# User warning level is logically between INFO (20) and WARNING (30) levels
+USER_WARNING_LEVEL = 25
+
 REPO_CONFIG_FILE = '.osbs-repo-config'
 ADDITIONAL_TAGS_FILE = 'additional-tags'
 REPO_CONTAINER_CONFIG = 'container.yaml'
--- a/osbs/utils/__init__.py
+++ b/osbs/utils/__init__.py
@@ -9,6 +9,7 @@ from __future__ import print_function, absolute_import, unicode_literals
 from functools import wraps
 import contextlib
 import copy
+import json
 import logging
 import os
 import os.path
@@ -28,7 +29,8 @@ from hashlib import sha256
 from osbs.repo_utils import RepoConfiguration, RepoInfo, AdditionalTagsConfig
 from osbs.constants import (OS_CONFLICT_MAX_RETRIES, OS_CONFLICT_WAIT,
                             GIT_MAX_RETRIES, GIT_BACKOFF_FACTOR, GIT_FETCH_RETRY,
-                            OS_NOT_FOUND_MAX_RETRIES, OS_NOT_FOUND_MAX_WAIT)
+                            OS_NOT_FOUND_MAX_RETRIES, OS_NOT_FOUND_MAX_WAIT,
+                            USER_WARNING_LEVEL)
 
 # This was moved to a separate file - import here for external API compatibility
 from osbs.utils.labels import Labels  # noqa: F401
@@ -586,6 +588,20 @@ def retry_on_exception(exception_type):
         return retry
 
     return do_retry_on_exception
+
+
+def user_warning_log_handler(self, message):
+    """
+    Take arguments to transform them into JSON data
+    and send them into the logger with USER_WARNING level
+    """
+    assert isinstance(message, str)
+
+    content = {
+        'message': message,
+    }
+    msg = json.dumps(content)
+    self._log(USER_WARNING_LEVEL, msg)
 
 
 class RetryFunc(object):
--- a/osbs/__init__.py
+++ b/osbs/__init__.py
@@ -9,10 +9,16 @@ from __future__ import print_function, absolute_import, unicode_literals
 
 import logging
 
+from osbs.constants import USER_WARNING_LEVEL
+from osbs.utils import user_warning_log_handler
 from osbs.version import __version__  # noqa
 
 
 def set_logging(name="osbs", level=logging.DEBUG):
+    # add new level to loggers
+    logging.addLevelName(USER_WARNING_LEVEL, "USER_WARNING")
+    logging.Logger.user_warning = user_warning_log_handler
+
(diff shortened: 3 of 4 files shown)


osbs/utils/__init__.py
```diff
@@ -588,6 +590,13 @@ def retry(*args, **kwargs):
     return do_retry_on_exception
 
 
+def user_warning_log_handler(self, msg, *args, **kwargs):
+    if isinstance(msg, dict):
+        msg = json.dumps(msg, indent=2)
```
MartinBasti, line 595: IMHO we want single line, no indentation

osbs/utils/__init__.py
```diff
@@ -588,6 +590,13 @@ def retry(*args, **kwargs):
     return do_retry_on_exception
 
 
+def user_warning_log_handler(self, msg, *args, **kwargs):
+    if isinstance(msg, dict):
```
MartinBasti, line 594: Can we have an abstraction instead of raw dict. Fucntion:

```
def log_user_warning(msg, ...potential_future_extensions):
    jsonized_msg = magic(....)
    logger.user_warning(jsonized_msg)
```

osbs/utils/__init__.py
```diff
@@ -588,6 +590,13 @@ def retry(*args, **kwargs):
...
+        msg = json.dumps(msg, indent=2)
+
+    self._log(USER_WARNING_LEVEL, msg, args, **kwargs)
```
MartinBasti, line 597: We quite violates how logger works here. It expect formatting string and args are applied to it, but when dict is passed into it, it's transformed into string, can it have some side effects?

ssalatsk: Good catch :+1:

osbs/utils/__init__.py
MartinBasti: You probably misunderstood what I wanted. I'd like to have and extra function, that will create proper json dict. Instead of calling
```
logger.user_warning({'msg': 'yolo'})
```
I want to have abstraction
```
logger.user_warning(msg='yolo')
```

MartinBasti: extensible for future:
```
logger.user_warning(msg='yolo', doc_ecode=123, kind='deprecation')
```

MartinBasti: User warning istself should be responsible for creating that dictionary used for json

osbs/utils/__init__.py
```diff
@@ -588,6 +590,17 @@ def retry(*args, **kwargs):
...
 
 
+def user_warning_log_handler(self, msg, *args, **kwargs):
+    def convert_warn_message(msg):
+        if isinstance(msg, dict):
+            return json.dumps(msg)
+
+        return msg
+
+    msg = convert_warn_message(msg)
```
tkdchen, line 600: How is inner function `convert_warn_message` useful for this case?

osbs/constants.py
```diff
@@ -76,6 +76,9 @@
 ATOMIC_REACTOR_LOGGING_FMT = \
     '%(asctime)s platform:%(arch)s - %(name)s - %(levelname)s - %(message)s'
 
+# Numeric representation of user warnings loggers' level
+USER_WARNING_LEVEL = 25
```
tkdchen, line 80: It would be good to give an informative description of number 25.

tkdchen (approved):
LGTM.

osbs/utils/__init__.py
ssalatsk: It's deprecated now

osbs/constants.py
ssalatsk: Added

osbs/utils/__init__.py
```diff
@@ -588,6 +590,16 @@ def retry(*args, **kwargs):
     return do_retry_on_exception
 
 
+def user_warning_log_handler(self, **kwargs):
+    allowed_keywords = ['message']
+
+    # Allow parameters with specified keys only
+    kwargs = {k: v for k, v in kwargs if k in allowed_keywords}
```
MartinBasti, line 597: So you silently ignores invalid keyword?

osbs/utils/__init__.py
```diff
@@ -588,6 +590,16 @@ def retry(*args, **kwargs):
...
+
+    msg = json.dumps(kwargs)
+    self._log(USER_WARNING_LEVEL, msg)
```
MartinBasti, line 600: How about enforcing it at API level? So devs will know what's allowed from function description
```suggestion
def user_warning_log_handler(self, message: str):
    assert isinstance(message, str)
    content = {
        'message': message
    }
    msg = json.dumps(content)
    self._log(USER_WARNING_LEVEL, msg)
```

ssalatsk: Fixed, thanks for suggesting :+1:

osbs/utils/__init__.py
```diff
@@ -588,6 +590,17 @@ def retry(*args, **kwargs):
     return do_retry_on_exception
 
 
+def user_warning_log_handler(self, message):
+    if not isinstance(message, str):
```
MartinBasti, line 594: I'd say just assert is ok as this is used only by developers

osbs/utils/__init__.py
```diff
@@ -588,6 +590,17 @@ def retry(*args, **kwargs):
     return do_retry_on_exception
 
 
+def user_warning_log_handler(self, message):
```
MartinBasti, line 593: So we don't want py typing in new code? (as you didn't transformed it from my example here)

MartinBasti: Missing docstring

ssalatsk: Because osbs-client still supports Py2 and function annotations aren't implemented there (see [PEP 3107](https://www.python.org/dev/peps/pep-3107/)), we cannot add it here :(

MartinBasti: Aah, true sorry. I'm living on the edge

MartinBasti (approved):

MartinBasti:
@tkdchen do you agree with changes I requested? Can we merge?

rcerven (approved):

tkdchen (approved):
@MartinBasti Yes.

MartinBasti:
@ssalatsk please provide release notes

ssalatsk:
Release notes:
- Loggers can handle and store explicit user warnings

PR #1018, opened 2021-03-19
Print user warnings to osbs-client command line output

ssalatsk:

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -30,7 +30,7 @@ from osbs.exceptions import (OsbsNetworkException, OsbsException, OsbsAuthExcept
                              OsbsResponseException)
 from osbs.cli.capture import setup_json_capture
 from osbs.utils import (paused_builds, TarReader, TarWriter, get_time_from_rfc3339,
-                        graceful_chain_get, ImageName)
+                        graceful_chain_get, UserWarningsStore, ImageName)
 from six.moves.urllib.parse import urljoin
 
 logger = logging.getLogger('osbs')
@@ -302,7 +302,9 @@ def cmd_cancel_build(args, osbs):
 
 
 def _print_build_logs(args, osbs, build):
+    user_warnings = UserWarningsStore()
     build_id = build.get_build_name()
+
     # we need to wait for kubelet to schedule the build, otherwise it's 500
     build = osbs.wait_for_build_to_get_scheduled(build_id)
     if not args.no_logs:
@@ -313,11 +315,20 @@ def _print_build_logs(args, osbs, build):
         print("Build submitted (%s), watching logs (feel free to interrupt)" % build_id)
         try:
             for line in build_logs:
+                if user_warnings.is_user_warning(line):
+                    user_warnings.store(line)
+                    continue
+
                 print('{!r}'.format(line))
         except Exception as ex:
             logger.error("Error during fetching logs for build %s: %s", build_id, repr(ex))
 
         osbs.wait_for_build_to_finish(build_id)
+
+        if user_warnings:
+            print("USER WARNINGS")
+            print(user_warnings)
+
         return _display_build_summary(osbs.get_build(build_id))
     else:
         if args.output == 'json':
@@ -426,15 +437,24 @@ def cmd_build_logs(args, osbs):
     build_id = args.BUILD_ID[0]
     follow = args.follow
 
+    user_warnings = UserWarningsStore()
     logs = osbs.get_build_logs(build_id, follow=follow,
                                wait_if_missing=args.wait_if_missing,
                                decode=True)
 
-    if follow:
-        for line in logs:
+    for line in logs:
+        if user_warnings.is_user_warning(line):
+            user_warnings.store(line)
+
+        if follow:
             print(line)
-        return
-    print(logs, end="")
+
+    if not follow:
+        print(logs, end="\n")
+
+    if user_warnings:
+        print("USER WARNINGS")
+        print(user_warnings)
 
 
 def cmd_watch_build(args, osbs):
--- a/osbs/utils/__init__.py
+++ b/osbs/utils/__init__.py
@@ -30,7 +30,7 @@ from osbs.repo_utils import RepoConfiguration, RepoInfo, AdditionalTagsConfig
 from osbs.constants import (OS_CONFLICT_MAX_RETRIES, OS_CONFLICT_WAIT,
                             GIT_MAX_RETRIES, GIT_BACKOFF_FACTOR, GIT_FETCH_RETRY,
                             OS_NOT_FOUND_MAX_RETRIES, OS_NOT_FOUND_MAX_WAIT,
-                            USER_WARNING_LEVEL)
+                            USER_WARNING_LEVEL, USER_WARNING_LEVEL_NAME)
 
 # This was moved to a separate file - import here for external API compatibility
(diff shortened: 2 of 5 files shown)


osbs/cli/main.py
```diff
@@ -313,11 +315,20 @@ def _print_build_logs(args, osbs, build):
         print("Build submitted (%s), watching logs (feel free to interrupt)" % build_id)
         try:
             for line in build_logs:
+                if "- USER_WARNING -" in line:
```
MartinBasti, line 318: :eyes:  Why a hardcoded magic value? Don't we have defined constant for that somewhere?

osbs/utils/__init__.py
```diff
@@ -726,3 +726,31 @@ def copy(self):
             namespace=self.namespace,
             repo=self.repo,
             tag=self.tag)
+
+
+class UserWarningsStore(object):
+    def __init__(self):
+        # asctime (platform:arch)? - name - levelname - message
+        self.__regex = r'^[\d\s,:-]+ (platform:.+ )?- \D+ - USER_WARNING - (\{.*\})$'
```
MartinBasti, line 734: Do we really need two underscores `__` instead of just one `_`?

MartinBasti: Could we just use regexp to `<whatever> - USER_WARNING - (<whatever>)$`?

USER_WARNING should be constant

osbs/utils/__init__.py
```diff
@@ -726,3 +726,31 @@ def copy(self):
...
+        Extract data from given log record with USER_WARNING level
+        and store an understandable message in set
+        """
+        data_search = re.search(self.__regex, line)
+        if not data_search:
+            return
+
+        try:
+            data = json.loads(data_search.group(2))
+        except ValueError:
```
MartinBasti, line 748: Should we at least log something?

osbs/cli/main.py
```diff
@@ -426,15 +437,24 @@ def cmd_build_logs(args, osbs):
...
+
+        if follow:
             print(line)
-        return
-    print(logs, end="")
+
+    if not follow:
+        print(logs, end="\n")
+
+    print("---------------------------- USER WARNINGS ----------------------------")
```
MartinBasti, line 455: This is not nice, terminals have different length. Lets go without `-----`

osbs/utils/__init__.py
```diff
@@ -726,3 +726,36 @@ def copy(self):
             namespace=self.namespace,
             repo=self.repo,
             tag=self.tag)
+
+
+class UserWarningsStore(object):
+    def __init__(self):
+        # asctime (platform:arch)? - name - levelname - message
+        self.regex = r'^.+ - USER_WARNING - (\{.*\})$'
+        self.__user_warnings = set()
```
rcerven, line 735: why 2 underscores? use just 1

osbs/utils/__init__.py
```diff
@@ -726,3 +726,36 @@ def copy(self):
...
+        try:
+            data = json.loads(data_search.group(1))
+        except ValueError:
+            message = 'Incorrect JSON data input for a user warning: %s'
+            logger.warning(message, data_search.group(1))
+            return
+
+        message = data['message']
+
+        self.__user_warnings.add(message)
```
rcerven, line 758: just use directly data

self._user_warnings.add(data['message'])

ssalatsk: I've made it for future extensions, maybe we'll have to combine retrieved data, so I've decided to save the message into a corresponding variable

osbs/cli/main.py
```diff
@@ -313,11 +315,19 @@ def _print_build_logs(args, osbs, build):
...
+                    user_warnings.store(line)
+                    continue
+
                 print('{!r}'.format(line))
         except Exception as ex:
             logger.error("Error during fetching logs for build %s: %s", build_id, repr(ex))
 
         osbs.wait_for_build_to_finish(build_id)
+
+        print("USER WARNINGS")
```
MartinBasti, line 328: maybe we should print this only when there are some warnings present

ssalatsk: Good catch :+1:

osbs/cli/main.py
```diff
@@ -313,11 +315,20 @@ def _print_build_logs(args, osbs, build):
...
+                    user_warnings.store(line)
+                    continue
+
                 print('{!r}'.format(line))
         except Exception as ex:
             logger.error("Error during fetching logs for build %s: %s", build_id, repr(ex))
 
         osbs.wait_for_build_to_finish(build_id)
+
+        if str(user_warnings):
```
MartinBasti, line 328: :eyes:

MartinBasti: `__len__`, `__bool__` ?

ssalatsk: It allows us to control if there is some content in user_warnings. Is there something wrong?

ssalatsk: Good point :+1:

osbs/utils/__init__.py
```diff
@@ -726,3 +726,39 @@ def copy(self):
...
+        return re.match(self.regex, line)
+
+    def store(self, line):
+        """
+        Extract data from given log record with USER_WARNING level
+        and store an understandable message in set
+        """
+        data_search = re.search(self.regex, line)
+        if not data_search:
+            return
```
MartinBasti, line 747: Nitpick: This should fail if we try to store something else than user warning

osbs/utils/__init__.py
```diff
@@ -726,3 +726,39 @@ def copy(self):
             namespace=self.namespace,
             repo=self.repo,
             tag=self.tag)
+
+
+class UserWarningsStore(object):
+    def __init__(self):
+        # asctime (platform:arch)? - name - levelname - message
+        self.regex = r'^.+ - USER_WARNING - (\{.*\})$'
```
MartinBasti, line 734: Loglevel name should be constant in `osbs.constants` and that constant should be used here instead of hardcoding the same thing on several places
https://github.com/containerbuildsystem/osbs-client/blob/master/osbs/__init__.py#L19

osbs/utils/__init__.py
```diff
@@ -726,3 +726,41 @@ def copy(self):
             namespace=self.namespace,
             repo=self.repo,
             tag=self.tag)
+
+
+class UserWarningsStore(object):
+    def __init__(self):
+        # (asctime (platform:arch)? - name) - levelname - message
+        self.regex = r' - '.join((r'^.+', USER_WARNING_LEVEL_NAME, r'(\{.*\})$'))
```
MartinBasti, line 734: Why not `r'...'.format(USER_WARNIN.....)` ?

ssalatsk: Because I added additional brackets here, it requires an additional argument for this string. Here is the thrown exception: 
`>       self.regex = r'^.+ - {} -(\{.*\})$'.format(USER_WARNING_LEVEL_NAME)

E       IndexError: Replacement index 1 out of range for positional args tuple`

MartinBasti: `r'^.+ - {} -({{.*\}})$'.format(USER_WARNING_LEVEL_NAME)` This will work, but ok, it seems equally ugly

MartinBasti (approved):

rcerven (approved):

rcerven:
rn updated: Print user warnings to osbs-client command line output

PR #1019, opened 2021-03-19
User warnings handler requires positional arguments

ssalatsk:
*CLOUDBLD-4640

Signed-off-by: Serhii Salatskyi <ssalatsk@redhat.com>

- [n/a] Pull request has a link to an osbs-docs PR for user documentation updates

### DIFF
--- a/osbs/utils/__init__.py
+++ b/osbs/utils/__init__.py
@@ -601,7 +601,7 @@ def user_warning_log_handler(self, message):
         'message': message,
     }
     msg = json.dumps(content)
-    self._log(USER_WARNING_LEVEL, msg)
+    self._log(USER_WARNING_LEVEL, msg, None)
 
 
 class RetryFunc(object):
--- a/tests/utils/test_utils.py
+++ b/tests/utils/test_utils.py
@@ -16,6 +16,7 @@ import datetime
 import re
 import sys
 import requests
+import logging
 from time import tzset, sleep
 from pkg_resources import parse_version
 from textwrap import dedent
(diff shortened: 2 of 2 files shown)


ssalatsk:
This change is required due to base log method has required `args` parameter (see [source code](https://github.com/python/cpython/blob/master/Lib/logging/__init__.py#L1589))

rcerven (approved):

osbs/utils/__init__.py
```diff
@@ -601,7 +601,7 @@ def user_warning_log_handler(self, message):
         'message': message,
     }
     msg = json.dumps(content)
-    self._log(USER_WARNING_LEVEL, msg)
+    self._log(USER_WARNING_LEVEL, msg, None)
```
MartinBasti, line 604: Shouldn't be this an empty list instead of None?

ssalatsk: It doesn't matter, `args` apply into record only when they are defined, because it's None, it never will be used (see [code ](https://github.com/python/cpython/blob/28ad12f8fe889a741661eb99daacebd9243cc1ba/Lib/logging/__init__.py#L359) where it applies)

tests/utils/test_utils.py
```diff
@@ -618,7 +618,7 @@ def test_image_name_comparison():
     (11, None)
 ))
 def test_user_warnings_handler(message, expected):
-    def mocked_log(level, message):
+    def mocked_log(level, message, args):
```
MartinBasti, line 621: I consider this test wrong because it already misses issue instead of catching it, I think test should use `caplog` fixture instead of mocking

osbs/utils/__init__.py
MartinBasti: ok

tests/utils/test_utils.py
lkolacek: +1

ssalatsk: The test has been changed

MartinBasti (approved):

lkolacek (approved):

PR #1034, opened 2021-09-22
Add methods for pipeline manipulation

mkosiarc:
CLOUDBLD-6301

### DIFF
--- a/osbs/osbs_http.py
+++ b/osbs/osbs_http.py
@@ -15,8 +15,7 @@ from distutils.version import LooseVersion
 import sys
 import logging
 import json
-from six.moves import http_client
-
+import http
 
 from osbs.exceptions import OsbsException, OsbsNetworkException, OsbsResponseException
 from osbs.constants import (
@@ -53,6 +52,9 @@ class HttpSession(object):
 
     def put(self, url, **kwargs):
         return self.request(url, "put", **kwargs)
+
+    def patch(self, url, **kwargs):
+        return self.request(url, "patch", **kwargs)
 
     def delete(self, url, **kwargs):
         return self.request(url, "delete", **kwargs)
@@ -137,12 +139,12 @@ class HttpStream(object):
         headers = headers or {}
         method = method.lower()
 
-        if method not in ['post', 'get', 'put', 'delete']:
+        if method not in ['post', 'get', 'put', 'patch', 'delete']:
             raise RuntimeError("Unsupported method '%s' for curl call!" % method)
 
         args = {}
 
-        if method in ['post', 'put']:
+        if method in ['post', 'put', 'patch']:
             headers['Expect'] = ''
 
         if not verify_ssl:
@@ -208,7 +210,7 @@ class HttpStream(object):
             for line in self.req.iter_lines(**kwargs):
                 yield line
         except (requests.exceptions.ChunkedEncodingError,
-                http_client.IncompleteRead):
+                http.client.IncompleteRead):
             return
 
     def close(self):
--- a/osbs/tekton.py
+++ b/osbs/tekton.py
@@ -1,0 +1,696 @@
+"""
+Copyright (c) 2015, 2019, 2021 Red Hat, Inc
+All rights reserved.
+This software may be modified and distributed under the terms
+of the BSD license. See the LICENSE file for details.
+"""
+import json
+import time
+import logging
+import base64
+import os
+import requests
+import copy
+
+
+from osbs.exceptions import OsbsResponseException, OsbsAuthException, OsbsException
+from osbs.constants import (DEFAULT_NAMESPACE, SERVICEACCOUNT_SECRET, SERVICEACCOUNT_TOKEN,
+                            SERVICEACCOUNT_CACRT)
+from osbs.osbs_http import HttpSession
+from osbs.kerberos_ccache import kerberos_ccache_init
+from urllib.parse import urljoin, urlencode, urlparse, parse_qs
+from requests.utils import guess_json_utf
+
+logger = logging.getLogger(__name__)
+
+# Retry each connection attempt after 30 seconds, for a maximum of 10 times
+WATCH_RETRY_SECS = 30
+WATCH_RETRY = 10
+MAX_BAD_RESPONSES = WATCH_RETRY // 3
+# Give up after 12 hours
+WAIT_RETRY_HOURS = 12
+WAIT_RETRY = WAIT_RETRY_HOURS * 3600 // (WATCH_RETRY_SECS * WATCH_RETRY)
+
+API_VERSION = "tekton.dev/v1beta1"
+
+
+def check_response(response, log_level=logging.ERROR):
+    if response.status_code not in (
+            requests.status_codes.codes.ok,
+            requests.status_codes.codes.created,
+    ):
+        if hasattr(response, 'content'):
+            content = response.content
+        else:
+            content = b''.join(response.iter_lines())
+
+        logger.log(log_level, "[%d] %s", response.status_code, content)
+        raise OsbsResponseException(message=content, status_code=response.status_code)
+
+
+class Openshift(object):
+    def __init__(self, openshift_api_url, openshift_oauth_url,
+                 k8s_api_url=None,
+                 verbose=False, username=None, password=None, use_kerberos=False,
+                 kerberos_keytab=None, kerberos_principal=None, kerberos_ccache=None,
+                 client_cert=None, client_key=None, verify_ssl=True, use_auth=None,
+                 token=None, namespace=DEFAULT_NAMESPACE):
+        self.os_api_url = openshift_api_url
+        self.k8s_api_url = k8s_api_url
+        self._os_oauth_url = openshift_oauth_url
+        self.namespace = namespace
+        self.verbose = verbose
+        self.verify_ssl = verify_ssl
+        self._con = HttpSession(verbose=self.verbose)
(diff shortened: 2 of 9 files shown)


osbs/tekton.py
```diff
@@ -0,0 +1,709 @@
...
+            logger.error("[%s] 'Location' header is missing in response, cannot retrieve token",
+                         r.status_code)
+            return ""
+        parsed_url = urlparse(redir_url)
+        fragment = parsed_url.fragment
+        parsed_fragment = parse_qs(fragment)
+        self.token = parsed_fragment['access_token'][0]
+        return self.token
+
+    def get_serviceaccount_tokens(self, username="~"):
```
MartinBasti, line 223: Can you remind me why do we need this?

osbs/osbs_http.py
```diff
@@ -54,6 +54,9 @@ def post(self, url, **kwargs):
     def put(self, url, **kwargs):
```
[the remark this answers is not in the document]
mkosiarc: not duplicated, renamed as there were some issues when importing stuff such as "from http import HTTPStatus"
"

osbs/tekton.py
```diff
@@ -0,0 +1,709 @@
...
+                "namespace": self.namespace,
+                "labels": labels
+            },
+            "spec": {
+                "pipelineRef": {
+                    "name":self.pipeline_name
+                }
+            }
+        }
+        return self.os._patch(url, data=json.dumps(data),
```
MartinBasti, line 432: Will this update or rewrite all labels?

Original implementation updates labels, i.e. it first downloads current labels

osbs/tekton.py
```diff
@@ -0,0 +1,709 @@
...
+            self.wait_for_pod()
+        url = self.os._build_url(
+            self.api_path,
+            self.api_version,
+            f"pods/{self.pod_name}"
+        )
+        r = self.os._get(url)
+        return r.json()
+
+    def get_logs(self, follow=False, wait=False):
```
MartinBasti, line 584: I'm very confused about return value and return type

osbs/tekton.py
```diff
@@ -0,0 +1,709 @@
...
+        for changetype, pod in self.os.watch_resource(self.api_path, self.api_version, resource_type='pods', resource_name=self.pod_name):
+            try:
+                status = pod['status']['phase']
+            except KeyError:
+                logger.error("Pod does not have any status")
+                continue
+            if status in ['Running', 'Succeeded', 'Failed']:
+                return pod
+            else:
+                logger.debug("Pod is not in the state I am waiting for")
```
MartinBasti, line 683: Would be nice to mention what is the status

osbs/osbs_http.py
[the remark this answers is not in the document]
mkosiarc: Hmm, so there was still some issues in the dependencies. and anyway http.py module uses http.client.IncompleteRead exception, so I would prefer to just change the name

osbs/tekton.py
mkosiarc: updated

osbs/tekton.py
mkosiarc: I am not completely sure this going to stay the way it is - https://github.com/containerbuildsystem/osbs-client/blob/master/osbs/cli/main.py#L516

MartinBasti: aha, probably we should also do some cleanup of CLI commands (not here)

osbs/osbs_http.py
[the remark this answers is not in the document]
MartinBasti: ok

osbs/tekton.py
```diff
@@ -0,0 +1,711 @@
...
+                # Are we running inside a pod? If so, we will have a
+                # token available which can be used for authentication
+                self.use_auth = self.can_use_serviceaccount_token()
+        else:
+            self.use_auth = use_auth
+            if not auth_credentials_provided:
+                # We've been told to use authentication but no
+                # credentials have been given. See if we're running
+                # inside a pod, and if so use the provided token.
+                self.can_use_serviceaccount_token()
```
chmeliik: Is it ok to just discard the return value here? Shouldn't this raise an exception if `use_auth and not auth_credentials_provided and not self.can_use_serviceaccount_token()`?

chmeliik (commented):
Ok, so I got part of the way through the PR before I realized I was reviewing copy-pasta'd code from osbs/core.py (man, I hate so much about it).

@mkosiarc could you help me and other reviewers understand what is new here? Ideally in commit messages?

Feel free to ignore the other comments, I'm just going to leave them here out of spite for the old code :smile:

osbs/tekton.py
```diff
@@ -590,180 +355,357 @@ def log_and_sleep():
...
-            try:
-                build_response = self.wait(build_id, BUILD_FINISHED_STATES)
-                return build_response
-            except OsbsWatchBuildNotFound:
-                # this is woraround for https://github.com/openshift/origin/issues/2348
-                logger.warning("I'm going to wait again. Retry #%d.", retry)
-                continue
-        raise OsbsException("Failed to wait for a build: %s" % build_id)
+                # (Unknown, Started), (Unknown, PipelineRunCancelled)
+                logger.debug("Waiting for pipeline run, current status %s, reason %s", status, reason)
```
chmeliik, line 484: It's a bit confusing that Started doesn't mean started (at least according to this method). Could that be explained in a comment or docstring? Or maybe the method name needs to change?

chmeliik: Also, what should happen if the pipeline doesn't start? Can the last status be something other than 'True', 'False'? The method would just implicitly return None in that case

osbs/tekton.py
```diff
@@ -590,180 +355,357 @@ def log_and_sleep():
...
+            except KeyError:
+                logger.error("pipeline run does not have any task runs")
                 continue
-        raise OsbsException('Failed to schedule a build in {} attempts: {}'.format(WAIT_RETRY,
-                                                                                   build_id))
+            # pipeline run finished succesfully or failed
+            if status in ['True', 'False']:
+                yield task_runs
+            elif status == 'Unknown' and reason == 'Running':
+                yield task_runs
```
chmeliik, line 499: I'm having a hard time understanding what the method does. It yields a list of task runs every time the status changes, that much I understand, but what does that list actually mean? Is it a list of task runs currently in progress? A list of all the task runs that were created up until that point? Something else?

Would it make sense to de-duplicate the taskruns here and flatten the result from `Iterator[List[str]]` to `Iterator[str]`?

And is it necessary to check the status+reason?

chmeliik (commented):
Got a couple comments for PipelineRun, some of them also apply to TaskRun and maybe Pod.

Those three classes are similar in many ways, should they share a base class that implements the common stuff?

osbs/tekton.py
lkolacek: +1 can you add explanation docstring to this method please?

osbs/tekton.py
```diff
@@ -1,33 +1,20 @@
-"""
-Copyright (c) 2015, 2019 Red Hat, Inc
-All rights reserved.
-
-This software may be modified and distributed under the terms
-of the BSD license. See the LICENSE file for details.
-"""
-from __future__ import print_function, unicode_literals, absolute_import, division
 import json
```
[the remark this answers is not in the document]
mkosiarc: Fixed

osbs/tekton.py
```diff
@@ -590,180 +355,357 @@ def log_and_sleep():
...
+class PipelineRun():
+    def __init__(self, os, pipeline_name, pipeline_run_name):
+        self.os = os
+        self.pipeline_name = pipeline_name
+        self.pipeline_run_name = pipeline_run_name
+        self.api_path = 'apis'
+        self.api_version = API_VERSION
+
+    def start_pipeline_run(self):
+        url = self.os._build_url(
```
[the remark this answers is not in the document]
mkosiarc: Made those methods public

osbs/tekton.py
mkosiarc: I added explanation to docstring, and cleared up that method as per adam's suggestions.

osbs/tekton.py
```diff
@@ -590,180 +355,357 @@ def log_and_sleep():
...
+            },
+            "spec":{
+                "pipelineRef":{
+                    "name":self.pipeline_name
+                },
+                "status": "PipelineRunCancelled"
+            }
+        }
+        return self.os._patch(url, data=json.dumps(data),
+                          headers={"Content-Type": "application/merge-patch+json", "Accept": "application/json"})
```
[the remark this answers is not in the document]
mkosiarc: Added docs to commit msg.

osbs/tekton.py
mkosiarc: For your second comment - I think the last status cannot be anything other than True or False. So I am expecting that if the start pipeline code returns 200 OK status code, then the pipeline will have one of the statuses mentioned, Unknown, True or False.

chmeliik: Ack, the table here pretty much answered both my questions -> https://tekton.dev/docs/pipelines/pipelineruns/#monitoring-execution-status

I'm fine with keeping the name, maybe just include that link as an explanation

mkosiarc: kept the name, added doc link to method docstring

chmeliik (approved):
LGTM, assuming the duplication between tekton.py and core.py is removed later

PR #1046, opened 2021-10-13
Add unit tests for tekton module

mkosiarc:

### DIFF
--- a/osbs-client.spec
+++ b/osbs-client.spec
@@ -83,12 +83,6 @@ mv %{buildroot}%{_bindir}/osbs %{buildroot}%{_bindir}/osbs-%{python3_version}
 ln -s  %{_bindir}/osbs-%{python3_version} %{buildroot}%{_bindir}/osbs-3
 ln -s  %{_bindir}/osbs-%{binaries_py_version} %{buildroot}%{_bindir}/osbs
 
-%if 0%{?with_check}
-%check
-LANG=en_US.utf8 py.test-%{python3_version} -vv tests
-%endif # with_check
-
-
 %files
 %doc README.md
 %{_bindir}/osbs


MartinBasti:
```
Fix minor bugs
```

I have no idea what and why was fixed, this is not how commit should look like

osbs/tekton.py
```diff
@@ -331,40 +316,15 @@ def log_and_sleep():
...
-
-                if 'type' not in j:
-                    logger.error("Watch event has no 'type': %s", j)
-                    continue
-
-                yield (j['type'].lower(), j['object'])
+            logger.debug("retrieving fresh version of object")
+            fresh_response = self.get(get_url)
+            check_response(fresh_response)
+            yield fresh_response.json()
```
chmeliik, line 327: Wait, how does this method "watch" anything now? It will just just fetch current info about an object WATCH_RETRY times, no?

To be fair, the original method doesn't make a whole lot of sense to me either, but at least it iterates through a streamed response, which I thought was where the "watching" happened

rcerven (approved):

osbs-client.spec
```diff
@@ -83,12 +83,6 @@ mv %{buildroot}%{_bindir}/osbs %{buildroot}%{_bindir}/osbs-%{python3_version}
 ln -s  %{_bindir}/osbs-%{python3_version} %{buildroot}%{_bindir}/osbs-3
 ln -s  %{_bindir}/osbs-%{binaries_py_version} %{buildroot}%{_bindir}/osbs
 
-%if 0%{?with_check}
-%check
-LANG=en_US.utf8 py.test-%{python3_version} -vv tests
-%endif # with_check
```
chmeliik, line 89: Maybe also remove the section that defines with_check
```spec
%if 0%{?fedora}
# rhel/epel has older incompatible version of pytest (no caplog)
%global with_check 1
%endif
```

osbs/tekton.py
chmeliik: Ack, works for me :+1:

osbs-client.spec
mkosiarc: removed

mkosiarc: although there is another section that uses that check -
`%if 0%{?with_check}
BuildRequires:  python3-dateutil
BuildRequires:  python3-pytest
BuildRequires:  python3-flexmock
BuildRequires:  python3-six
BuildRequires:  python3-dockerfile-parse
BuildRequires:  python3-requests
BuildRequires:  python3-requests-kerberos
BuildRequires:  python3-PyYAML
BuildRequires:  python3-jsonschema
%endif # with_check`

chmeliik: Ah, then maybe keep everything related to `with_check` and set it to 0 instead?

chmeliik: Although that's probably going to mess with `test.sh`... why is everything always held together with duct tape? :smile: 
If it causes issues, just use what worked originally

chmeliik: Works for me, as long as `test.sh` still works (I think it installs dependencies based on the specfile)

mkosiarc: of course it didn't work, keeping the original change :D

mkosiarc: could you please update the checklist?

PR #1060, opened 2021-11-10
check if the repo ws locally modified before resetting

hjmodi:
In OSBS2 we'll reuse the same repo multiple times,
we need to make sure it's not modified locally
before resetting

* CLOUDBLD-7976

### DIFF
--- a/osbs/utils/__init__.py
+++ b/osbs/utils/__init__.py
@@ -38,7 +38,7 @@ from six.moves.urllib.parse import urlparse
 
 from dockerfile_parse import DockerfileParser
 from osbs.exceptions import (OsbsException, OsbsResponseException,
-                             OsbsValidationException, OsbsCommitNotFound)
+                             OsbsValidationException, OsbsCommitNotFound, OsbsLocallyModified)
 
 logger = logging.getLogger(__name__)
 ClonedRepoData = namedtuple('ClonedRepoData', ['repo_path', 'commit_id', 'commit_depth'])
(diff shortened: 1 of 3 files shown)


osbs/utils/__init__.py
```diff
@@ -158,11 +158,28 @@ def clone_git_repo(git_url, target_dir=None, commit=None, retry_times=GIT_MAX_RE
...
     retry_delay = GIT_BACKOFF_FACTOR
     target_dir = target_dir or os.path.join(tempfile.mkdtemp(), "repo")
     commit = commit or "master"
+
+    if os.path.isdir(target_dir) and len(os.listdir(target_dir)) != 0:
+        if exists_ok:
+            try:
+                logger.info("'%s' directory already exists, trying to checkout ref '%s' for "
+                            "repo %s", target_dir, commit, git_url)
+                repo_commit, repo_depth = reset_git_repo(target_dir, commit, depth)
```
chmeliik, line 173: `reset_git_repo` does a `git reset --hard`. That should be fine, because in OSBS2 we will only be modifying copies of the repo, not the original, but to be absolutely safe  maybe the reset function could instead raise an error if any changes would be discarded by the hard reset

hjmodi: updated to check if there are any local uncommitted changes.

rcerven (approved):

osbs/utils/__init__.py
```diff
@@ -148,7 +148,7 @@ def checkout_git_repo(git_url, target_dir=None, commit=None, retry_times=GIT_MAX
 
 
 def clone_git_repo(git_url, target_dir=None, commit=None, retry_times=GIT_MAX_RETRIES, branch=None,
-                   depth=None):
+                   depth=None, exists_ok=False):
```
chmeliik, line 151: Shouldn't be needed after the changes in https://github.com/containerbuildsystem/atomic-reactor/pull/1708

hjmodi: removed.

chmeliik (approved):
LGTM :+1:

PR #1067, opened 2021-12-07
pipeline run get error message, should be more careful about reading

rcerven:
non-existing keys, also cancel pipeline run if get_logs fails

* CLOUDBLD-8220

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -32,12 +32,18 @@ def print_json_nicely(decoded_json):
 
 
 def _print_pipeline_run_logs(pipeline_run, user_warnings_store):
+    """
+    prints pipeline run logs
+
+    :return: bool, True if logs are correctly returned,
+                   False when reading logs fails or is not iterable
+    """
     pipeline_run_name = pipeline_run.pipeline_run_name
 
     pipeline_run_logs = pipeline_run.get_logs(follow=True, wait=True)
     if not isinstance(pipeline_run_logs, collections.abc.Iterable):
         logger.error("'%s' is not iterable; can't display logs", pipeline_run_name)
-        return
+        return False
     print(f"Pipeline run created ({pipeline_run_name}), watching logs (feel free to interrupt)")
     try:
         for line in pipeline_run_logs:
@@ -46,9 +52,11 @@ def _print_pipeline_run_logs(pipeline_run, user_warnings_store):
                 continue
 
             print('{!r}'.format(line))
+        return True
     except Exception as ex:
         logger.error("Error during fetching logs for pipeline run %s: %s",
                      pipeline_run_name, repr(ex))
+        return False
 
 
 def _get_build_metadata(pipeline_run, user_warnings_store):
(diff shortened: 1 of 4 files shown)


osbs/cli/main.py
```diff
@@ -49,6 +49,7 @@ def _print_pipeline_run_logs(pipeline_run, user_warnings_store):
     except Exception as ex:
         logger.error("Error during fetching logs for pipeline run %s: %s",
                      pipeline_run_name, repr(ex))
+        return 1
```
mkosiarc, line 52: This seems very unpythonic to me, I would maybe prefer raising some exception here
and then caling this method in a try block

osbs/tekton.py
```diff
@@ -506,22 +506,36 @@ def get_error_message(self):
...
         err_message += "\npipeline run errors:\n"
-        task_runs_status = data['status']['taskRuns']
 
-        for task_name, stats in task_runs_status.items():
-            if stats['status']['conditions'][0]['reason'] == 'Succeeded':
-                continue
-
-            err_message += f"pipeline task '{task_name}' failed:\n"
+        if 'taskRuns' in data['status']:
+            task_runs_status = data['status']['taskRuns']
```
mkosiarc, line 511: task_runs_status = data['status'].get('taskRuns') or something similiar would work the same and we could save one conditional statement

mkosiarc: and one indent level

mkosiarc: I am sorry, I meant this
`task_runs_status = data['status'].get('taskRuns', {})`

osbs/cli/main.py
ben-alkov: Personally, I don't like to see:
-1 Completely different types being returned
-2 Multiple exit points. Inverting the logic of the if statement, assigning a return value, and exiting once with a return of that value is much cleaner and easier to follow

osbs/cli/main.py
```diff
@@ -80,14 +81,22 @@ def _get_build_metadata(pipeline_run, user_warnings_store):
 
 def print_output(pipeline_run, export_metadata_file=None):
     user_warnings_store = UserWarningsStore()
-    _print_pipeline_run_logs(pipeline_run, user_warnings_store)
+    get_logs_failed = _print_pipeline_run_logs(pipeline_run, user_warnings_store)
```
ben-alkov, line 84: Misleading, because getting the logs *technically* fails when `pipeline_run_logs` is *not* an instance of collections.abc.Iterable

osbs/cli/main.py
rcerven: we are silencing exception here so we don't have to catch it later, so why again raise another?

osbs/tekton.py
rcerven: ok will do that, but won't save conditional anyway since I have to do else part

osbs/cli/main.py
mkosiarc: then don't silence it and just raise it, so outside code depending on this function knows there was a failure based on an exception and not return value of 1.

osbs/cli/main.py
rcerven: I am not covering that case when it isn't even iterable because that is some weird corner case, I am covering ONLY when during get_logs openshift gives ConnectionError(ReadTimeoutError)

osbs/tekton.py
mkosiarc: ah, then you can keep the original condition if you want.

rcerven: already changed that

osbs/cli/main.py
rcerven: BUT following functions still have to print the summary ......
I am only returning 1 so I have some indications that get_logs really failed and if build still didn't finish additionally cancel it

rcerven: I've added return 1 also to the case when it isn't iterable

rcerven: ok returning 1 if everything is fine so now I am returning just int

chmeliik: Ack, could we have a docstring that explains what the return value means then?

rcerven: added

tests/cli/test_main.py
```diff
@@ -103,13 +106,26 @@ def test_print_output_failure(tmpdir, capsys):
...
-     ]))
+     ]
+
+    def get_logs():
+        for log in log_entries:
+            yield log
+        if get_logs_failed:
+            raise Exception("error reading logs")
+
+    ppln_run.should_receive('get_logs').and_return(get_logs())
```
chmeliik, line 122: How does this test work? `get_logs()` will raise an exception right on this line, how come the test is not failing?

chmeliik: Oh nvm, generators

chmeliik (approved):
LGTM :+1:

ben-alkov (approved):

rcerven:
no rn, fix for unreleased change

PR #1070, opened 2021-12-15
sanitize pipeline run label values for openshift

rcerven:
* CLOUDBLD-8241

### DIFF
--- a/osbs/tekton.py
+++ b/osbs/tekton.py
@@ -18,7 +18,7 @@ from osbs.constants import (DEFAULT_NAMESPACE, SERVICEACCOUNT_SECRET, SERVICEACC
                             SERVICEACCOUNT_CACRT)
 from osbs.osbs_http import HttpSession
 from osbs.kerberos_ccache import kerberos_ccache_init
-from osbs.utils import retry_on_conflict
+from osbs.utils import retry_on_conflict, sanitize_strings_for_openshift
 from urllib.parse import urljoin, urlencode, urlparse, parse_qs
 from requests.utils import guess_json_utf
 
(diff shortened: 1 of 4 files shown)


ben-alkov (approved):
:tada:

osbs/tekton.py
```diff
@@ -380,6 +380,11 @@ def start_pipeline_run(self):
             raise OsbsException("No input data provided for pipeline run to start")
 
         run_name = self.input_data.get('metadata', {}).get('name')
+        labels = self.input_data.get('metadata', {}).get('labels', {})
+
+        if labels:
```
tkdchen, line 385: This `if` looks imply two things, `metadata` might be omitted, if no, then `metadata.labels` might be omitted. However, the PipelineRun definition is read from predefined yaml file, and in function `_set_binary_container_pipeline_data`, the metadata  is accessed directly:

```
pipeline_run_data['metadata']['name'] = pipeline_run_name
```

If the `pipeline_run_data` should always has key `metadata`, then it is unnecessary to get metadata by using `dict.get`, `self.input_data.get('metadata', {})`.

Meanwhile, IMO, `start_pipeline_run` is not a good place to prepare the labels, it should just start the pipeline run as the name implies. `_set_binary_container_pipeline_data` prepares the pipeline run data including labels, there should be the right place to sanitize the lable values.

osbs/tekton.py
```diff
@@ -380,6 +380,11 @@ def start_pipeline_run(self):
...
+            sanitized_labels = {k: sanitize_strings_for_openshift(v) for k, v in labels.items()}
```
tkdchen, line 386: `https://github.com/blah` is sanitized into `httpsgithub.comblah`. Is this what we need in the metadata.labels?

osbs/tekton.py
chmeliik: +1 to sanitize labels in the `_set_*_pipeline_data` methods instead

osbs/tekton.py
rcerven: well they were there before, we can think about removing some of them in the future

osbs/tekton.py
rcerven: about first note, this method should be general one, not JUST the one which is used by _set* methods

rcerven: I prefer to handle it in tekton module, since it is ocp thing, I could make helper method here and call that if you want, but I want to keep it here

chmeliik: That's a good point. I would argue that those api methods already have to deal with some Tekton/OCP details that they maybe shouldn't, but I agree that the tekton module is a better place to sanitize labels

chmeliik (approved):
LGTM :+1:

rcerven:
no rn update, fix for unreleased feature

PR #1071, opened 2021-12-22
wait_for_finish method, which waits for pipeline run to finish, after we

rcerven:
finished reading logs, there is race between all pods finished and
pipeline run changing status

we will wait just for certain amount of time, if pipeline run still
didn't finish by then, there is something really wrong and that will be
already handled

* CLOUDBLD-8359

- n/a] JSON/YAML configuration changes are updated in the relevant schema

### DIFF
--- a/osbs/tekton.py
+++ b/osbs/tekton.py
@@ -27,6 +27,9 @@ logger = logging.getLogger(__name__)
 # Retry each connection attempt after 5 seconds, for a maximum of 10 times
 WATCH_RETRY_SECS = 5
 WATCH_RETRY = 10
+# Retry waiting for pipeline to finish for 5 seconds, for a maximum of 10 times
+WAIT_RETRY_SECS = 5
+WAIT_RETRY = 10
 MAX_BAD_RESPONSES = WATCH_RETRY // 3
 # Give up after 12 hours
 WAIT_RETRY_HOURS = 12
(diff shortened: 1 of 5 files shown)


hjmodi (approved):

lkolacek (approved):
Code change looks good to me. Have you captured this race condition issue? Can you post more info to CLOUDBLD-8359?

osbs/tekton.py
```diff
@@ -556,6 +559,16 @@ def has_not_finished(self):
...
 
+    def wait_for_finish(self):
+        for _ in range(WAIT_RETRY):
+            if self.has_not_finished():
+                logger.info("Waiting for pipeline run '%s' to finish, sleep for %ss",
+                            self.pipeline_run_name, WAIT_RETRY_SECS)
+                time.sleep(WAIT_RETRY_SECS)
+            else:
+                logger.info("Pipeline run '%s' finished", self.pipeline_run_name)
+                break
```
chmeliik, line 570: Should we raise an exception if the loop exits normally (i.e. the pipeline does not finish within the timeout)?

osbs/tekton.py
```diff
@@ -556,6 +559,16 @@ def has_not_finished(self):
     def was_cancelled(self):
         return self.status_reason == 'PipelineRunCancelled'
 
+    def wait_for_finish(self):
```
chmeliik, line 562: I would appreciate a docstring that explains how this method should be used - only for "almost finished" pipeline runs, not newly started ones.

osbs/tekton.py
rcerven: there isn't any benefit in that, we will then later anyway call build_has_succeeded and build_not_finished (and cancel)

this is really just for getting some time for pipeline run to finish

osbs/tekton.py
rcerven: added

osbs/tekton.py
chmeliik: Ok, as long as the purpose of the function is documented this seems fine

chmeliik (approved):

rcerven:
no rn update, fix for unreleased changes

PR #1091, opened 2022-04-25
fix task name for the logs

hjmodi:
Return the logs with `pipelineTaskName` instead of
task run name. Since, task run name is just
pipeline run name appended with some random
ID it's not of much use. Instead, the pipeline
task name is the one we need since it can help
distinguish the platforms in the logs of specific
tasks.

* CLOUDBLD-9886


chmeliik (approved):
LGTM :+1:

chmeliik:
Just a reminder that osbs-client version in atomic-reactor will need to be bumped after this is merged

rcerven (approved):

PR #1093, opened 2022-05-10
Revert "bump json schema draft version to allow deprecated key"

rcerven:
This reverts commit dec83c506b003866efede03137cbf4d1c209890e.


hjmodi (approved):

tkdchen (approved):
BTW, what's the reason of this revert?

chmeliik:
Don't we also need to remove the `deprecated` keys? AFAIK that was the reason why the version was bumped in the first place
https://github.com/containerbuildsystem/osbs-client/blob/aad8c396bcc907f2bdcb1c88ceaf99725719c26c/osbs/schemas/container.json#L107

@tkdchen the reason is that osbs-client needs to work with jsonschema < 3, which doesn't have `Draft7Validator` :disappointed:
(because osbs-client is installed on Brew builders, which are RHEL machines, with ancient libraries)

hjmodi:
> Don't we also need to remove the `deprecated` keys? AFAIK that was the reason why the version was bumped in the first place
> 
> https://github.com/containerbuildsystem/osbs-client/blob/aad8c396bcc907f2bdcb1c88ceaf99725719c26c/osbs/schemas/container.json#L107
> 
> @tkdchen the reason is that osbs-client needs to work with jsonschema < 3, which doesn't have `Draft7Validator` disappointed (because osbs-client is installed on Brew builders, which are RHEL machines, with ancient libraries)

It doesn't fail, it doesn't do anything tbh. I just added it to be safe.

chmeliik (approved):

PR #1097, opened 2022-05-27
handle missing 'terminated' key for tasksteps in get_error_message

rcerven:
* CLOUDBLD-9865

### DIFF
--- a/osbs/tekton.py
+++ b/osbs/tekton.py
@@ -498,14 +498,19 @@ class PipelineRun():
 
             if 'steps' in stats['status']:
                 for step in stats['status']['steps']:
-                    exit_code = step['terminated']['exitCode']
-                    if exit_code == 0:
-                        continue
+                    if 'terminated' in step:
+                        exit_code = step['terminated']['exitCode']
+                        if exit_code == 0:
+                            continue
 
-                    reason = step['terminated']['reason']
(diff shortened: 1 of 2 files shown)


osbs/tekton.py
```diff
@@ -498,14 +498,19 @@ def get_error_message(self):
...
+                        exit_code = step['terminated']['exitCode']
+                        if exit_code == 0:
+                            continue
+
+                        reason = step['terminated']['reason']
+                        err_message += f"task step '{step['name']}' failed with exit " \
+                                       f"code: {exit_code} " \
+                                       f"and reason: '{reason}'"
+                    else:
+                        err_message += f"task step '{step['name']}' is missing 'terminated' key " \
```
chmeliik, line 511: If we logged something here about the taskRun status, could it help us figure  out why this is happening? Or is this one of those things that are logged only in koji-containerbuild that we'll never see?

chmeliik (commented):
LGTM, just wondering if we could do something to help debug the problem eventually

osbs/tekton.py
rcerven: task run status doesn't contain anything useful apart from succeeded or failed, if you want details you need to check steps, which is this code doing

mkosiarc (approved):

rcerven:
no rn update, fix for unreleased change

PR #1100, opened 2022-06-06
save entire pipeline info in exported metadata

hjmodi:
* CLOUDBLD-10188

### DIFF
--- a/osbs/cli/main.py
+++ b/osbs/cli/main.py
@@ -64,6 +64,7 @@ def _get_build_metadata(pipeline_run, user_warnings_store):
         "pipeline_run": {
             "name": pipeline_run.pipeline_run_name,
             "status": pipeline_run.status_reason,
+            "info": {}
         },
         "results": {
             "user_warnings": [],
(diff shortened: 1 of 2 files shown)


osbs/cli/main.py
```diff
@@ -64,6 +64,7 @@ def _get_build_metadata(pipeline_run, user_warnings_store):
         "pipeline_run": {
             "name": pipeline_run.pipeline_run_name,
             "status": pipeline_run.status_reason,
+            "info": {}
```
tkdchen, line 67: Instead of injecting this "info", I suggest a new structure for this output:

```python
{
    "pipeline_run": pipeline_run info,
    "user_warnings": [],
    "error_msg": "",
}
```

Therefor, we have a compact structure and don't need to duplicate the name, status and repositories.

hjmodi: I don't think that's a good idea because those are derived properties and if we remove it from here, the code that currently depends on it will have to replicate the logic to find those properties in multiple places.

chmeliik (approved):
LGTM :+1:

PR #1102, opened 2022-06-15
Fix delay in log retrieval

hjmodi:
* CLOUDBLD-10225

### DIFF
--- a/osbs/tekton.py
+++ b/osbs/tekton.py
@@ -665,7 +665,7 @@ class PipelineRun():
             except KeyError:
                 logger.error("pipeline run does not have any status")
                 continue
-            # pipeline run finished succesfully or failed
+            # pipeline run finished successfully or failed
             if status in ['True', 'False']:
                 return pipeline_run
             elif status == 'Unknown' and reason == 'Running':
@@ -683,7 +683,7 @@ class PipelineRun():
         does not have information about all of its task runs, especially when there are multiple
         sequential tasks.
         """
-        watched_task_runs = []
+        watched_task_runs = set()
         for pipeline_run in self.os.watch_resource(
                 self.api_path,
                 self.api_version,
@@ -697,10 +697,16 @@ class PipelineRun():
                 continue
             for task_run_name, task_run_data in task_runs.items():
                 if task_run_name not in watched_task_runs:
-                    watched_task_runs.append(task_run_name)
+                    watched_task_runs.add(task_run_name)
                     yield task_run_data['pipelineTaskName'], task_run_name
-            # all task runs accounted for
-            if len(pipeline_run['status']['pipelineSpec']['tasks']) == len(task_runs):
+
+            try:
+                status = pipeline_run['status']['conditions'][0]['status']
+            except KeyError:
+                logger.error("pipeline run does not have any status")
+                return
+            # pipeline run finished successfully or failed
+            if status in ['True', 'False']:
                 return
 
     def _get_logs(self):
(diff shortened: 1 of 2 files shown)


osbs/tekton.py
```diff
@@ -683,25 +683,31 @@ def wait_for_taskruns(self):
...
         """
-        watched_task_runs = []
+        watched_task_runs = set()
         for pipeline_run in self.os.watch_resource(
                 self.api_path,
                 self.api_version,
                 resource_type="pipelineruns",
                 resource_name=self.pipeline_run_name,
         ):
+            try:
```
rcerven, line 693: just use https://github.com/containerbuildsystem/osbs-client/blob/master/osbs/tekton.py#L553

osbs/tekton.py
```diff
@@ -683,25 +683,31 @@ def wait_for_taskruns(self):
...
             except KeyError:
                 logger.error("pipeline run does not have any task runs")
                 continue
             for task_run_name, task_run_data in task_runs.items():
                 if task_run_name not in watched_task_runs:
-                    watched_task_runs.append(task_run_name)
+                    watched_task_runs.add(task_run_name)
                     yield task_run_data['pipelineTaskName'], task_run_name
-            # all task runs accounted for
-            if len(pipeline_run['status']['pipelineSpec']['tasks']) == len(task_runs):
```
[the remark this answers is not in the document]
hjmodi: moved

rcerven (approved):

osbs/tekton.py
```diff
@@ -699,6 +699,14 @@ def wait_for_taskruns(self):
                 if task_run_name not in watched_task_runs:
                     watched_task_runs.add(task_run_name)
                     yield task_run_data['pipelineTaskName'], task_run_name
+            try:
+                status = pipeline_run['status']['conditions'][0]['status']
+            except KeyError:
+                logger.error("pipeline run does not have any status")
+                return
```
chmeliik, line 706: Should we `return` here? The other method (`wait_for_start`) keeps waiting if there's no status

tests/test_tekton.py
```diff
@@ -820,20 +821,25 @@ def test_wait_for_start(self, pipeline_run):
         responses.add(responses.GET, PIPELINE_RUN_URL, json=PIPELINE_RUN_JSON)
         resp = pipeline_run.wait_for_start()
 
-        assert len(responses.calls) == 2
```
chmeliik, line 823: Why the changes in this test? There were no changes in the method

(I'm guessing the `time.sleep()` mocking is to make things faster, but what about this one?)

osbs/tekton.py
```diff
@@ -697,11 +697,8 @@ def wait_for_taskruns(self):
                 continue
             for task_run_name, task_run_data in task_runs.items():
                 if task_run_name not in watched_task_runs:
-                    watched_task_runs.append(task_run_name)
+                    watched_task_runs.add(task_run_name)
                     yield task_run_data['pipelineTaskName'], task_run_name
-            # all task runs accounted for
-            if len(pipeline_run['status']['pipelineSpec']['tasks']) == len(task_runs):
-                return
```
chmeliik, line 704: Now that I think about it, this commit on its own only makes the problem worse, by removing the only early-return condition that this method had. It's fixed in the next one, but maybe it should be a single commit.

osbs/tekton.py
rcerven: wait_for_start is waiting for prun to start, and AT this point in wait_for_tasks, prun already started so it has status, and if not something broke with prun I'd say

chmeliik: Hmm, maybe raise an error then? Or don't even catch the KeyError and let it fail?

chmeliik: Hmm, alright then. I don't like ignoring this kind of error (one that should never happen), but fair enough :)

chmeliik (approved):
LGTM :+1:

PR #1103, opened 2022-06-23
Remove dead code

hjmodi:


rcerven (approved):

osbs/cli/main.py
```diff
@@ -422,14 +417,8 @@ def cli():
...
     parser, args = cli()
-    try:
-        os_conf = Configuration(conf_file=args.config,
-                                cli_args=args)
-    except OsbsException as ex:
-        logger.error("Configuration error: %s", ex.message)
-        return -1
 
-    is_verbose = os_conf.get_verbosity()
+    is_verbose = args.verbose
```
chmeliik, line 403: Do we only want to get verbosity from the CLI, not from the config file?

(the TBD comment about that was in the prev. commit, but I don't know if we were sure about it or if it was just an idea)

osbs/cli/main.py
```diff
@@ -406,14 +401,14 @@ def cli():
                         help="Export build metadata as JSON file")
     args = parser.parse_args()
 
-    if getattr(args, 'func', None) is cmd_build_source_container:
+    if getattr(args, 'func') is cmd_build_source_container:
```
chmeliik, line 386: Later in `main()` there is logic about printing the help text if the `func` attribute is not set. If that's not relevant anymore, should it also be cleaned up? Does the parser guarantee that func is set if `parse_args()` succeeds?

mkosiarc (approved):

osbs/cli/main.py
hjmodi: hmm, we do need to show help. fixed this to default to `None` instead of failing

chmeliik (approved):
LGTM :+1:

hjmodi:
> @hjmodi there's a typo in a commit message
> 
> > remove string conversion function needed for py2
> > we no longer need to support py3
> 
> I am assuming it's supposed to be "we no longer need to support py2"

fixed the commit message

PR #1105, opened 2022-07-14
don't fail because of connection or read timeouts and retry on read i…

rcerven:
…ssues

### DIFF
--- a/osbs/tekton.py
+++ b/osbs/tekton.py
@@ -26,16 +26,15 @@ from requests.utils import guess_json_utf
 
 logger = logging.getLogger(__name__)
 
-# Retry each connection attempt after 5 seconds, for a maximum of 10 times
+# Retry each connection attempt after 5 seconds, for a maximum of 20 times
 WATCH_RETRY_SECS = 5
-WATCH_RETRY = 10
-# Retry waiting for pipeline to finish for 5 seconds, for a maximum of 10 times
+WATCH_RETRY = 20
+MAX_BAD_RESPONSES = 20
+
+# Retry waiting for pipeline to finish for 5 seconds, for a maximum of 5 hours
 WAIT_RETRY_SECS = 5
-WAIT_RETRY = 10
-MAX_BAD_RESPONSES = WATCH_RETRY // 3
-# Give up after 12 hours
-WAIT_RETRY_HOURS = 12
-WAIT_RETRY = WAIT_RETRY_HOURS * 3600 // (WATCH_RETRY_SECS * WATCH_RETRY)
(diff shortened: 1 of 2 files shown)


chmeliik (commented):
This seems to mix multiple changes into one commit. Changing constants, changing retry logic, changing which exceptions we retry on. For some of them, I'm not sure I understand why the changes were made. Could you explain each change in the commit message (and/or split them into multiple commits)?

osbs/tekton.py
```diff
@@ -26,16 +26,15 @@
...
+
+# Retry waiting for pipeline to finish for 5 seconds, for a maximum of 5 hours
 WAIT_RETRY_SECS = 5
-WAIT_RETRY = 10
-MAX_BAD_RESPONSES = WATCH_RETRY // 3
-# Give up after 12 hours
-WAIT_RETRY_HOURS = 12
-WAIT_RETRY = WAIT_RETRY_HOURS * 3600 // (WATCH_RETRY_SECS * WATCH_RETRY)
+WAIT_RETRY_HOURS = 5
+WAIT_RETRY = (WAIT_RETRY_HOURS * 3600) // WATCH_RETRY_SECS
```
chmeliik, line 37:
```suggestion
WAIT_RETRY = (WAIT_RETRY_HOURS * 3600) // WAIT_RETRY_SECS
```

osbs/tekton.py
```diff
@@ -318,40 +317,45 @@ def log_and_sleep():
...
-                yield fresh_response.json()
+
+            except OsbsException as exc:
+                if (not isinstance(exc.cause, requests.ConnectionError) and
+                        not isinstance(exc.cause, requests.Timeout)):
+                    raise
+            except requests.exceptions.ConnectionError:
+                pass
+            except requests.exceptions.Timeout:
+                pass
```
chmeliik, line 358: Why do we now also retry on these? Doesn't the underlying `HttpStream` class already do these retries?

rcerven: HttpsStream handles those, but this is also done in _stream_logs

and since we are now using watcher way more than in osbs1, I've added it here as well

osbs/tekton.py
chmeliik: I see. The only weird detail is that the formula uses **WATCH**_RETRY_SECS - I'm guessing that should be **WAIT**_RETRY_SECS?

rcerven: right, fixed

chmeliik (approved):
LGTM :+1:

ben-alkov (approved):
LGTM

PR #1106, opened 2022-07-27
Make ImageName objects immutable

nerieae:
CLOUDBLD-8232

Instances of the ImageName class are currently both hashable and mutable at the same time. This commit makes them immutable and adds a replace() method to allow for changing the object's attributes by returning a copy, not by the in-place mutation. **Code that mutates ImageName objects in place is not changed just yet.**

Signed-off-by: Oliver Lesko <olesko@redhat.com>

### DIFF
--- a/osbs/utils/__init__.py
+++ b/osbs/utils/__init__.py
@@ -383,12 +383,29 @@ class ImageName(object):
                       |------------------------| image
     |------------------------------------------| image
     """
+    __DEFAULT = object()  # constant class variable to allow for None assignment
 
     def __init__(self, registry=None, namespace=None, repo=None, tag=None):
-        self.registry = registry
-        self.namespace = namespace
-        self.repo = repo
-        self.tag = tag or 'latest'
+        self.__registry = registry
+        self.__namespace = namespace
+        self.__repo = repo
+        self.__tag = tag or 'latest'
+
+    @property
+    def registry(self):
+        return self.__registry
+
+    @property
+    def namespace(self):
+        return self.__namespace
+
+    @property
+    def repo(self):
+        return self.__repo
+
+    @property
+    def tag(self):
+        return self.__tag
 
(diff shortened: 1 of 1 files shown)


osbs/utils/__init__.py
```diff
@@ -383,12 +383,29 @@ class ImageName(object):
...
     |------------------------------------------| image
     """
+    __DEFAULT = object()  # constant class variable to allow for None assignment
 
     def __init__(self, registry=None, namespace=None, repo=None, tag=None):
-        self.registry = registry
-        self.namespace = namespace
-        self.repo = repo
-        self.tag = tag or 'latest'
+        self.__registry = registry
```
MartinBasti, line 389: I'm not sure if want to use double underscore, it makes inheritance hard. Would be enough to have just one?

osbs/utils/__init__.py
```diff
@@ -403,17 +420,16 @@ def parse(cls, image_name):
...
-            result.registry = s[0]
-            result.namespace = s[1]
-        result.repo = s[-1]
+            result = result.replace(registry=s[0], namespace=s[1], repo=s[2])
 
         for sep in '@:':
             try:
-                result.repo, result.tag = result.repo.rsplit(sep, 1)
+                repo, tag = result.repo.rsplit(sep, 1)
+                result = result.replace(repo=repo, tag=tag)
```
chmeliik, line 432: Instead of `replace`-ing the previous result, could we just figure out the `registry`, `repo` etc. variables and create the ImageName object at the end?

chmeliik (commented):
One minor comment, otherwise looks good so far :+1:

osbs/utils/__init__.py
MartinBasti: Would make sense to use frozen dataclass here, that should provide all what we need?

PR #1107, opened 2022-08-04
get_error_message won't show anymore all failed tasks/steps, but only

rcerven:
those which have actual error message set based on task results

* CLOUDBLD-10627

### DIFF
--- a/osbs/tekton.py
+++ b/osbs/tekton.py
@@ -507,9 +507,7 @@ class PipelineRun():
 
         if plugin_errors:
             for plugin, error in plugin_errors.items():
-                err_message += f"Error in plugin {plugin}: {error}\n"
-
-        err_message += "\npipeline run errors:\n"
+                err_message += f"Error in plugin {plugin}: {error};\n"
 
         task_runs_status = data['status'].get('taskRuns', {})
 
@@ -518,8 +516,6 @@ class PipelineRun():
             if stats['status']['conditions'][0]['reason'] == 'Succeeded':
                 continue
 
-            err_message += f"pipeline task '{task_name}' failed:\n"
-
             if 'steps' in stats['status']:
                 for step in stats['status']['steps']:
                     if 'terminated' in step:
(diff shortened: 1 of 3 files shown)


ben-alkov (approved):

osbs/tekton.py
```diff
@@ -518,34 +518,22 @@ def get_error_message(self):
...
-                           f" '{pipeline_run_condition['message']}'\n"
+                        if 'message' in step['terminated']:
+                            try:
+                                message_json = json.loads(step['terminated']['message'])
+                                for message in message_json:
+                                    if message['key'] == 'task_result':
+                                        err_message += f"Error in {task_name}: " \
+                                                       f"{message['value']};\n"
+                            except Exception:
+                                continue
```
chmeliik, line 536: Does this have to catch the generic `Exception`?

Should we at least log the exception?

osbs/tekton.py
```diff
@@ -507,9 +507,9 @@ def get_error_message(self):
 
         if plugin_errors:
             for plugin, error in plugin_errors.items():
-                err_message += f"Error in plugin {plugin}: {error}\n"
+                err_message += f"Error in plugin {plugin}: {error};\n"
 
-        err_message += "\npipeline run errors:\n"
+        err_message += "\npipeline run failed;\n"
```
chmeliik, line 512: Do we want this here in the middle/start of the error message? I think most of the time, only `Error in plugin ...` and `Error in <task> ...` will be relevant.

Maybe, if we go through plugin_errors and all taskRuns and still have no error message, we could then add this at the end?  Something like `pipeline run failed for unknown reason`?

osbs/tekton.py
rcerven: I don't want it to fail here for any reason, because then it would fail for koji-c as well and we wouldn't get any useful messages

can log it

osbs/tekton.py
rcerven: ok

chmeliik (approved):
LGTM :+1:

rcerven:
no rn updated, fix for unreleased feature

PR #1110, opened 2022-08-26
rotate fetching logs

hjmodi:
* CLOUDBLD-10692

### DIFF
--- a/osbs/tekton.py
+++ b/osbs/tekton.py
@@ -720,10 +720,13 @@ class PipelineRun():
                     "Pipeline run '%s' does not have any task runs yet",
                     self.pipeline_run_name)
                 continue
+            current_task_runs = []
             for task_run_name, task_run_data in task_runs.items():
                 if task_run_name not in watched_task_runs:
                     watched_task_runs.add(task_run_name)
-                    yield task_run_data['pipelineTaskName'], task_run_name
+                    current_task_runs.append((task_run_data['pipelineTaskName'], task_run_name))
+
+            yield current_task_runs
 
             try:
                 status = pipeline_run['status']['conditions'][0]['status']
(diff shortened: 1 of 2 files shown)


osbs/tekton.py
```diff
@@ -720,10 +720,13 @@ def wait_for_taskruns(self):
                     "Pipeline run '%s' does not have any task runs yet",
                     self.pipeline_run_name)
                 continue
+            current_task_runs = []
             for task_run_name, task_run_data in task_runs.items():
                 if task_run_name not in watched_task_runs:
                     watched_task_runs.add(task_run_name)
-                    yield task_run_data['pipelineTaskName'], task_run_name
+                    current_task_runs.append((task_run_data['pipelineTaskName'], task_run_name))
```
chmeliik, line 727: Is it possible that, when running the 4 parallel tasks, they won't start at the same time and `current_task_runs` will only include some of them?

hjmodi: yes, it seems like it is possible

hjmodi: not a 100% sure

rcerven: I don't think it is about if those tasks start at the same time, but if they appear in ['status']['taskRuns'] at the same moment, in some new state

which should most likely happen, I'd try this out, since this is at least simple solution

rcerven (approved):

osbs/tekton.py
chmeliik: I've stopped believing that Tekton will do the reasonable thing, but +1 to test the current solution :smile:

osbs/tekton.py
```diff
@@ -751,10 +754,19 @@ def _get_logs(self):
 
     def _get_logs_stream(self):
         self.wait_for_start()
-        for pipeline_task_name, task_run_name in self.wait_for_taskruns():
-            for log_line in TaskRun(os=self.os, task_run_name=task_run_name).get_logs(follow=True,
-                                                                                      wait=True):
-                yield pipeline_task_name, log_line
+        steaming_task_runs = {}
```
chmeliik, line 757: typo: streaming

chmeliik (approved):
LGTM, just a typo

chmeliik:
Thinking if there's any way to unit-test that the behaviour is what we want. If you didn't come up with any simple solution, feel free to ignore that check-list item

osbs/tekton.py
hjmodi: fixed it

hjmodi:
> Thinking if there's any way to unit-test that the behaviour is what we want. If you didn't come up with any simple solution, feel free to ignore that check-list item

extended the test to simulate behaviour

PR #1112, opened 2022-09-07
Fix sorting key

mkosiarc:
There might be cases where the task runs do not
have the startTime, mostly just if they did not execute

I believe that this is the problem behind CLOUDBLD-10953

### DIFF
--- a/osbs/tekton.py
+++ b/osbs/tekton.py
@@ -55,9 +55,22 @@ def check_response(response, log_level=logging.ERROR):
 
 
 def get_sorted_task_runs(task_runs: Dict[str, Any]) -> List[Tuple[str, Dict[str, Any]]]:
-    return sorted(task_runs.items(),
-                  key=lambda x: datetime.strptime(x[1]['status']['startTime'],
-                                                  '%Y-%m-%dT%H:%M:%SZ').timestamp())
+    def custom_key(x):
+        """
+        Handles cases where the startTime key is missing.
+        These items are put at the end.
+        """
+        missing_start_time = "startTime" not in x[1]["status"]
+        containing_start_time = (
+            datetime.strptime(
+                x[1]["status"]["startTime"], "%Y-%m-%dT%H:%M:%SZ"
+            ).timestamp()
+            if not missing_start_time
+            else None
+        )
+        return (missing_start_time, containing_start_time)
+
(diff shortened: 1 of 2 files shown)


osbs/tekton.py
```diff
@@ -55,9 +55,21 @@ def check_response(response, log_level=logging.ERROR):
...
+        Handles cases where the startTime key is missing.
+        These items are put at the end.
+        """
+        missing_start_time = ('startTime' not in x[1]['status'])
+        return (
+            missing_start_time,
+            datetime.strptime(x[1]["status"]["startTime"], "%Y-%m-%dT%H:%M:%SZ").timestamp()
+            if not missing_start_time
+            else None,
+        )
```
chmeliik, line 70: minor readability suggestion:
 ```suggestion
        return (missing_start_time, start_time)
```
(and do the if else logic above)

File: tests/test_tekton.py [hunk 2]
```diff
@@ -1209,3 +1209,21 @@ def custom_watch(api_path, api_version, resource_type, resource_name,
...
+        "TaskRunSecond": {"status": {"startTime": "2022-09-07T18:23:51Z"}},
+        "TaskRunThird": {"status": {"startTime": "2022-09-07T18:24:51Z"}},
+        "TaskRunFirst": {"status": {"startTime": "2022-09-07T18:22:51Z"}},
+        "TaskRunMissingKey": {"status": {}},
+    }
+    expected_sorted = [
+        ("TaskRunFirst", {"status": {"startTime": "2022-09-07T18:22:51Z"}}),
+        ("TaskRunSecond", {"status": {"startTime": "2022-09-07T18:23:51Z"}}),
+        ("TaskRunThird", {"status": {"startTime": "2022-09-07T18:24:51Z"}}),
+        ("TaskRunMissingKey", {"status": {}}),
```
chmeliik, line 1225: It might be good to have at least two task runs with missing startTime in the unit test. I think it will still work fine but not 100% sure
```python
>>> (True, None) < (True, None)
False
```

mkosiarc: updated

osbs/tekton.py
mkosiarc: updated

chmeliik (approved):
LGTM :+1:

rcerven (approved):

PR #1114, opened 2022-09-29
handle cases in tekton module when tekton object is removed during run

rcerven:

### DIFF
--- a/osbs/tekton.py
+++ b/osbs/tekton.py
@@ -40,7 +40,7 @@ WAIT_RETRY = (WAIT_RETRY_HOURS * 3600) // WAIT_RETRY_SECS
 API_VERSION = "tekton.dev/v1beta1"
 
 
-def check_response(response, log_level=logging.ERROR):
+def check_response(response, log_level=logging.INFO):
     if response.status_code not in (
             requests.status_codes.codes.ok,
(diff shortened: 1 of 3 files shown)


osbs/tekton.py
```diff
@@ -592,6 +599,11 @@ def has_succeeded(self):
         return status_reason == 'Succeeded'
 
     def has_not_finished(self):
+        data = self.data
```
mkosiarc, line 602: I notice we do this assignment in code more often. Do we need this or could simply use self.data?

mkosiarc (approved):

osbs/tekton.py
rcerven: self.data is property which calls get_info to get fresh object

IF we would be using self.data everywhere, it would be constantly getting new data even tho they are still the same

I guess it isn't required in all cases (when used just once), but it certainly doesn't hurt

ben-alkov (approved):

osbs/tekton.py
mkosiarc: Yes, I was refering to cases when it is used just once - such as this one. But it is not a big issue.

rcerven:
fix to unreleased feature, no rn

PR #1146, opened 2024-12-06
feat(utils): implement symlink sandbox

ben-alkov:

What/why: implement detection/removal of unsafe symlinks in repos, specifically
covering cachi2 use case: Cachito already does this

How:

- copypasta `_enforce_sandbox()` and related unit tests from Cachito
  ("cachito/cachito/workers/tasks/general.py" and
   "cachito/tests/test_workers/test_tasks/test_general.py", respectively)
- add call to `_enforce_sandbox()`
- add CLI boolean arg `remove-unsafe-symlinks`, which toggles removing all
  symlinks which point to location(s) outside of any cloned repository

### DIFF
--- a/osbs/utils/__init__.py
+++ b/osbs/utils/__init__.py
@@ -12,6 +12,7 @@ import json
 import logging
 import os
 import os.path
+from pathlib import Path
 import random
 import re
 import shutil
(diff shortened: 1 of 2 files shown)


osbs/utils/__init__.py
```diff
@@ -69,6 +70,50 @@ def __repr__(self):
         return self.uri
 
 
+def _enforce_sandbox(repo_root: str, *, remove_unsafe_symlinks: bool) -> None:
```
MartinBasti, line 73: This shouldn't be a private method if we want to use it in atomic-reactor

ben-alkov:
Moved implementation to atomic-reactor - it fits better there.

